PR #8, opened 2015-04-07
Load videos remotely from the network

qingfeng:
pls, review. thanks

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -53,7 +53,13 @@
 
 - (void)setSrc:(NSString *)source
 {
-  _videoURL = [[NSURL alloc] initFileURLWithPath:[[NSBundle mainBundle] pathForResource:source ofType:@"mp4"]];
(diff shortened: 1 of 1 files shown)


brentvatne:
@qingfeng - This is a good step forward, ultimately I would like to move towards supporting the same syntax as `Image` by having the `source` prop be an object with the `uri` param. This would work well with adding an `onLoadStart` callback too, so we could show some kind of loading indicator.

One nit and I'll merge - could you change indentations to two spaces instead of tabs?

Thanks! :smile:

brianchirls:
Yes, good start. I think we'd need to scan for file extensions as well for local files. It may not always be mp4.

qingfeng:
sorry, I modified the two space

brentvatne:
Thanks :smile_cat:

qingfeng:
:beer:

PR #9, opened 2015-04-08
Initial set of changes to make urls work

johanneslumpe:
Ok this is very much a work in progress. URL-loading works now, but I might have broken playing of local files. Or it might be an issue with my project not recognizing the local file.

I assume this PR will still change quite a bit, but it's a start :)


johanneslumpe:
I think we should move the progress update timer creating and of course invalidation into `setSrc`

qingfeng:
:beer:

brianchirls:
Good progress. What happens if the source is changed?

Again, I suggest changing `onProgress` to `onTimeupdate` (or `onTimeUpdate`) so progress can be used for notifications of partial buffering updates. See [here](http://stackoverflow.com/questions/8996355/is-kvo-on-avplayeritem-loadedtimeranges-possible) for an idea on how to implement that, by the way. Refer to my note in #8. Thanks.

johanneslumpe:
@brianchirls Right now, if you change the source, the old player will be discarded and the new source will be loaded and will start playing as soon as it becomes ready to play.

Your note in #8 refers to being able to change the file type - that can be done now, by setting the `type` property of the `src` object. But you mentioned checking for a file extension. That might also be a nice way to just automatically detect it. I'd propose that we do both. If `type` is not set - fall back to extension scanning. If `type` is set, use the passed in type, regardless of the file's actual extension.

brianchirls:
Discarding the old player sounds good.

Oops, sorry. I meant my notes in https://github.com/brentvatne/react-native-video/issues/7#issuecomment-90703802

brentvatne:
@brianchirls

> Again, I suggest changing onProgress to onTimeupdate (or onTimeUpdate) so progress can be used for notifications of partial buffering updates. See here for an idea on how to implement that, by the way. Refer to my note in #8. Thanks.

`onTimeUpdate` is a good idea, in the interest of not making this such a big diff I'll open another issue for that

brentvatne:
Added an example project to test it out, all working great! Thanks!

PR #21, opened 2015-04-11
Add end notification

mwilc0x:
This PR allows for user to handle end of file playback. 

The currentTime == duration approach doesn't work so well as it depends on comparing time, and I found that there were multiple progress updates for a given second.

User is given option to have an onEnd prop which they can then use to call a function to do whatever they like once file has ended playback.

### DIFF
--- a/RCTVideo.h
+++ b/RCTVideo.h
@@ -4,6 +4,7 @@ extern NSString *const RNVideoEventLoaded;
 extern NSString *const RNVideoEventLoading;
 extern NSString *const RNVideoEventProgress;
 extern NSString *const RNVideoEventLoadingError;
+extern NSString *const RNVideoEventEnd;
 
 @class RCTEventDispatcher;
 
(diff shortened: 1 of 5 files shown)


mwilc0x:
cc @johanneslumpe

mwilc0x:
Oops only now just saw the 'onComplete' todo item. Let me know if you'd like me to rename.

johanneslumpe:
@mjw56 since the html5 video event is `ended`, maybe we should rename it to `onEnded` (instead of `onComplete`). But let's see what @brentvatne position is on this :)

brentvatne:
@mjw56 @johanneslumpe - I like `onEnded`! once it's renamed I'll merge this! :smile:

brentvatne:
@mjw56 @johanneslumpe - actually scratch that, let's stick with `onEnd` - `ended` seems more like a property of the html video element, and so the name makes sense there, but for a callback `onEnd` seems to work better :+1:

mwilc0x:
:D thanks @brentvatne !

PR #28, opened 2015-04-28
Provide a fallback if the duration cannot be get

### LINKED ISSUE #26: Add the ability to play a live video stream
It would be a nice feature to be able to stream live video with this component.

Currently if I try to play a live stream (e.g. from a Wowza streaming server referencing the stream's `playlist.m3u8` file) it begins to play but then an error is thrown in XCode on line 142 in `node_modules/react-native/React/Executors/RCTContextExecutor.m`:

```
while (kCFRunLoopRunStopped != CFRunLoopRunInMode(kCFRunLoopDefaultMode, [[NSDate distantFuture] timeIntervalSinceReferenceDate], NO)) {
      RCTAssert(NO, @"not reached assertion"); // runloop spun. that's bad.
    }
```

Swatto:
This to avoid to send an errors stack caused by the `NaN` given in the JSON of the RCTEventDispatcher.

fixes #26

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -135,7 +135,7 @@ static NSString *const statusKeyPath = @"status";
   if (object == _playerItem) {
     if (_playerItem.status == AVPlayerItemStatusReadyToPlay) {
       [_eventDispatcher sendInputEventWithName:RNVideoEventLoaded body:@{
-        @"duration": [NSNumber numberWithFloat:CMTimeGetSeconds(_playerItem.duration)],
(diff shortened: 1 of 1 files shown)


brentvatne:
Seems reasonable to me! Thanks!

brentvatne:
Published in `0.3.2` - do you have an example project showing a live stream somewhere? Curious to try it

Swatto:
This easy one that I know is the [hitbox.tv](http://developers.hitbox.tv/) API (Twitch want some credentials) where you can have `http://api.hitbox.tv/player/hls/{{username}}.m3u8` (the account must live stream)

PR #32, opened 2015-05-08
Use the `ViewStylePropTypes` since they work &  inherit from `LayoutPropTypes`

rpastorelle:
Fixes the false-positive for invalid prop types mentioned here: https://github.com/brentvatne/react-native-video/issues/31


brentvatne:
Good idea, thanks!

PR #34, opened 2015-05-11
Update react-native requirement in example

nickhudkins:
react-native-video wants ^0.4.0, so let's give it to it.


brentvatne:
Thank you sir!

PR #50, opened 2015-06-10
isStatic property added in the source object

### LINKED ISSUE #49: Play source outside of Main bundle directory
I have an issue trying to play a video that I recorded, the file is outside of the mainBundle directory and so the player can't of course find the file.
The issue is related to that code:

```
NSURL *url = (isNetwork || isAsset) ?
    [NSURL URLWithString:uri] :
    [[NSURL alloc] initFileURLWithPath:[[NSBundle mainBundle] pathForResource:uri ofType:type]];
```

A solution would be to override the isNetwork property, but unfortunately impossible on the js side, I think the variable is initialized by React.
@brentvatne do you have any idea ? We could maybe create a prop to force the use of URLWithString

Thanks,

maxs15:
In order to be able to specify a path outside of the main bundle directory
close #49

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -151,10 +151,11 @@ static NSString *const statusKeyPath = @"status";
 {
   bool isNetwork = [RCTConvert BOOL:[source objectForKey:@"isNetwork"]];
   bool isAsset = [RCTConvert BOOL:[source objectForKey:@"isAsset"]];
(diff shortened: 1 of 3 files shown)


brentvatne:
Great, now we just have to make it build the correct url according to that boolean

maxs15:
That is building the correct url

brentvatne:
@maxs15 - it looks like https://github.com/brentvatne/react-native-video/pull/54 takes care of this for us, so we can use `file:` paths now!

PR #52, opened 2015-06-10
Add playableDuration property to onProgress event

### LINKED ISSUE #51: Pass playable duration as a parameter on video progress
This can be calculated using AVPlayerItem's loadedTimeRanges property and is a really useful information to have.

isair:
Closes #51

I am new to the React Native and JS world and didn't know how to test this by creating a private node package. Therefore approach with caution.

The method itself is O(n) in time complexity. However, this shouldn't be a problem as the count of loadedTimeRanges should always be negligibly small if I'm not mistaken.

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -80,7 +80,7 @@ static NSString *const statusKeyPath = @"status";
 /*!
  * Calculates and returns the playable duration of the current player item using its loaded time ranges.
  *
- * \returns The playable duration of the current player item.
+ * \returns The playable duration of the current player item in seconds.
  */
 - (NSNumber *)calculatePlayableDuration {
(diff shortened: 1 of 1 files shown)


brentvatne:
Awesome, this looks good to me. cc @johanneslumpe 

@isair - to test something like this I would probably just edit the source file directly from the `node_modules` directory of a project that you have it installed in

isair:
@brentvatne Makes sense, funny how simple solutions can elude us sometimes.

ycg520520:
本地（设备上的）视频获取不到playableDuration的值，这个值始终为0，
如： source={{uri: 'file:///storage/emulated/0/broadchurch.mp4'}}

PR #54, opened 2015-06-14
support for absolute path and file:// paths

machard:
Added support for absolute local paths and file://


brentvatne:
Great, thanks!

PR #55, opened 2015-06-16
Fixes setting properties after changing media source

### LINKED ISSUE #53: Properties not being set after altering media source
I think there is a bug here: https://github.com/brentvatne/react-native-video/blob/master/RCTVideo.m#L133

I'm not sure what the fix is, but it seems that for example, the resizeMode property never gets set on the new instance created. I guess all the properties should be set, but I wasn't sure where this was occurring to be able to do a pull request.

markburns:
Fixes #53

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -39,6 +39,8 @@ static NSString *const statusKeyPath = @"status";
   float _rate;
   BOOL _muted;
   BOOL _paused;
+  BOOL _repeat;
+  NSString * _resizeMode;
 }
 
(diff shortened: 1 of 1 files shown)


brentvatne:
Thanks @markburns! Squashed this into one commit and merged in https://github.com/brentvatne/react-native-video/commit/9723f8849fca2d750885256cd1d466dc51ff0ca8 and pushed to npm in 0.4.4

markburns:
Great thanks

PR #56, opened 2015-06-18
Fix 'Invalid number value (NaN) in JSON write' crash

isair:
I found that switching video sources causes a crash due to a NaN value being passed as playableDuration. Other different cases might lead to a crash because of the same issue as well.


brentvatne:
@isair - thanks for this fix! I haven't had any need for this library recently, so it's great that some people are helping to maintain it

brentvatne:
Released in `0.4.5`

isair:
Happy to contribute!

PR #65, opened 2015-06-26
Remove Notification Center observers on dealloc

ghost:
Removes observers on deallocation of the RCTVideo class in service of `0.4.8`. However, this PR may not be needed if the observers are removed from the `init` method.


brentvatne:
Thanks @jhabdas :smile: - silly me. This is in 0.4.9

Issue #66, opened 2015-06-26
Pausing videos do not always work


isair:
I believe this to be yet another threading issue. I didn't have this issue with the remote videos in my app, but when I added video pre-loading to it using a hacky way (just move video component to background, assign the next video to it, and pause it until it is needed) I realized that videos kept playing even when paused property was correctly being set.

isair:
`setPaused` seems to be called correctly with the right values. When `setPaused` is called with paused as true in my app, the progress update timer is stopped properly but player is not being paused. So it is definitely a threading issue. Will fix this.

Since all the property setters seem to be called asynchronously we should check them all and make sure what needs to be on the main thread stays on the main thread. I'll check and open another issue if necessary.

---

From Apple's documentations:

AVPlayer serializes notifications of changes that occur dynamically during playback on a dispatch queue. By default, this queue is the main queue (see dispatch_get_main_queue). To ensure safe access to a player’s nonatomic properties while dynamic changes in playback state may be reported, you must serialize access with the receiver’s notification queue. In the common case, such serialization is naturally achieved by invoking AVPlayer’s various methods on the main thread or queue.

PR #70, opened 2015-06-29
Fix issues that arise due to accessing _player on a queue other than main

isair:
For more information on why this is being done: https://github.com/brentvatne/react-native-video/issues/66#issuecomment-115884568

Also renamed the method playerDidEnd as playerDidReachEnd.

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -321,13 +321,15 @@ static NSString *const statusKeyPath = @"status";
     CMTime tolerance = CMTimeMake(1000, timeScale);
 
     if (CMTimeCompare(current, cmSeekTime) != 0) {
-      [_player seekToTime:cmSeekTime toleranceBefore:tolerance toleranceAfter:tolerance completionHandler:^(BOOL finished) {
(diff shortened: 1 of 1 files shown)


isair:
@brentvatne I need some clarification on a matter. Based on the fact that calling `[_player pause]` on the main queue has fixed the problem I mentioned in #66, I assumed react-native didn't call property setters just on the main thread. However, if that was the case, then there should have been problems with `setSrc` as well. There is no information in React Native's [native UI component documentation](http://facebook.github.io/react-native/docs/nativecomponentsios.html#content), so I hope maybe you can shed some light on this.

isair:
@brentvatne Merging this as-is without modifying the setSrc method as I haven't ran into a bug about it so far. Could this whole async business be a bug with react-native as it isn't mentioned in the docs?

brentvatne:
Hey @isair sorry for the delay on getting back to you here - I'm actually not sure about whether or not setSrc would need to be called on the main thread. @johanneslumpe do you have any insight on this?

Issue #73, opened 2015-07-08
Initial paused state has no effect


casio:
Setting the `paused` prop to true doesnt have the expected effect.
Instead the video starts playing whatsoever.

gazzer82:
i'm seeing the same issue, any ideas on a fix?

brentvatne:
Not ideal! I wonder what caused this regression. PR with a fix very welcome! :smile:

isair:
Weird, seems like setPaused is called with `false` as its parameter after it is called with `true`.

isair:
Debugged it. The problems is caused by `setRate` being called on `_player` after `[_player pause]` is called.

PR #76, opened 2015-07-10
Call setRate instead of pause/play as they are equivalent

dpeek:
Video was resuming after seek when setRate was called after play/pause in applyModifiers. According
to [Apple's docs](https://developer.apple.com/library/prerelease/ios/documentation/AVFoundation/Reference/AVPlayer_Class/index.html#//apple_ref/occ/instm/AVPlayer/play) play/pause are equivalent to setRate:1.0 and setRate:0.0, meaning that if rate was
1.0 the video resumed after seeking completed.

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -297,12 +297,12 @@ static NSString *const statusKeyPath = @"status";
   if (paused) {
     [self stopProgressTimer];
     dispatch_async(dispatch_get_main_queue(), ^{
-      [_player pause];
+      [_player setRate:0.0];
     });
   } else {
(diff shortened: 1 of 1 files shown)


brentvatne:
This makes sense to me, @johanneslumpe what do you think?

isair:
Seems logical for me as well. Haven't tested but sounds like this might fix #73 too.

Edit: By the way, @dpeek, the `applicationWillResignActive` implementation needs to be updated for these changes as well.

Edit 2: @brentvatne I can confirm that this fixes #73, at least in my project.

PR #77, opened 2015-07-10
Update for 0.7.1 release

christopherdro:


brentvatne:
tks tks

isair:
@brentvatne So I guess calling dispatch_async inside our methods is no longer needed.

brentvatne:
@isair - truth, this will happen by default now, good catch

PR #78, opened 2015-07-14
Disables default system animation when resizing AVPlayerLayer

dpeek:
Not sure if this something you will want – perhaps it could go behind a flag. When animating video size, the default system animation makes the rendered size of the video lag behind the bounds of the react view. This patch disables the default animation, ensuring the video renders at the correct size.


brentvatne:
Seems reasonable to me, if you want to submit a follow up PR to make this a flag that would be great. Thanks @dpeek!

PR #84, opened 2015-08-08
video breaks on react-native 0.9.0-rc

jim-lake:
You can't modify this.props on 0.9.0-rc, so you just need to create a new object, not modify this.props to work correctly on new versions.  Also removes a bunch of warnings on 0.8.0.


brentvatne:
Thanks :)

brentvatne:
Published to npm in 0.5.1

PR #92, opened 2015-09-16
Use lodash instead of merge, unbreaking the component (for us)

jcfrancisco:
Not sure if others are experiencing this, but a few of us (@emibob, @ilanasufrin, @benmidi & I) could not get any videos to play at all (iOS, React 0.10).

We tracked it down to this line in `Video.ios.js` that was using the `merge` library. For whatever reason, the third param was being ignored. Using underscore/lodash's `extend` method fixes it for us.


brentvatne:
Works for me, will push update to npm momentarily!

brentvatne:
@jcfrancisco - ok this is on npm with 0.6.1

jcfrancisco:
@brentvatne Thanks!

PR #105, opened 2015-11-11
fix "sometimes video cannot stop correctly"

guodong000:


brentvatne:
Interesting, sorry I never went over this, @guodong000 do you think this is still relevant? If so, can you rebase?

mattapperson:
This PR will be closed by the end of the week if you don't currently have the time to rebase. You would be welcome to re-open at your convenience

PR #108, opened 2015-11-25
Passes onError and onLoadStart through

### LINKED ISSUE #94: onVideoError and onVideoLoadStart native properties are unset
It looks like the `onVideoError` and `onVideoLoadStart` properties are not being [set as nativeProps](https://github.com/brentvatne/react-native-video/blob/3aa46875a58b857861377dc1528088ae027627e2/Video.ios.js#L83-L95).

Should they be set to `this._onError` and `this._onLoadStart`?

cancan101:
closes #94

### DIFF
--- a/Video.ios.js
+++ b/Video.ios.js
@@ -92,6 +92,8 @@ var Video = React.createClass({
       onVideoLoad: this._onLoad,
       onVideoProgress: this._onProgress,
       onVideoEnd: this._onEnd,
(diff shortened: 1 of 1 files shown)


rclai:
Ah.. I was wondering why those weren't working. Can this get merged in?

brentvatne:
I think this is fixed on master now :)

PR #124, opened 2016-01-07
allow use external storage resource

buhe:


mattapperson:
@buhe I would like to get this merged ASAP. If you could just rebase then we will get right on that! (Sorry it sat on the shelf so long 😞)

mattapperson:
This PR will be closed by the end of the week if you don't currently have the time to rebase. You would be welcome to re-open at your convenience

PR #126, opened 2016-01-08
Added isAsset same as iOS implementation

jpgarcia:
I've added the isAsset field as it did not work as expected when providing an uri like "/data/data/..."


isair:
@jpgarcia Thanks for your contribution. :+1:

PR #138, opened 2016-02-02
Fix the android version, and upgrade to react native 0.19

tal952:
according to https://github.com/facebook/react-native/issues/5649, we need to use:
import com.facebook.react.uimanager.annotations.ReactProp;
and not 
import com.facebook.react.uimanager.ReactProp.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -16,6 +16,6 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


isair:
@tal952 Thanks for your contribution. :thumbsup:

PR #149, opened 2016-02-10
Examples working on android/ios again

### LINKED ISSUE #144: VideoPlayer example is broken (again)
#### Steps:

```
# start fresh...
git clone https://github.com/brentvatne/react-native-video.git
cd react-native-video
npm install
cd Examples/VideoPlayer/
npm install
open VideoPlayer.xcodeproj
# then hit build/run
```
#### Result:

screen shot 2016-02-05 at 5 25 55 pm

corbt:
The examples are currently broken on iOS and Android because the library was recently upgraded to ES6 modules and the example still used the `require` syntax without specifying the default import.

I ended up upgrading the example to ES6 entirely (because why not) and updating it to use React Native 0.19. The examples are now working for both Android and iOS.

Fixes #144.

### DIFF
--- a/Examples/VideoPlayer/.flowconfig
+++ b/Examples/VideoPlayer/.flowconfig
@@ -14,17 +14,19 @@
 
 # Ignore react and fbjs where there are overlaps, but don't ignore
 # anything that react-native relies on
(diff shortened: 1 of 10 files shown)


isair:
Thanks for your contribution!

PR #153, opened 2016-02-19
Fixes bug making it impossible to pause video from controls

oblador:
When using `controls={true}` is basically impossible to pause with the pause button since the `playbackLikelyToKeepUp` event is called just after. This PR adds a listener to see if the video was paused due to empty buffer and only then will it play the video again when the `playbackLikelyToKeepUp` event is triggered.


isair:
LGTM. Thanks for your contribution!

xing-zheng:
Thank you! @oblador

PR #158, opened 2016-02-28
fixed syntax error

ejb1123:


xing-zheng:
@isair 
Yes, There is a typo in document, It should be "StyleSheet" but "Stylesheet".

Thank you! @ejb1123

brentvatne:
thanks ;)

PR #162, opened 2016-03-09
Keep the player valid after playing completion

xing-zheng:
@brentvatne @isair 
If we keep the player valid (mMediaPlayerValid == true), we can seek to start position and replay the video after playing completion.

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactVideoView.java
+++ b/android/src/main/java/com/brentvatne/react/ReactVideoView.java
@@ -270,7 +270,6 @@ public class ReactVideoView extends ScalableVideoView implements MediaPlayer.OnP
 
     @Override
     public void onCompletion(MediaPlayer mp) {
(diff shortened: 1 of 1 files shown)


brentvatne:
@isair - not sure what the best thing to do here is, your call!

isair:
Been swamped for a while now, I'm going to have to take a detailed look later at these Android related PRs to ensure nothing is getting broken. Can't verify these changes with one look as I'm not an Android developer at all; just read some documentations here and there to put our implementation together.

The proposed change here may cause a problem with the progress updating code [here](https://github.com/xing-zheng/react-native-video/blob/patch-1/android/src/main/java/com/brentvatne/react/ReactVideoView.java#L86). If that's the case, just modifying `seekTo` might be a better idea to enable seeking after completion.

xing-zheng:
Thanks for your review, @isair

PR #166, opened 2016-03-12
support content:/ uri in android

tianjianchn:


brentvatne:
Seems reasonable to me!

PR #181, opened 2016-04-07
Support fullscreen mode

sjchmiela:
Added `fullscreen` property triggering fullscreen player. (`true` — show, `false` — dismiss)

Also, added callbacks regarding it's visibility.

```
onFullscreenPlayerWillPresent: PropTypes.func,
onFullscreenPlayerDidPresent: PropTypes.func,
onFullscreenPlayerWillDismiss: PropTypes.func,
onFullscreenPlayerDidDismiss: PropTypes.func,
```

### DIFF
--- a/RCTVideo.h
+++ b/RCTVideo.h
@@ -1,6 +1,7 @@
 #import "RCTView.h"
 #import <AVFoundation/AVFoundation.h>
 #import "AVKit/AVKit.h"
+#import "UIView+FindUIViewController.h"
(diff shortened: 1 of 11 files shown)


jallens:
@sjchmiela just only work on iOS?

sjchmiela:
Unfortunately yes.

brentvatne:
@sjchmiela - looks good, thanks!

PR #182, opened 2016-04-08
[Podspec] Set podspec version from the package.json and fix git source in the Podspec

negativetwelve:
React Native recently started reading the version from the package.json file which I think is a great addition. I also fixed the git source link (was erroring before) and added the git tag (cocoapods recommends setting one and I agree).

@brentvatne I added the author field since it was required by cocoapods, let me know if you want different text in there (I used the same field from the package.json, maybe we should just grab this info from there?)

lastly, the platform was set at 8.0, I was able to run this on our project which has a 7.0 deployment target so I lowered the platform value here, let me know if any of the implementation uses iOS 8 specific APIs.

thanks!

### DIFF
--- a/react-native-video.podspec
+++ b/react-native-video.podspec
@@ -3,12 +3,12 @@ Pod::Spec.new do |s|
   s.version      = "0.7.1"
   s.summary      = "A <Video /> element for react-native"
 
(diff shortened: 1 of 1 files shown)


brentvatne:
Looks great, thanks!

PR #201, opened 2016-04-29
README: prevent blocking outside music

skleest:


brentvatne:
Good stuff, thanks! :)

PR #233, opened 2016-05-31
Add original video width/height for android

jalbrecht:
Added "originalWidth" and "originalHeight" values to onLoad event for android

https://developer.android.com/reference/android/media/MediaPlayer.html#getVideoWidth()
https://developer.android.com/reference/android/media/MediaPlayer.html#getVideoHeight()


bcalik:
This should not be merged because it does not match the response in iOS version.
(https://github.com/brentvatne/react-native-video/commit/e7cae171fba3e8c82fb45545187a0a54ccacaaea and https://github.com/brentvatne/react-native-video/commit/91abfbde8b519e88d7eca71f9503d6a5311dc265).

Here is the proper PR:
https://github.com/brentvatne/react-native-video/pull/239

PR #236, opened 2016-06-02
Added possibility to play video in background and when app is inactive.

thoblr:
When streaming live video it's sometimes desirable to let the video continue to play when app is entering background. This PR makes it possible to configure the video component to allow this.

Prior to this PR iOS stops playing video when entering background but Android contiues to play video. This PR unifies the behaviour between iOS and Android so that when `playInBackground` is set to `true` both platforms will continue to play video and naturally `false` will make both platforms stop the video.

Also added a iOS specific prop called `playWhenInactive` which prevents the video to stop if control center or notification center is shown. Again when streaming live video you don't want to stop the playback when for example using control center to change volume.


sjchmiela:
Great, thanks!

crysfel:
Thanks for this!! In what version number is this code included? Or is it in master only? @brentvatne It would be nice to create a release 👍

Monte47:
Hey so is there a way to have the video resume when the control panel is closed on IOS?

PR #239, opened 2016-06-04
[Android] naturalSize callback to match iOS version.

bcalik:
onLoad event now returns naturalSize object containing width, height and orientation on Android.
Closes https://github.com/brentvatne/react-native-video/issues/203


sjchmiela:
Thanks!

PR #244, opened 2016-06-15
Fix native modules

orktes:


orktes:
Sorry was supposed to be a PR to our forks master :)

PR #252, opened 2016-06-22
replace bind statements

mroswald:
In order to reduce code and get more overview, I replaced the bind statements from constructor and removed the constructor itself.


brentvatne:
Thanks, this is much better

PR #258, opened 2016-06-28
This is a bug, you don't need to call the local `setPaused(false)`, a…

nemothecapt:
This is a bug, , it would not pause the video on iOS.

You don't need to call the local `setPaused(false)`, as it trigger to play the video again when paused from the native controllers. It always go to the `else` section of `setPaused()`, because `_paused` was never set to `true`

The video automatically Pause or continue to Play after it hit a stall and ready to play again

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -348,10 +348,6 @@ static NSString *const playbackRate = @"rate";
     } else if ([keyPath isEqualToString:playbackBufferEmptyKeyPath]) {
       _playerBufferEmpty = YES;
     } else if ([keyPath isEqualToString:playbackLikelyToKeepUpKeyPath]) {
-      // Continue playing (or not if paused) after being paused due to hitting an unbuffered zone.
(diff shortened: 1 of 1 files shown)


mattapperson:
We will get this reviewed this week @nemothecapt

RCTVideo.m
```diff
@@ -348,10 +348,6 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
     } else if ([keyPath isEqualToString:playbackBufferEmptyKeyPath]) {
       _playerBufferEmpty = YES;
     } else if ([keyPath isEqualToString:playbackLikelyToKeepUpKeyPath]) {
-      // Continue playing (or not if paused) after being paused due to hitting an unbuffered zone.
-      if ((!_controls || _playerBufferEmpty) && _playerItem.playbackLikelyToKeepUp) {
```
mattapperson, line 352: I am not sure I understand why this is an issue if native controls pause the video, there is a check for if there are no controls... can you please try explaining in other terms?

mattapperson:
Closing in favor of #325

PR #260, opened 2016-06-30
send progress events only while playing or buffering, resume after host resumes

madamyk:


mattapperson:
This is needing a rebase @whitecurrant would you mind? after that we will get right on a review

mattapperson:
This PR will be closed by the end of the week if you dont currently have the time to rebase. You would be welcome to re-open at your convenience

PR #285, opened 2016-08-02
add seekableDuration

eschan:
When playing a live HLS videos, duration of the video is returning indefinite.  seekableDuration will give you the beginning and up most bound of the live video.  The seekable duration time will increase as the live stream continues, and stop when the video source is done live streaming.

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -110,6 +110,17 @@ static NSString *const playbackRate = @"rate";
     return(kCMTimeInvalid);
 }
 
+- (CMTimeRange)playerItemSeekableTimeRange
+{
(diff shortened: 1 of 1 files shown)


mattapperson:
Interesting PR. No matching android API though is there? or even can there be?

eschan:
I don't think android supports HLS out of the box, you might need a specific HLS player.

mattapperson:
@eschan Just did some quick research, HLS does exist on Android but it sucks...

So we can put that aside and make this iOS only, we just need to document the property. Would you mind doing that? or would you like me to add a help wanted label and see if I can grab someone else to do that part?

mattapperson:
merged and adding prop in another PR

PR #291, opened 2016-08-04
Added deallocation of observers

Cordazar:
Added deallocation of observers to prevent an NSInternalInconsistencyException.

Example of exception: 

``` obj-c
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x7f853b14e7b0 of class AVPlayer was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x7f853b1552c0> (
<NSKeyValueObservance 0x7f853b2aadd0: Observer: 0x7f853b1a20f0, Key path: rate, Options: <New: NO, Old: NO, Prior: NO> Context: 0x0, Property: 0x7f853b286f50>
)'
```

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -106,7 +106,7 @@ static NSString *const playbackRate = @"rate";
     {
         return([playerItem duration]);
     }
-    
+
     return(kCMTimeInvalid);
 }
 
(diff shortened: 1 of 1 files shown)


mattapperson:
Thanks for this @Cordazar! I was just about to implement this myself!

PR #305, opened 2016-08-24
prevent 'cover' resizeMode animation

nemophrost:
Prevent ‘cover’ resizeMode from animating the video to the correct size (when component is created).

### DIFF
--- a/RCTVideo.m
+++ b/RCTVideo.m
@@ -578,6 +578,9 @@ static NSString *const playbackRate = @"rate";
     if( _player )
     {
         _playerViewController = [self createPlayerViewController:_player withPlayerItem:_playerItem];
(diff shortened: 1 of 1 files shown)


rclai:
This would be great.

What's interesting is that I experience this not when the component is being created but when I change the `source` from one to another.

Before I found your PR, I changed the `setSrc` implementation to not tear down the `_playerLayer` but rather reassign the `player` property of `_playerLayer` to the newly created `_player` if a `_playerLayer` already exists.

mattapperson:
@nemophrost seems there are some merge conflicts, would you mind rebasing?

nemophrost:
See #307 for rebased PR

PR #307, opened 2016-08-29
prevent 'cover' resizeMode animation

nemophrost:
Prevent ‘cover’ resizeMode from animating the video to the correct size (when component is created).


mattapperson:
Thanks!

PR #327, opened 2016-09-15
fix typo

mvayngrib:


skv-headless:
Please merge it. Even example fails with compilation error

PR #329, opened 2016-09-17
upgrade example to rn33

skv-headless:
working on iOS.
Fails on android same way as in master for me.


skv-headless:
Such an old react native version doesn't compile in xcode version.

PR #381, opened 2016-11-16
Replace use of RCTBridge with RCTBubblingEventBlock

jameshartt:
Thanks to https://github.com/facebook/react-native/commit/aa4428cd132bb0d0dbc950b66d3b5f2a3c5b9322 coming up in the 0.38.0 release of RN, the way that the Obj-C code sends events needs to change, currently whilst using the 0.38.0-rc0 it has broken all the Javascript -> Native events on iOS. This means no progress or video information gets received.

This is a working solution that we have implemented on a fork of ours, we had to add extra native only props to the Javascript side and haven't fully tested on the other platforms (Android never had the problem and now you've merged Windows support, we don't even have the resources to test that), although we will doing a few Android changes over the next couple of days.

### DIFF
--- a/ios/RCTVideoManager.m
+++ b/ios/RCTVideoManager.m
@@ -14,28 +14,6 @@ RCT_EXPORT_MODULE();
   return [[RCTVideo alloc] initWithEventDispatcher:self.bridge.eventDispatcher];
 }
 
-/* Should support: onLoadStart, onLoad, and onError to stay consistent with Image */
-
-- (NSArray *)customDirectEventTypes
-{
-  return @[
-    @"onVideoLoadStart",
-    @"onVideoLoad",
-    @"onVideoError",
-    @"onVideoProgress",
-    @"onVideoSeek",
-    @"onVideoEnd",
-    @"onVideoFullscreenPlayerWillPresent",
-    @"onVideoFullscreenPlayerDidPresent",
-    @"onVideoFullscreenPlayerWillDismiss",
-    @"onVideoFullscreenPlayerDidDismiss",
-    @"onReadyForDisplay",
-    @"onPlaybackStalled",
-    @"onPlaybackResume",
-    @"onPlaybackRateChange"
-  ];
-}
-
 - (dispatch_queue_t)methodQueue
 {
     return dispatch_get_main_queue();
@@ -55,6 +33,21 @@ RCT_EXPORT_VIEW_PROPERTY(seek, float);
 RCT_EXPORT_VIEW_PROPERTY(currentTime, float);
 RCT_EXPORT_VIEW_PROPERTY(fullscreen, BOOL);
 RCT_EXPORT_VIEW_PROPERTY(progressUpdateInterval, float);
+/* Should support: onLoadStart, onLoad, and onError to stay consistent with Image */
+RCT_EXPORT_VIEW_PROPERTY(onVideoLoadStart, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoLoad, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoError, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoProgress, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoSeek, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoEnd, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerWillPresent, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerDidPresent, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerWillDismiss, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerDidDismiss, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onReadyForDisplay, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackStalled, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackResume, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackRateChange, RCTBubblingEventBlock);
 
 - (NSDictionary *)constantsToExport
 {
--- a/Video.js
+++ b/Video.js
@@ -180,6 +180,16 @@ Video.propTypes = {
   src: PropTypes.object,
   seek: PropTypes.number,
   fullscreen: PropTypes.bool,
+  onVideoLoadStart: PropTypes.func,
+  onVideoLoad: PropTypes.func,
+  onVideoError: PropTypes.func,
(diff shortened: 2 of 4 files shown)


larryranches:
Great work! When can this be merged into master?

carleryd:
I am getting a runtime exception trying to load a video with this commit.
```
  } else if (object == _player) {
      if([keyPath isEqualToString:playbackRate]) {
          self.onPlaybackRateChange(@{@"playbackRate": [NSNumber numberWithFloat:_player.rate], // Thread 1: EXC_BAD_ACCESS (code=1, address=0xc)
                                      @"target": self.reactTag});
          if(_playbackStalled && _player.rate > 0) {
              self.onPlaybackResume(@{@"playbackRate": [NSNumber numberWithFloat:_player.rate],
                                      @"target": self.reactTag});
              _playbackStalled = NO;
          }
      }
```

Any ideas? I had problems with onLoad() and onLoadStart() not triggering on react-native@0.38

billdozr:
Got a same problem as @carleryd above on iPhone 4s (iOS 8.4) simulator.

larryranches:
maybe this is not ready? I wrote up an issue on the callback function not triggering #401

MateuszMxxxx:
I have same problem as @carleryd also. I've commented one line (this faulty one) to check if this fix the events, and it does :)
I hope that some contributor will look into that problem soon :)

mattapperson:
@jameshartt I am seeing the same issue as @carleryd, are you ale to look into this?

jameshartt:
Apologies, been away on long weekend. Will have a look now, this was cherry picked from other commits, so could have missed something.

jameshartt:
The example project works just fine for me. I'm thinking it's an iOS8 issue. Our app doesn't support it, so just downloading the sim to have a look.

jameshartt:
Still no issues in the iOS8.4 simulator. Are you running the example project with my changes? I don't know what to suggest or how to move forward. This works for me!

XHTeng:
It works , thank you !

pawlowskim:
`
      if([keyPath isEqualToString:playbackRate]) {
          self.onPlaybackRateChange(@{@"playbackRate": (...)
`

Before calling any callback, we should check if it isn't nil. Otherwise, this can crash application.

Just add nil check before every callback:

` if([keyPath isEqualToString:playbackRate] && self.onPlaybackRateChange != nil) {
          self.onPlaybackRateChange(@{@"playbackRate": (...)`

jameshartt:
The thing is, this defensive programming shouldn't be needed. This callback __should__ exist. This could be hiding another issue.

jameshartt:
I could be missing something with how RN works with these Bubbling event blocks. Do they recommend always checking if they are nil?

larryranches:
Our Continuous Integration suddenly broke with this PR but it works fine with the current version in npm. Any ideas? 

❌  ld: library not found for -lRCTVideo

❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)

jameshartt:
Going to need more information. What project are you building? The example? Is you CI accessible publicly? I can't seem to see it.

I've not seen these issues before.

larryranches:
No, unfortunately our CI is not accessible publicly. I was thinking that something didn't get linked properly within the RCTVideo library or something?

ios/RCTVideoManager.m
```diff
@@ -55,6 +33,21 @@ - (dispatch_queue_t)methodQueue
...
+RCT_EXPORT_VIEW_PROPERTY(onVideoSeek, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoEnd, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerWillPresent, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerDidPresent, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerWillDismiss, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onVideoFullscreenPlayerDidDismiss, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onReadyForDisplay, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackStalled, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackResume, RCTBubblingEventBlock);
+RCT_EXPORT_VIEW_PROPERTY(onPlaybackRateChange, RCTBubblingEventBlock);
```
jasonfma, line 50: What's the difference between using `RCTBubblingEventBlock` and `RCTDirectEventBlock`? I'm just curious because their docs say to use the `RCTDirectEventBlock`.

https://github.com/facebook/react-native/wiki/Breaking-Changes#038

Thanks for doing this btw. I can't wait for this fix to get in.

jameshartt:
@larryranches The main Xcode project never built for me, as it's not properly linked with anything RN. I thought this was a Node Module thing and looked at it no further. I'm not sure how I can help without making radical changes to get the project to build.

ios/RCTVideoManager.m
jameshartt: You make a good point, the docs really aren't clear (they don't say at all) on the difference between the two. They basically behave as the same thing. Due to the nature of the original calls, `RCTBubblingEventBlock` seemed like the one to call. But I was never 100% sure. But hey, it works. I understand most issues disguised as questions get batted away, but might be worth trying to resurrect something like this. https://github.com/facebook/react-native/issues/11148

jameshartt: I'm happy to change it if it still works. I don't remember experimenting with `RCTDirectEventBlock`

larryranches:
@jameshartt Thanks for letting me, might just be our CI system as the project does build in local xcodeproj. Any ideas when we can get this PR into master?

hellogerard:
🎉 👍

hellogerard:
I also got the above runtime exception on the `onPlaybackRateChange` line. I found that it seems to be caused by using the `playInBackground` option. If set that to true: crash. If I take it out, everything works. I will open a new issue.

UpCoder:
I also meet the question. onProgress onLoad onEnd all don't trigger. But in the demo is right. the code as follow:
```
<Video
                        source={{uri: "broadchurch"}}
                        style={styles.fullScreen}
                        rate={this.state.rate}
                        paused={this.state.paused}
                        volume={this.state.volume}
                        muted={this.state.muted}
                        resizeMode={this.state.resizeMode}
                        onLoad={this.onLoad}
                        onProgress={this.onProgress}
                        onEnd={() => { AlertIOS.alert('Done!') }}
                        repeat={false}
                    />
```

smkhalsa:
Anyone discover a fix for the `onPlaybackRateChange` bug? @hellogerard I'm getting the runtime error on the iPhone 5 simultator even without `playInBackground`

jackyun1024:
I make forked repository for applying suggestion of @pawlowskim.

It works for me.

https://github.com/younseunghyun/react-native-video/tree/0.38-nill

ios/RCTVideoManager.m
bishalg: Considering the Code on "**RCTSlider.h**"
https://github.com/facebook/react-native/blob/master/React/Views/RCTSlider.h
```
@property (nonatomic, copy) RCTBubblingEventBlock onValueChange;
@property (nonatomic, copy) RCTDirectEventBlock onSlidingComplete;
```

The `RCTBubblingEventBlock` is intended for things that has multiple callbacks possible like - "Value Changed" and will likely fire multiple changes.

But again even if `RCTDirectEventBlock` has same signature it is for single event callback for event like - "Completed".

PR #373, opened 2016-11-09
feat(Windows): Adds Windows support to react-native-video

rozele:
This PR adds react-native-windows support to react-native-video.  The Video component is implemented with a XAML MediaPlayerElement.  Most of the features implemented by Android (and some additional ones) are implemented by Windows.

Known issues and missing features include:
* onReadyForDisplay event
* local URI convention (e.g., "broadchurch" in examples changed to require("./broadchurch.mp4")
* `playableDuration` in `onVideoProgress` event is always 0.0
* `playInBackground` is not yet supported
* Volume settings are applied, but the UWP control does not handle it properly

PR #385, opened 2016-11-17
Initial WPF Support

kevinvangelder:
#373 added UWP (Windows 8.1/10) support. This PR adds WPF (Windows 7) support.

There are a couple small gaps (resize mode and looped videos not stopping on JS reload are probably the biggest ones) but the majority of the features are functioning.


kevinvangelder:
@mattapperson Any feedback or questions about this?

mattapperson:
@kevinvangelder sorry for the slow reply, this is really great though! Give me a day or 2 to review and ill get bback to you

kevinvangelder:
No problem. I just remembered that I referenced a local copy of the ReactNative.Net46.csproj because we haven't yet released it in a node module. This will cause errors if you try to load it in Visual Studio, so you probably want to hold off on reviewing until we get that released and I update this branch.

mattapperson:
@kevinvangelder just checking in on that local copy of ReactNative.Net46.csproj

kevinvangelder:
@mattapperson Thanks for following up. I'm hoping we'll have an initial release of react-native-wpf at some point this week (so far, just a placeholder https://www.npmjs.com/package/react-native-wpf) and then it will just take me 5 minutes update to this PR and run everything again to verify it still works. :)

mattapperson:
@kevinvangelder is this ready to merge?

kevinvangelder:
Yeah, it should be good to go.

alixeb:
any plan to merge this soon. we are currently building a Windows 10 UWP app.
We have tried to integrate against master branch, but no luck
Or perhaps it should be working already for Windows 10?

mattapperson:
@alixeb thanks for the bump. I missed @kevinvangelder's reply

kevinvangelder:
@alixeb Feel free to ping me in an issue if you run into problems implementing this.

alixeb:
@kevinvangelder do you know what video format is supported on Windows 10?
It looks like we have sound, but no image.
we have raised an issue here ==> https://github.com/react-native-community/react-native-video/issues/548
we will try to provide more details soon

kevinvangelder:
@alixeb If you haven't already, try setting an explicit width and height. On Windows images and videos seem to collapse when only using flexbox.

lznt:
I tried also this, and was not able to get the Video running, I have my on hack implemented that uses MediaPlayer, it plays but of course i'd rather use this solution as it is much more consistent with iOS and Android. 

All I get is a black screen, even after setting width and height explicit. Also no errors, it seems that the videoplayer just does not render at all.

lznt:
I also tried the example and after npm installing packager crashes on; 

This error is caused by a @providesModule declaration with the same name across two different files.
    at HasteMap._updateHasteMap (c:/react-native-video/example/node_modules/react-native/packager/src/node-haste/DependencyGraph/HasteMap.js:158:13)
    at p.getName.then.name (c:/react-native-video/example/node_modules/react-native/packager/src/node-haste/DependencyGraph/HasteMap.js:133:31)

Is RNWindows on purpose a downloadable dependency for rnvideo?

alixeb:
@lznt what solved it for me, is to change the target in the property of your project.
In Visual Studio: Properties of your project ==> need Anniversary Edition

screenshot 154

asgvard:
I'm keep getting an errors from RN packager about @providesModule too after updating to the version with WPF support.
Why is "react-native-windows" is a mandatory dependency now?
Seems it's not only results in collisions like this:
```
This warning is caused by a @providesModule declaration with the same name across two different files.
jest-haste-map: @providesModule naming collision:
  Duplicate module name: WebSocket
  Paths: .../node_modules/react-native-windows/Libraries/WebSocket/WebSocket.windows.js collides with .../node_modules/react-native/Libraries/WebSocket/WebSocket.js
```
But also causes RN Windows to override normal RN .js files, which results in for example error on iOS about wrong props for StatusBar like "light-content", because for RN Windows it's only "default" or "dark-content". Which means it actually serves RN windows .js files instead of normal ones even for iOS. I'm not saying it's necessary a problem with RN Video package, but rather with the combination of RN and RNWindows, which introduced by adding "react-native-windows" as a mandatory dependency here, which is kind of undesirable for those who are not using Windows support at all. For now using 1.0.0 of RN Video is fine, because it doesn't have this RN Windows dep, but for future versions it will cause problems mentioned above.

RN Version is 0.44 by the way.

kevinvangelder:
The Windows files should only override the iOS files when you're on a Windows platform, and if that's not the case you should probably file an issue on that repo.

Unfortunately, due to the less than ideal way Facebook has decided to implement platform support, having the dependency is necessary in order to compile any given package's Windows project. The extra install size makes me die a little inside each time, but unfortunately we haven't been able to reduce it to a dev dependency.

rozele:
This sounds like a bug in http://github.com/facebook/metro-bundler.

gpeal:
@kevinvangelder This is a huge dependency to ad to react-native-video.
react-native-windows has very specific react/react-native dependencies and now using react-native 0.46 is impossible to use with react-native-video because of the transitive deps of react-native-windows.

gpeal:
@kevinvangelder Also relevant is that the only 0.46 tagged release of react-native-windows has a dependency of 0.46-rc not ^0.46

PR #403, opened 2016-12-02
issue 310 - Added a new property, ignoreSilentSwitch. 

duhseekoh:
When true, audio will play even when the silent switch on an iOS device is set to silent. When false, the audio will toggle with the silent switch. Sets the AVAudioSession to either playback or ambient.

Essentially, this... https://github.com/react-native-community/react-native-video/issues/310

If you guys are good with this I'll update the example project and readme.


mattapperson:
I like this, but could you please update the readme noting this as an iOS only feature?

duhseekoh:
Yea, I'll add to the readme and update the example project in the next couple of days.

mattapperson:
Thanks! You rock!

duhseekoh:
Made the example app and readme changes: 

image

duhseekoh:
Just realized, no matter what, this is going to take over the audio session settings, even if the property isn't set.

Hmm. It might make sense to have an enum setting instead of the boolean value.

Something like:
```
<Video ... audioMode= 'ambient' | 'playback'
// instead of
<Video ... ignoreSilentSwitch = true | false
```

That way if audioMode isn't set, then it won't take over apps audio just because a video exists in the app.

SemonCat:
This feature is useful,

any reason not merge it?

mattapperson:
@SemonCat because as @duhseekoh pointed out, another API might be better

larryranches:
@duhseekoh Great stuff!  Do you have any updates on changing the ignoreSilentSwitch bool to a Api that accepts an enum?

alexanderjarvis:
I'm also looking at something similar to this. What this doesn't handle is multiple videos.

GeoffreyPlitt:
Any update on this?

GeoffreyPlitt:
My company has created a $200 bounty for this PR, hoping someone can fix the issues soon. https://www.bountysource.com/issues/37409911-add-playwhilemuted-property

bsiddiqui:
@duhseekoh are you still interested in merging this? If not, I'm happy to update the api

duhseekoh:
Yea looking back at this I'm good with merging it. We can't have this
library be the end all be all for sound management for an application, so a
boolean property to ignore the silent switch makes sense.

That being said I think there is a future here that handles sound
management for multiple videos on screen at one time, multiple sound
options, etc... it just depends on the level of sophistication we want in
this package and how much control we give through its API (preferably a
limited surface area).

alexanderjarvis:
I created a version of this that handles multiple mounted video components and only stops the audio session when they are all unmounted: https://github.com/drivetribe/react-native-video/blob/fork/ios/RCTVideoAudio.m

It may not be exactly what you need but perhaps by posting it here it helps someone.

larryranches:
@duhseekoh @bsiddiqui Are you guys planning on merging this PR soon?

duhseekoh:
Updates: Changed ignoreSilentSwitch to accept two string values, ignore and obey.
This accounts for the case where the user does not want to modify the audio session from a particular video instance. The user would not use the ignoreSilentSwitch property at all in that case.
Also, the audio session will only be updated when the video is unpaused, instead of whenever the video component has updated props. This allows for multiple videos to be on the screen, with the most recent video unpaused (aka played) being the video that has control over the audio session.

Would be great to have someone pull this down / into their project and test it.

larryranches:
@duhseekoh Just tested this out and it's working perfectly in my project 👍  Thanks! Lets get this PR merged :)

duhseekoh:
Great. @mattapperson - Anything else needed for a merge?

akhan118:
> Updates: Changed ignoreSilentSwitch to accept two string values, ignore and obey.
> This accounts for the case where the user does not want to modify the audio session from a particular video instance. The user would not use the ignoreSilentSwitch property at all in that case.
> **Also, the audio session will only be updated when the video is unpaused, instead of whenever the video component has updated props.** This allows for multiple videos to be on the screen, with the most recent video unpaused (aka played) being the video that has control over the audio session.
> 
> Would be great to have someone pull this down / into their project and test it.

Thank you for working on this feature , I have it a set it to where the video is paused as soon as the video is loaded and I'm observing that `ignoreSilentSwitch="ignore"` is not being evaluated as described above.

 However If I removed ` paused={true} ` , `ignoreSilentSwitch="ignore"`  works as intended. 

How can I get `ignoreSilentSwitch="ignore"`  to work when using  ` paused={true} `  ?

example

```
 <Video
 source={{uri: videoUrl}}
  controls={true}
  ignoreSilentSwitch="ignore"
  fullscreenAutorotate={true}
  paused={true}
 />
```


Thank you

PR #408, opened 2016-12-06
Update react-native-video.podspec so source_files points to ios/ dir

gilbox:
When directory structure was refactored, source files were moved to `ios/` dir, but the podspec wasn't updated to reflect this.


mattapperson:
good catch! Thanks!

PR #411, opened 2016-12-08
employ onPlaybackRateChange instead to smooth the poster disappearance

linmic:
This is to enhance the user experience by smoothing the poster disappearance. With `onProgress` it could lead to unexpected video flash/stall, and with `onPlaybackRateChange` it works much better.

This is related to my previous PR #400.

@mattapperson Could you please help to review this? Thanks.

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -58,10 +58,6 @@ export default class Video extends Component {
   };
 
   _onProgress = (event) => {
-    if (this.state.showPoster) {
-      this.setState({showPoster: false});
-    }
-
     if (this.props.onProgress) {
       this.props.onProgress(event.nativeEvent);
     }
@@ -126,6 +122,10 @@ export default class Video extends Component {
   };
 
   _onPlaybackRateChange = (event) => {
+    if (this.state.showPoster && (event.nativeEvent.playbackRate === 1)) {
+      this.setState({showPoster: false});
+    }
+
     if (this.props.onPlaybackRateChange) {
       this.props.onPlaybackRateChange(event.nativeEvent);
     }


linmic:
This is a super fast merge. Thank you @mattapperson!

mattapperson:
NP, it was an easy one to review

mattapperson:
Though I just thought... what happens with this code if event.nativeEvent.playbackRate is not 1 and this.state.showPoster is true... the poster would never be removed right?

mattapperson:
perhaps a warning or something in the console should be added?

linmic:
I thought if the video is being played by any means, the `event.nativeEvent.playbackRate` would be surely changed from `0` to `1`? no?

Maybe I can change it to `if (this.state.showPoster && (event.nativeEvent.playbackRate !== 0))`, would this be better?

It would be great to have an event say `onPlay` to do the trick, but we don't have it by now. So I guess this is the best way to go?

mattapperson:
Your right, my bad... working with a custom version of the module at the moment where we have slow-mo assigned to playbackRate.

linmic:
@mattapperson I've created another PR #412, I think it would be better to make this change to make sure poster will be removed at any rate other than 0. Thank you for reminding!

rayronvictor:
`_onPlaybackRateChange` is never fired on android

rayronvictor:
playbackRateChange event [is not implemented](https://github.com/react-native-community/react-native-video/blob/737d348ff0501fc405d1ea62cd36d0bd46d70da8/android/src/main/java/com/brentvatne/react/ReactVideoView.java#L42-L43)

jariwalabhavesh:
@rayronvictor  @mattapperson  : _onPlaybackRateChange is never fired on android. It is still issued in version: 4.4.1 which cause poster not a hiding issue in android.

PR #426, opened 2016-12-23
ExoPlayer

AndrewJack:
Implementation of react-native-video using [Google's ExoPlayer library](https://github.com/google/ExoPlayer)

View the [ExoPlayer README](https://github.com/drivetribe/react-native-video/blob/13938e95d9f1b6a9f7dd28a8aa6f784f5a4b92b1/android-exoplayer/README.md) to see the new props and the props that still need to be implemented.

This also adds a buffering callback to iOS and ExoPlayer Android.

Long term I am hoping we can move to ExoPlayer as the default, if not sooner.

p.s. if you need any help maintaining this project I'd be happy to help out on the Android side.

PR #429, opened 2016-12-29
[Android] Fix `require(...)` file resource on production

### LINKED ISSUE #338: Android - Video is not playing on release apk
RN 0.34
Using asset system (`source={require('../path/to/video.mp4')}`)
Working great on iOS and Android debug (`app-debug.apk`)
But not working on Android release APK

Any thoughts?

jhen0409:
Fix #338.

Should use `drawable` instead of `raw` for `setRawData`, see [`ResourceDrawableIdHelper` of react-native Image](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/imagehelper/ResourceDrawableIdHelper.java#L54-L57) usage. ([getAssetDestPathAndroid.js](https://github.com/facebook/react-native/blob/24d2bbfbab174aa528aa73cccbbefb67f95c2451/local-cli/bundle/getAssetDestPathAndroid.js))

Build the [example](https://github.com/react-native-community/react-native-video/tree/master/example) release apk and use [Apktool](https://github.com/iBotPeaches/Apktool) decode, can be simply seen:


el-lsan:
@jhen0409 I tried your solution, but I'm still getting the same error:
`W/MediaPlayer: Couldn't open file on client side; trying server side: java.io.FileNotFoundException: No content provider:` 

FYI i'm trying to load 4 videos (each ~100kb) on my walkthrough component and only 2 of them are loading randomly. (on IOS everything works fine)


```
import Video from 'react-native-video';
var resolveAssetSource = require('react-native/Libraries/Image/resolveAssetSource');
...
...
...
  render() {
    let sourceVideo1 = resolveAssetSource(require('./video1.mp4')) || {};
    let sourceVideo2 = resolveAssetSource(require('./video2.mp4')) || {};
    let sourceVideo3 = resolveAssetSource(require('./video3.mp4')) || {};
    let sourceVideo4 = resolveAssetSource(require('./video4.mp4')) || {};
    return (
             ...
             ...
             <Video
              ref={(ref) => {
                this.firstPagePlayer = ref
              }}
              source={sourceVideo1}
              key="video1"
              resizeMode="cover"
              repeat={false} 
              playInBackground={false}
              playWhenInactive={false}
              style={[styles.image, {backgroundColor: 'white'}]}
            />
....

```

jhen0409:
@el-lsan could you provide your react-native & Android version?

Also I [updated example](https://github.com/react-native-community/react-native-video/pull/429/commits/c3dd2a6d20f9aed2c5f88adede0e988c2497cc22), you can try it out.

jhen0409:
I updated this PR, just use `drawable` instead of `raw` for `setRawData`. (See [`ResourceDrawableIdHelper` of Image usage](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/imagehelper/ResourceDrawableIdHelper.java#L54-L57), the assets file should in `drawable`)

el-lsan:
@jhen0409 I'm using `"react-native": "^0.38"` and `Android 6`,
I also tested using latest RN `0.39.2`but still facing the same issue.

jhen0409:
@el-lsan try print `identifier` with `Log`, if it is not 0, the problem may related to [ScalableVideoView](https://github.com/yqritc/Android-ScalableVideoView), I also have problem of video load in the emulator.

jhen0409:
Resolved the conflict and also fix `android-exoplayer` part.

@mattapperson

Video.js
```diff
@@ -180,7 +180,7 @@ export default class Video extends Component {
         uri,
         isNetwork,
         isAsset,
-        type: source.type,
+        type: source.type || 'mp4',
```
jhen0409: It looks removed by #426 (See [this](https://github.com/react-native-community/react-native-video/pull/426/files#diff-2a97390647ad4fd466bccc2c92e7864eL165)), but it will broken native prop update (if it is `undefined`), so I added back it.

AndrewJack: The exoplayer implementation can infer the type of video, but can be overridden by the `type` field.
By adding a default value the player doesn't infer the `type` and just assumes it is mp4, and playback fails if it isn't mp4.

Could we make this an empty string instead?

jhen0409: @AndrewJack OK, will change to empty string. :)

robbue:
Is this ready to be merged?

krailler:
I need this fix.

tarunmehta-quovantis:
Did any one find how to resolve this issue ?

PR #444, opened 2017-01-10
Protect events

### LINKED ISSUE #418: playInBackground causes crash
Hello,

After merging #381, which fixed all the callbacks, `playInBackground` is now causing a runtime exception (this crash is alluded to in that PR). To reproduce, add a true `playInBackground` attribute to the `<Video />` component, e.g.:

```js
<Video
  ...
  playInBackground
  ...
/>
```

in the file `example/index.ios.js`.

Thanks.

Screenshot of crash:

screen shot 2016-12-13 at 2 00 09 pm

olofd:
This is protection against nil values of events. These are almost never needed in this library because the JS-file Video.js sets upp all events with defaults. Something that is not needed when these guards is in place, and this would allow the JS-file to be cleaned up.

But it also prevents a crash today and will fix #418 amongst others.
The reason these crash is because the event-blocks has not jet been initialized when the event is bound to be fired and therefore the app crashes. There are no guarantees when a setter will be fired i RN, This is why we need this.. then it will not matter when the event blocks get set.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -184,7 +184,7 @@ static NSString *const playbackRate = @"rate";
    CMTime currentTime = _player.currentTime;
    const Float64 duration = CMTimeGetSeconds(playerDuration);
    const Float64 currentTimeSecs = CMTimeGetSeconds(currentTime);
-   if( currentTimeSecs >= 0) {
+   if( currentTimeSecs >= 0 && self.onVideoProgress) {
       self.onVideoProgress(@{
                              @"currentTime": [NSNumber numberWithFloat:CMTimeGetSeconds(currentTime)],
                              @"playableDuration": [self calculatePlayableDuration],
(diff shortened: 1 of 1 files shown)


hellogerard:
> The reason these crash is because the event-blocks has not jet been initialized when the event is bound to be fired and therefore the app crashes. There are no guarantees when a setter will be fired i RN,

@olofd Interesting. I had assumed that React Native would guarantee this. There's nothing in the documentation about these event handlers getting set up asynchronously. Do you have any links or further reading on the subject?

olofd:
@hellogerard They might not be set up async, but what I ment is that there is no guarantee about in which order the setters are called in. 

In this case (#418) it was the block/event `onPlaybackRateChange` that caused the crash, this was (probably) because  the setter for `src` ie. `setSrc`was called first, triggering the video to start, then triggering `observeValueForKeyPath`with a new playback rate. First after this, the setter for the block `onPlaybackRateChange` was called setting the listener. (At least my theory).
And all of this could have happened synchronously.

I think it's better to guard before any event invocation. This is good for multiple reasons imho. Will protect agains syncronization-bugs like this, will only generate data and invoke the JS-bridge if there is an actual listener on that event and it will make it possible to clean up the JS-file.

IMO the event-name-mapping in the JS-file is highly confusing in this library for instance. But that is a later discussion.

hellogerard:
🎆 🎉

Issue #368, opened 2016-10-29
Resize video on dynamic animated components


proProbe:
Hello!

I currently need to somehow resize a video during an animation. Im changing the videoplayers size with a Layoutanimation which makes the video stretch.

So my question is if it is possible to resize the video somehow so it fits to the new height and width of the container? Reloading the video is also a plausible solution but Im not sure how to even do this.

Should also notice that this is mainly on Android and not on IOS

arjan:
Hi, I stumbled into this as well and fixed it (for Android), the PR is here: https://github.com/react-native-community/react-native-video/pull/432

awitherow:
Does anyone have the iOS experience to duplicate the changes for this as well? Our application is using this package and we have this problem on iOS.

arjan:
On iOS it should not happen. However, make sure you specify the `muted` attribute! Otherwise iOS does not follow the `resizeMode` specifier.

DanielMSchmidt:
@arjan That worked for us in iOS, just opened a PR, because this behaviour is kind of weird and unintuitive. Thanks for your help, we really appreciate that you took the time! 👍

arjan:
NP :) I also found this out more or less by accident.

proProbe:
great this seems to be fixed! thanks everyone!

hamzaouhssain1993:
I think i have a similar problem
I have this code 
`<Video

                    ref={this.attachVideoRef}
                    source={{ uri: this.mediaShowed.displayPath() }}
                    onEnd={this.handleVideoEnd}
                    paused={this.isPaused}
                    style={s.mediaViewerDefault}
                    onVideoError={$warn}
                    onError={$warn}
                    resizeMode="contain"
                />`
and 
`    mediaViewerDefault: {
        flex: 1,
        alignItems: 'center',
        justifyContent: 'center',
    },`

but when I change the Url to another video, only in android the size kept is the one from the previous video so the results come with a stretched video

Foukz:
@hamzaouhssain1993 
I've had a similar issue. Rerendering <Video /> component after data change worked for me. You can do it for instance by key property


```
<Video
     source={{
        uri: videoData.path,
     }}
     key={videoData.path}
     style={{ height: '100%', width: '100%' }}
     resizeMode="contain"
 />
```

hamzaouhssain1993:
Yes it worked with the Key! Thank you!

PR #459, opened 2017-01-17
Add remark on muted for cover size

DanielMSchmidt:
This is related to the comment on #368 
I wanted to add this remark as I found it non-trivial to have to add such an attribute for full cover mode


DanielMSchmidt:
cc @jsierles @brentvatne @isair Do you have any opinion on this topic in general? Would love to see this PR merged as this behavior is not to be expected and actually costed us a lot of debugging time. Would be cool to change the behavior in this case, but merging this first would be huge plus

PR #506, opened 2017-02-23
corrected the doc about paused property

linmic:


DevAlien:
I think that is the default value, so this PR is wrong.

linmic:
The intent was to correct the misleading comment, since `paused={false}` won't `pauses playback entirely`. However, it makes senses to keep the default value of `paused`, so I updated the PR a bit.

PR #526, opened 2017-03-12
Pausing progress if it was playing and the stating again

deshiknaves:
When loading a network resource, when seeking the player has really weird behavior of it continuing to play till it has buffered enough to play. Also when it is buffering, it doesn't notify that it is buffering. At least this way, some logic can be added to `onProgress` to track when the audio is not actually playing.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -532,9 +532,12 @@ static NSString *const timedMetadata = @"timedMetadata";
     CMTime current = item.currentTime;
     // TODO figure out a good tolerance level
     CMTime tolerance = CMTimeMake(1000, timeScale);
+    BOOL wasPaused = _paused;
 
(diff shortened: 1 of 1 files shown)


djw27:
@deshiknaves can you please explain why this was added and what the intended behaviour is?

This introduced a regression for us where the 'paused' and 'rate' properties are being ignored as a result of this change. It looks like the paused and rate props are being ignored when setting the video to playing with this change.

Reproduced by seeking a video to 0 then setting the paused and rate props to pause the video. Video will autoplay from 0.

deshiknaves:
When you scrub a track, and it doesn't have it buffered already, then it continues to play until it has enough to play. Normal behavior is for it to stop and then start up when it can play.

This was to make it stop if it is playing if it's playing and then restart it when there is enough to play. Maybe it needs to check more flags?

Issue #530, opened 2017-03-14
paused prop not being honored


anshul-kai:
`paused` prop is no longer working.

xavieramoros:
Related to #494

ghost:
Closing dupe

Issue #494, opened 2017-02-14
Android video plays on load regardless of paused flag (device/release only)


briancalvium:
using `1.0.0`
I have the following in JS:

```
<RNVideo
  ...
  paused={true}
/>
```

in debug, this works as expected - user sees first frame of video, and video is paused. In release (on a Samsung S6), the video plays. 

I added a log statement to ReactVideoView:
```
if (mPaused) {
            if (mMediaPlayer.isPlaying()) {
                pause();
            } else {
                Log.d("VIDEO", "Attempted to pause when not playing!");
            }
        } else ...
```

and am seeing that message in LogCat.

sarmadka:
I confirm this. Having the same issue. For me the video is playing even in debug mode in the simulator.

DevAlien:
I found out the same issue, Videos play regardless of paused and of rate={0}

pepavesely:
Same problem.

davidjeffries:
Same issue as: #484 

Workaround, change the state on video load:
```
this.state = {paused: false};
```
Component:
```
paused={this.state.paused}
onLoad={() => {
  this.setState({
    paused: true
  });
}}
```

ghost:
**ROOT CAUSE**
I think I found the possible root cause. In ReactVideoView, the onHostResume is called soon after the player is initialized and launched. What I believe is an unintended consequence is that whatever value was assigned to mPaused gets overwritten by the value of mActiveStatePauseStatus (which is always false when the class is initialized). This happens in the onHostResume when the setPausedModifier method is called.

**SOLUTION**
In the ReactVideoView class, you need to initialize mActiveStatePauseStatusInitialized not to false but to the value of the paused property that comes down from React Native. To achieve that:

1) add this in the constructor:
       private boolean mActiveStatePauseStatusInitialized = false;
2) add this in the setPausedModifier method, after the first line of code:
       if   (  !mActiveStatePauseStatusInitialized  ) {
               mActiveStatePauseStatus = mPaused;
               mActiveStatePauseStatusInitialized = true;
       }

gorjanz:
I can confirm that @devmaster72 solution works. Is there a pull-request already, or should I create one?

Thank you!

xavieramoros:
I have the same issue as well, solved using @davidjeffries solution to avoid initial paused flag problem. 

[...]

PR #537, opened 2017-03-21
Fix/494

gorjanz:
In order to address #494 and #530 

two fixes are introduced with this pull request:
1. as suggested by @devmaster72 the auto-start of the video is fixed, by taking into the account the 'paused' prop passed
2. An IOException was constantly being thrown from the android Media player classes, because instead of only a URL, the media player, for a dataSource was expecting a whole ContentProvider initialised with context


ghost:
thank you @gorjanz for the PR

karltaylor:
@gorjanz Can you remember why you removed `headers` from 
```javascript
setDataSource(mThemedReactContext, parsedUrl, headers);`
```

This stops cookies being sent on android.

PR #563, opened 2017-04-09
Added flag for background mode

deshiknaves:
Added the missing implementation for play in background for exoplayer.


wschurman:
@mattapperson - Friendly ping :)

markusekblad:
@deshiknaves @mattapperson 
Great stuff, I had one issue tho,
I tried this today and for me the

```
public void onHostResume() {
    if (playInBackground) {
        return;
    }
    ...
}
```
causes the player to start as paused even with paused false prop passed.

It also didn't play in the background. However, if I remove that check from onHostResume, and add it instead to

```
    @Override
    protected void onDetachedFromWindow() {
        super.onDetachedFromWindow();
        if (playInBackground) {
            return;
        }
        stopPlayback();
    }
```

Then background audio works for me and there's no issue with the paused prop. I'm using a modal player and not the fullscreen one if that makes any difference.
(Tested on a Huawei Honor 8 and Samsung S6, both with Android 6.x)

PR #574, opened 2017-04-18
Warn that iOS needs a distant url to use https

maximelebastard:
I tried to load a distant mp4 video for about an hour before realizing that iOS 10 requires all the webservices and files loaded in the apps to use https.
I think a warning about that can be a good thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ using System.Collections.Generic;
 // "background.mp4" in your project. You can include multiple videos
 // on a single screen if you like.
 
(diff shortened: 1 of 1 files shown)


cyakimov:
You **can** load resources on HTTP if you add the proper exceptions to `Info.plist`. Look for _Configuring App Transport Security Exceptions_.

akshaygore06:
We have added a note in PR #1025. Thank you so much spotting this issue.

PR #587, opened 2017-04-26
Fix repeat for exoplayer

radko93:
Repeat was not working for me when using exoplayer. I found the reason: `setRepeatModifier` method was called after [this line](https://github.com/react-native-community/react-native-video/blob/master/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L208) in `initializePlayer`. This might not be the best solution here, so I'm waiting for any suggestions.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -404,9 +404,14 @@ class ReactExoplayerView extends FrameLayout implements
                 videoLoaded();
                 break;
             case ExoPlayer.STATE_ENDED:
-                text += "ended";
-                eventEmitter.end();
-                onStopPlayback();
+                if(repeat) {
+                    playerNeedsSource =  true;
+                    startPlayback();
+                } else {
+                    text += "ended";
+                    eventEmitter.end();
+                    onStopPlayback();
+                }
                 break;
             default:
                 text += "unknown";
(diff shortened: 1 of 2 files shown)


zachrnolan:
Would love to see this fix merged!

radko93:
It's currently not possible to switch from `true` to `false` in this PR (video will loop forever), I've tried to investigate this issue but now waiting for the response here: https://github.com/google/ExoPlayer/issues/2899.

radko93:
Seems that better repeat mechanism will be soon added to the Exoplayer (it's already on a dev branch), so probably we will ned to wait. I'm leaving this here PR for now.

maboelsoud:
I think this is causing #643 as a side effect. #628 is no good either

radko93:
@Mohamed-Abo-El-Soud I'm not sure if it's possible, I'm just doing something only when the video ends. As I said in the comment above, better repeat mechanism will be soon merged to exoplayer.

maboelsoud:
@radko93 side question: why does there seem to be so many little bugs using the android exoplayer and mediaplayer, where ios don't seem to have them? Is it that the functionality is somewhat new on the android side?

radko93:
@Mohamed-Abo-El-Soud Seems that exoplayer is still in development somehow. iOS is more mature and it's just one platform, it's always on Android.

cobarx:
They added a `setRepeatMode` function to control repeat behavior in ExoPlayer 2.5.0. I've gone ahead and implemented this in #1036. Thanks for diving into this and finding out that ExoPlayer was going to support it.

PR #601, opened 2017-05-04
fix a bug occurred when user switch between full screen mode and non full screen mode

zyangSir:
After entering full screen mode, the _playerLayer would be hide by
_playerViewController, which means not visible to user,
so we set the _playerLayer.player to nil,
because we don't need keeping the _playerLayer rendering video data.
 
Meanwhile, fix the bug that the _playerLayer's video would be stopped
(while the audio keeps going on), after user dismiss the _playerViewController

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -636,6 +636,20 @@ static NSString *const timedMetadata = @"timedMetadata";
         }
         if( viewController )
         {
+          /*
+           *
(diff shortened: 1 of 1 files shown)


cobarx:
Hi @zyangSir, is this bug still happening with the latest react-native-video. I want to make sure it hasn't gotten resolved by some of the newer fixes.

If it's still an issue, can you provide some sample code so I can see the issue happening and then test it with your PR to see it being fixed. Thanks!

PR #614, opened 2017-05-12
Add tvOS Static Library for Linking on tvOS Build Targets

FullstackJack:
Apple requires that tvOS apps only include libraries that were built against the same platform. In attempting to build a tvOS app with the current `libRCTVideo.a` archive, you will run into a clang error. In order to fix this issue, we need to add a new tvOS Static Library build target for this package. After a lot of trying, I was able to accomplish this task. Following convention from other React Native modules (i.e. RCTWebSockets), the build target Product Name should be named *RCTVideo-tvOS*. After saving the target, you'd need to manually link the tvOS library under Linking in your main project under tvOS build target, so adding `libRCTVideo-tvOS.a` will do the trick.

### DIFF
--- a/ios/RCTVideo.xcodeproj/project.pbxproj
+++ b/ios/RCTVideo.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		274428A51EC53469006D3DD4 /* RCTVideo.m in Sources */ = {isa = PBXBuildFile; fileRef = BBD49E3A1AC8DEF000610F8E /* RCTVideo.m */; };
+		274428A61EC5346D006D3DD4 /* RCTVideoPlayerViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 31CAFB2E1CADC77F009BCF6F /* RCTVideoPlayerViewController.m */; };
+		274428A71EC53470006D3DD4 /* RCTVideoManager.m in Sources */ = {isa = PBXBuildFile; fileRef = BBD49E3C1AC8DEF000610F8E /* RCTVideoManager.m */; };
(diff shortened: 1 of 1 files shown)


aditya27dev:
Hi @FullstackJack 

I also want to do the same. Can you help me please? I am not getting which file i need to rename.

FullstackJack:
You need to add a build target to the library using the naming scheme I mentioned. I don't have XCode in front of me right now.

FullstackJack:
It is really hard to find documentation on this, but you double click the library in the Project Navigator, which opens the project for that module. Once that's open, click the main project for the module in the subsequent Project Navigator. There should be a projects and targets pane on the right. Find the button that allows adding build targets (I think its above that list). Add a new static library build target tvOS as libRCTVideo-tvOS. Once it's been created, click the target in the targets pane and add all the source files to the build.

FullstackJack:
If this PR gets merged, nobody should have to go though the headache of setting this up.

FullstackJack:
Looks like there is another PR that has been introduced after this one which also updates the readme. I think that one will get merged soon.

aditya27dev:
Hello @FullstackJack 

Thanks for the reply.

I have renamed libRCTVideo.a to libRCTVideo-tvOS.a in project.pbxproj. And fortunately it is working. But i am afraid that renaming this file can give me error while releasing this app for iPhone (because i have renamed file for tvOS) and i want to release this app for both iOS and tvOS

FullstackJack:
That is not the correct approach. It has to be built against the correct
platform and yes, Apple will not allow it. The gpod news is a PR for tvOS
was merged today. Checkout the latest version.

aditya27dev:
Awesome.

PR #615, opened 2017-05-12
Update header paths to be compatible pre and post RN 0.40

mattvot:


mattvot:
Sorry, PRed into upstream instead of our fork

PR #619, opened 2017-05-15
validate uri first

c19354837:
when you don't set  'source'  in <Video />, iOS will crash, while
Android is OK.
Android will return immediately if uri is empty, and iOS just do the
same thing

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -27,6 +27,10 @@ export default class Video extends Component {
     this.setNativeProps({ seek: time });
   };
 
+  reload = (uri)=>{
+    this.setNativeProps({ src: {uri:uri} });
+  }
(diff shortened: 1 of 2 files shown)


c19354837:
More detail:

It will crash at [RCTVideo.m#L307](https://github.com/react-native-community/react-native-video/blob/master/ios/RCTVideo.m#L307)
`[[NSURL alloc] initFileURLWithPath:[[NSBundle mainBundle] pathForResource:uri ofType:type]];`

The log could be `*** - [NSURL initFileURLWithPath:]: nil string parameter`

Zisomedia:
Confirmed to work very nicely

PR #620, opened 2017-05-15
Switch to RCTSharedApplication & add photo support on iOS

chrisbianca:
`[UIApplication sharedApplication]` is not available in iOS share extensions as it is blocked by Apple.  Updated to use `RCTSharedApplication()` instead.


rvignesh89:
@chrisbianca any plans to finish this? I'm looking at support for `ph://` asset URLs especially. Actually I've forked the library and made changes to get it working for me. If you are not planning to proceed then I can create another PR for that alone.

hueniverse:
Consolidated into #2652

PR #625, opened 2017-05-19
Fix paused prop doesn't work when init the Video

Symous:
It caused by onResume life cycle method. when the first time to run onResume, the mActiveStatePauseStatus always be false, so it always play the video.

now we can use paused prop to control the video playing status.

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactVideoView.java
+++ b/android/src/main/java/com/brentvatne/react/ReactVideoView.java
@@ -326,6 +326,7 @@ public class ReactVideoView extends ScalableVideoView implements MediaPlayer.OnP
     public void setPausedModifier(final boolean paused) {
 
         mPaused = paused;
+        mActiveStatePauseStatus = paused;
 
         if ( !mActiveStatePauseStatusInitialized ) {
             mActiveStatePauseStatus = mPaused;


vomchik:
Very important fix :) 
Thanks!

krizpoon (commented):
one problem in this approach: if the video is playing and onHostPause is triggered, the video would not resume when onHostResume because onHostPause will invoke setPausedModifier(true);

abhideepmallick:
Life saver, please merge this!

krizpoon:
I think the latest master branch already fixed this

luisfuertes:
In master package.json file, version is 1.1.0, but on `npm install react-native-video@1.1.0`:
captura de pantalla 2017-07-20 a las 12 53 27

SharanyaRao91:
Manually added "mActiveStatePauseStatus = paused;" to my ReactVideoView.java file. But videos continue to autoplay. Here is what I have (RN - 0.44; react-native-video - 2.0.0) - 
                        rate={0}  //paused
                        paused={true}
                        volume={1}
                        muted={false}
                        playInBackground={false}    // Audio continues to play when app entering background.
                        playWhenInactive={false}
                        ignoreSilentSwitch={"obey"}
                        resizeMode="cover"
                        onError={e => console.log(e)}
                        onLoad={load => console.log(load)}
                        repeat={false}

afilp:
Does anybody maintain this library anymore? We need these PRs merged, if possible. Thanks!

cobarx:
I reworked this in #1082 so we don't even need the active status paused field. In my testing everything is working properly now. Will be released in 3.0.

PR #626, opened 2017-05-19
Change react-windows become optional package

jacobdam:
It's really excited that `react-native-video` supports Windows. But I believe that makes more sense to put `react-native-windows` as an optional dependency. By doing this, if users don't build app for Windows, they don't have to install `react-native-windows`.


jacobdam:
I figured out that after moving `react-native-windows` to `optionalDependencies`, yarn/npm still installs `react-native-windows` by default. The best way to avoid installing `react-native-windows` is split the package into 2 packages: `react-native-video` and `react-native-windows-video`.

PR #628, opened 2017-05-19
reloadSource if repeat is set after paused

jalieven:
When the Exoplayer first receives the paused=false before it receives repeat=true in the current implementation the video will not get repeated. Since the LoopingMediaSource will get instantiated when paused is set (which calls startPlayback, which calls initializePlayer, which results in a non looping source). So simply reloadSource after the repeat prop is set fixes this issue.


radko93:
Is this stopping a video if you change repeat prop while playing the video? I did a fix in #587 but it's a bit different. Maybe your PR is better if it does not have any side effects.

jalieven:
I haven't tested the change of repeat yet with this PR. Can verify this on monday.

jalieven:
@radko93 if you change the repeat prop while playing the video the video starts anew but plays once when repeat was changed to false and loops if it was changed to true. That is if you have the paused prop set to true

jalieven:
@radko93 When I try your PR #587 and change the repeat to false (while playing the video) where it was first true, the video keeps on repeating. Have you tested this use-case?

radko93:
@jalieven thanks for the feedback, I will try and fix that.

jalieven:
Anyhow if this use-case works your PR looks nicer indeed. Mine is kinda brute-force.

cobarx:
We no longer need to use a LoopingMediaSource since ExoPlayer added support for a repeat mode in 2.5.0. This is tested and working with the latest release.

PR #681, opened 2017-07-06
[tvOS] Add target 

JulienKode:
Now we can use react-native-video on tvOS 

xifprkajvs


Kureev (approved):
Nice! Looks good to me

PR #707, opened 2017-07-25
Feat/ontimedmetadata android

RWOverdijk:
This PR adds support for getting the id3 tag (currently support for the single timecode) in android. I have tested it and it functions identical to the other platforms in the projects.

### Notes

* Don't merge yet. The min sdk version now is 23. I have to make this optional first, otherwise this is a BC break.
* Any pointers on getting the tags in a "prettier" manner than this are welcome. I've spent a full working day on google trying to find a solution that's prettier, and all of them were worse than this one, which is more elegant in terms of performance and readability.

I'll be forking this module, with this PR merged as I can't wait for your feedback right now. I'll try to keep submitting PRs as long as I don't stray from the current codebase too far.

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactVideoView.java
+++ b/android/src/main/java/com/brentvatne/react/ReactVideoView.java
@@ -37,7 +37,6 @@ public class ReactVideoView extends ScalableVideoView implements
     MediaPlayer.OnPreparedListener,
     MediaPlayer.OnErrorListener,
     MediaPlayer.OnBufferingUpdateListener,
-    MediaPlayer.OnTimedMetaDataAvailableListener,
     MediaPlayer.OnCompletionListener,
     MediaPlayer.OnInfoListener,
     LifecycleEventListener,
@@ -197,8 +196,10 @@ public class ReactVideoView extends ScalableVideoView implements
             mMediaPlayer.setOnPreparedListener(this);
             mMediaPlayer.setOnBufferingUpdateListener(this);
             mMediaPlayer.setOnCompletionListener(this);
-            mMediaPlayer.setOnTimedMetaDataAvailableListener(this);
             mMediaPlayer.setOnInfoListener(this);
+            if (Build.VERSION.SDK_INT >= 23) {
+                mMediaPlayer.setOnTimedMetaDataAvailableListener(new TimedMetaDataAvailableListener());
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


RWOverdijk:
This is working great for me. Any feedback?

raihanabbas232:
@RWOverdijk thanks for your concern...looking for some way to test that 👍...
Secondly, do we have to make any changes in `ReactVideoViewManager.Java` class for MetaData???

RWOverdijk:
@raihanabbas232 All the changes required are in this PR.

fozzle:
thanks for this work and hope this gets merged soon

nromptea42:
Hi @RWOverdijk, did you manage to make the onTimedMetaData optional ? I'm looking to implement this in my app but I don't want to loose my users who uses api 22.
Thank you !

RWOverdijk:
Not sure I understand your question. This method worked for me and was optional.

nromptea42:
I took the code as it is in this pull request and applied it on a fork. It works great on android >=6.0 but it crashes on android 5.1. Maybe I did something wrong but I don't know what.
You said in your note : `The min sdk version now is 23. I have to make this optional first, otherwise this is a BC break.`
I want to know if it is possible to use your modifications  on android 5.1 without crashing (Of course, no timedmetadata will be triggered).

Ziv-Barber:
Not working. I'm not getting any metadata in the ontimedmetadata callback.
But I may doing something wrong in the encoder.
I'm using my plugin in wowza to add a AMF metadata to the stream.
I'm not getting a call to onTimedMetaDataAvailable in the java code (using this PR)

cobarx:
Is there a stream I can use to test this? I've tried a couple HLS streams and not had any luck getting onTimedMetadata to fire on iOS either, so I don't think they'll work for testing this. I'm totally not up on using timed metadata so a complete set of instructions would be great.

I can handle making this backwards compatible once I understand how to test this.

cobarx:
@RWOverdijk Ok I've updated the code so that you don't need to update to SDK 23. We don't want to make 23 the minimum as many people still want to support 4.1 or 4.4.

I've tried to test this with the only timed metadata stream I have and the listener is not detecting it. The URL is:
https://raw.githubusercontent.com/jwplayer/jwdeveloper-demos/master/demos/basic/audio-metadata/assets/index.m3u8
The same URL works fine on ExoPlayer and iOS.

Between only supporting newer devices, not detecting all metadata, and having to do major hacks to format the output, I'm not sure this is very workable. Is it worth supporting this or should we tell people to use ExoPlayer if they need timed metadata?

PR #717, opened 2017-08-01
Fix deprecations: createJSModules and React.PropTypes

john-bernardo1:


deathemperor:
can anybody look at this?

cobarx:
This was handled by #718, so we don't need this anymore. Thanks for working on it though!

PR #737, opened 2017-08-14
Add check to seekTo so that the time is among loadedTimeRanges

alizbazar:
Previously if `seekTo` was called before the time in question was loaded, this broke the player so that the `playableDuration` and the `currentTime` were wrong in `onProgress` event. Also, if `seekTo` was called again, this didn't do anything.

This commit fixes this by making sure the time is among `loadedTimeRanges`. If not, `onSeek` callback is called with `success: false`. 

Additionally, `success` parameter is also added to succeeding `onSeek` callback.


alizbazar:
Guys, any chance of having this merged? Or is there a problem with this?

cobarx:
@alizbazar 
Thank you for the detailed write-up. Looking at the code in detail, it doesn't feel like the right approach. My understanding is that loadedTimeRanges contains the portions of the video that are buffered. By checking against that, you're only verifying whether the position you want to seek to is buffered, not whether it's seekable. Any position within the seekableDuration should be seekable regardless of if it's buffered - if you seek to it, it will begin buffering then play.

Perhaps there is some other issue at play here. What do I need to do reproduce the issue you are encountering? I tried both streaming and large local videos and the whole video was seekable without issue, so I'm not able to troubleshoot and investigate this.

cobarx:
@alizbazar I'm going to close this for the reasons shown above. If the issue with the currentTime and playableDuration still exist, please provide an example and I'll look into it.

PR #758, opened 2017-08-29
Show iOS controls without using present full screen

LandonS23:
This allows you to show the controls without calling the present full screen method.


ryanvanderpol:
Can this get merged in please?

cobarx:
Maybe I'm missing something, but the only time I'm seeing createPlayerViewController get called is when I use presentFullscreenPlayer. Can you provide an example of how you're using this? It may also be that something has changed when they switched over to the new video player in iOS 11.

Also, I don't think we want to make this mandatory behavior, some people are going to want to provide their own controls UI, so this should be an optional prop.

PR #772, opened 2017-09-07
Updated README.md, because example was not working

janrop:
As mentioned here: https://github.com/react-native-community/react-native-video/issues/471 the example does not work without specifying the type of the source.


cobarx:
Thanks for pointing this out! Since the README has changed so much I did a fresh commit e4f8343 to handle this.

PR #783, opened 2017-09-16
Exoplayer - removed black screen while loading

maboelsoud:
Removed "shutterView" because it purposely turns the screen to black while the video is loading. This phenomenon is bound to be hinder appearance, as the black screen looks jarring and out of place.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -32,7 +32,6 @@ import java.util.List;
 public final class ExoPlayerView extends FrameLayout {
 
     private final View surfaceView;
-    private final View shutterView;
     private final SubtitleView subtitleLayout;
     private final AspectRatioFrameLayout layout;
     private final ComponentListener componentListener;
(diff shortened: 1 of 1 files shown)


maboelsoud:
@mattapperson can you review this?

mattapperson:
@Mohamed-Abo-El-Soud I would rather see this as a prop to turn this on or off, times exist when this is preferable.

maboelsoud:
I'll whip something up

cobarx:
@Mohamed-Abo-El-Soud What sort of situations are you running into issues with it being jarring?

Did you want to get this PR updated to make the behavior optional? I'd like the black screen to be the default as at least IMO it makes sense in most situations.

maboelsoud:
@cobarx 
I was trying to build an app which shows multiple videos. And one video plays at a time. Think of it like Instagram, as you scroll through your feed, a single video plays at a time (u load one video only for performance). Thing is, when the video is not playing, you gotta show something, anything. I can't just have a black screen, it looks ugly! 

So I put a thumbnail in-front of the video, and before the video starts playing, the thumbnail dissapears. But then another problem, when the video starts playing, you see thumbnail-then-black-screen-then-video. That looks jarring and weird. So I try to have a callback, when the first frame of the video loads, the thumbnail dissapears. But that doesn't work because the JavaScript thread is slower than native, so the video plays for a second before the thumbnail is removed.

In the end, I found that removing the black screen, fixed my problem nicely. This also allows you to put whatever you want, black screen, Pink screen, blue screen whatever. And also it seamlessly transitions from no-video to video while loading. If you want to put a custom loading screen, you can. So it's a win-win for everyone! What's not to like?

PR #790, opened 2017-09-21
Set fullScreenPlayer's orientation [iOS]

gevgasparyan:
Pass `autoRotate` (default true) and `screenOrientation` prop to FullScreenVideoPlayer.

```
<Video
            source={{ uri: 'some_source_here' }}
            .....
            autoRotate={false}
            fullScreenOrientation={Video.Constants.landscape}
          />
```
`fullScreenOrientation` can be one of these values (default is `all`). 
```
    all,
    allButUpsideDown,
    portrait,
    portraitUpsideDown,
    landscape,
    landscapeLeft,
    landscapeRight
```
fullScreenOrientation affects even when device's `rotation lock` is enabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,8 @@ using System.Collections.Generic;
        onError={this.videoError}               // Callback when video cannot be loaded
        onBuffer={this.onBuffer}                // Callback when remote video is buffering
        onTimedMetadata={this.onTimedMetadata}  // Callback when the stream receive some metadata
+       autoRotate={false}                      // [iOS] Determines if FullScreenPlayer can be rotated (default true).
+       fullScreenOrientation={Video.Constants.landscape} // [iOS] Determines FullScreenPlayer's orientation.
        style={styles.backgroundVideo} />
 
 // Later to trigger fullscreen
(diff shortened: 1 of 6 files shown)


shahen94:
@jhabdas Can we merge this ?

cobarx:
@gevorg94 It seems like having autoRotate is redundant, since they can control whether it is allowed to rotate by toggling between allowing all orientations and specific ones.

gevgasparyan:
@cobarx If you want to have only one orientation than you need to set autoRotate={false}, otherwise orientation will be changed when you rotate your device (assuming that screen lock is not enabled). So, by setting specific orientation and autoRotate={true}, your specified orientation will be an initial orientation and will be changed when you rotate your device.

cobarx:
@fabricioasap Yes, I've been going thru and reviewing and testing each PR so it takes a fair bit of time. I'll give this a go as soon as I can.

@gevorg94 One thought I had is that react-native-orientation already provides this functionality. What are your thoughts about whether it's worth duplicating that in this project vs. recommending people use that?

cobarx:
This functionality has been rolled into #1215, so closing.

PR #795, opened 2017-09-26
Take out auto-resume behavior on Android

lucask42:
Upon pausing the video by backgrounding the app the video would restart upon bringing it to the foreground.  This was due to using mActiveStatePauseStatus which is false after pausing via onHostPause().  This behavior is undesirable for our media playing app and also not the behavior on iOS.  An alternative to this would have been to use mPaused as the boolean in onHostResume().

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactVideoView.java
+++ b/android/src/main/java/com/brentvatne/react/ReactVideoView.java
@@ -553,10 +553,9 @@ public class ReactVideoView extends ScalableVideoView implements MediaPlayer.OnP
     @Override
     public void onHostPause() {
         if (mMediaPlayer != null && !mPlayInBackground) {
-            mActiveStatePauseStatus = mPaused;
-
             // Pause the video in background
             setPausedModifier(true);
+            mActiveStatePauseStatus = mPaused;
         }
     }
 


jasongonzales23:
:+1:

This change saved our bacon!

cobarx:
Even though the behavior isn't consistent between iOS & Android, we need to keep it the same until we do a major version bump so that this doesn't break for people. I'll merge this when we do a 3.0 release.

lucask42:
Ok, makes sense, thanks for the work you do on this

cobarx:
@jasongonzales23 @lucask42 
Ok, I've done some testing on this and what I'm seeing is that on both iOS & ExoPlayer when you return from the background the video starts playing if it wasn't paused before backgrounding.

I think the right choice is to keep that behavior for two reasons:
1. That's the existing behavior and doesn't require changes to apps
2. It makes sense for the video to always respect the paused prop. If I return to the app and paused is false, I'd expect the video to play not be paused.

If we take that approach, app developers who want the video to be paused after returning to the app  can setup an AppState handler. When the handler sees a background event, it should update the app state so that a `paused={true}` prop gets passed.

However, on Android MediaPlayer onHostPause is wiping out the paused prop so that approach doesn't work currently. To fix that, I think we copy the ExoPlayer approach where there is a separate method that doesn't update the prop that's used to pause the player when going to the background. I'm happy to write that code, but would like to get your thoughts first in case I missed something.

cobarx:
Ok I've created #1082 to implement this. Please give it a test. It's working perfectly for me.

Issue #584, opened 2017-04-25
Setting rate doesn't change video playback speed in Android


emik:
(Not using exoplayer)

emik:
Realised this is a recognised unimplemented feature

RickDT:
I'm seeing this too. I noticed the ExoPlayer README says that rate is not supported. Does that mean rate is not supported on Android at all as of 1.0.0?

bharath-kamath:
Would be good to have this on android.

elaffey:
Has there been update on this issue?

I have also noticed that the README for the ExoPlayer has rate as unsupported but from what I can tell using the ExoPlayer is not the default behaviour. 

By having the following in `android/settings.gradle` it will not use the ExoPlayer
```gradle
include ':react-native-video'
project(':react-native-video').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-video/android')
```

Despite this, rate does not work for android.

cobarx:
Setting the rate is now supported now that #798 is merged. However, the Android media player APIs only support setting the rate on Android 6.0 and up. I'd suggest using the ExoPlayer version instead since it doesn't have this limitation.

PR #798, opened 2017-09-28
Implement playback rate for Android 6.0+

Commander-lol:
This PR implements support for the `rate` prop on android devices running Marshmallow (API 23) or higher. Devices on lower versions of android are unaffected, and will behave as before.

Relates to #584, but doesn't strictly fix it because of the version restrictions

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactVideoView.java
+++ b/android/src/main/java/com/brentvatne/react/ReactVideoView.java
@@ -5,6 +5,7 @@ import android.content.res.AssetFileDescriptor;
 import android.graphics.Matrix;
 import android.media.MediaPlayer;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Handler;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


klirix:
The patch works great!

egorkhmelev:
This patch is great but leads to an issue when media player starts playing with paused state. Docs are clear about this: **After the object is prepared, calling it with zero speed is equivalent to calling pause(). After the object is prepared, calling it with non-zero speed is equivalent to calling start().**

https://developer.android.com/reference/android/media/MediaPlayer.html#setPlaybackParams(android.media.PlaybackParams)

So, if you init your video with rate of 1 it will start playing even with paused=true. I can definitely reproduce this on android versions of 7 and 8, but cannot reproduce on android 6.

cobarx:
I've gone ahead and added code to save the rate and apply it when the player unpauses to handle the concerns that @egorkhmelev brought up. This is working great in my testing, so merging this.

PR #800, opened 2017-09-29
Allow setting progressUpdateInterval to work on iOS

### LINKED ISSUE #521: progressUpdateInterval Not Working
I'm trying to do this:
```jsx
<Video
  source={{uri: this.props.uri}}
  repeat
  resizeMode="cover"
  progressUpdateInterval={16}
  onProgress={play => this.setState({
    playProgress: play.currentTime / play.playableDuration
  })}
/>
```

But onProgress is still being called on the default 250ms interval.

I added some logging in this component and found [setSource()](https://github.com/react-native-community/react-native-video/blob/d792427ce113637f127fb6bc2d92417c668882ce/ios/RCTVideo.m#L250) is being called **before** [setProgressUpdateInterval()](https://github.com/react-native-community/react-native-video/blob/d792427ce113637f127fb6bc2d92417c668882ce/ios/RCTVideo.m#L694), so `setSource()` is always using the default.

I couldn't figure out how to fix it :(

### Versions
```
iPhone 7+
react-native-video: 1.0.0
react-native: 0.42.0
```

danhodos:
The issue here was that the periodic time observer was getting added in `setSrc`, which was happening _before_ `setProgressUpdateInterval`. 

I haven't seen a deterministic way to ensure ordering of the property setters, so the simplest fix seemed to be to check for the existence of the periodic time observer in `setProgressUpdateInterval`, and if it existed to remove and re-add it after setting the new proper `progressUpdateInterval`.

Fixes #521.


marlti7:
can you merge it @jameshartt ?

cobarx:
Thanks for providing this fix, merged!

PR #801, opened 2017-09-29
Fix paused race condition.

jpbarela:
This PR fixes and issue were the player can be paused before the seek command finishes. This has been referenced before in #526.


onchainguy-btc:
Would be nice to see this one merged 👍

PR #816, opened 2017-10-12
Changes to work with react-native-windows 0.48.x

opi-smccoole:
Later versions need a different base signature for Event and setup for npm start.  Also updated dependencies for running the example on more up to date react-native-windows releases but the example does not seem to package correctly with the plugin referenced relatively on the file system.  I think this is due to https://github.com/facebook/metro-bundler/issues/1

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -3,15 +3,15 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "start": "node_modules/react-native/packager/packager.sh"
+    "start": "node node_modules/react-native/local-cli/cli.js start"
   },
   "dependencies": {
-    "react": "15.4.2",
-    "react-native": "^0.42.0",
+    "react": "^16.0.0-alpha.12",
+    "react-native": "^0.48.4",
     "react-native-video": "file:../",
-    "react-native-windows": "^0.40.0"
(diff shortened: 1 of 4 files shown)


KinanKazmi:
@opi-smccoole If i download your PR, would it work for latest of everything ( VS 2017, Win10 10.0.15063,  RN 0.49 ) ?
As the current master branch code breaks while the solution is just in building phase.

opi-smccoole:
@KinanKazmi It should work just fine as far as I know.  I am using VS 2017, target is set to 10.0 Build 14393, min version 10.0 Build 10586.  I have not moved to RN 0.49+ yet due to the breaking change they made in the packager (requiring string literals for import statements) so I don't know if there are issues there.  I'm using 0.48.4 without problems.

KinanKazmi:
Okay thanks. I'd like to try but your repo for this isnt public.

opi-smccoole:
Hmm, it's a fork from this repo so it has to be public.  I just checked it and its marked as public as well.  https://github.com/objectpartners/react-native-video/

KinanKazmi:
Oh its in /objectpartners/
I was looking for it in your profile.
Thanks!

benoitdion:
Superseded by #1871

PR #833, opened 2017-10-25
Fix playInBackground props in ExoPlayer

moduval:
As stated by markusekblad on pull request #563, enabling the flag playInBackground cause the player to not play on startup even if props paused is set to false.

This pull request fix this problem.

Bonus: I updated the readme of android-exoplayer and added instruction on how to use exoplayer instead of default player on android.


moduval:
After some investigation, the sound is killed after a certain time when running in background. We might need to implement something like this: 
https://google.github.io/ExoPlayer/faqs.html#how-do-i-keep-audio-playing-when-my-app-is-backgrounded

cobarx:
Yes, pretty sure you will have to setup a service to support background operation.

I know that react-native-track-player has support for this:
https://github.com/react-native-kit/react-native-track-player
Maybe you could use that as a foundation for supporting this.

zerbfra:
@moduval I'm working on this too. I have no experience in Android so it's kinda difficult. What @cobarx links is exactly what I'm trying to do.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -209,12 +206,14 @@ private void initializePlayer() {
...
             exoPlayerView.setPlayer(player);
             audioBecomingNoisyReceiver.setListener(this);
-            setPlayWhenReady(!isPaused);
             playerNeedsSource = true;
 
             PlaybackParameters params = new PlaybackParameters(rate, 1f);
             player.setPlaybackParameters(params);
         }
+
+        setPlayWhenReady(!isPaused);
```
cobarx: Why was this change made, I'm not sure what it does. When testing with my changes, the video plays automatically even if I switch sources.

PR #844, opened 2017-11-03
adds supported file extensions Android Expansion files

mathewmeconry:
adds mp4 as the only supported file extension for Android Expansion files


cobarx:
Merged! Thanks

PR #847, opened 2017-11-08
If uriString starts with "file://" scheme handle it as a Uri

atsykhonia:
Use setDataSource(Context, Uri) instead of setDataSource(String path) when uriString starts with "file://" scheme. Strictly speaking, in such case uriString represents an Uri object and to convert it to the path the scheme should be removed.


gardusi (approved):

atsykhonia:
Hi @giovanni-gardusi 
As far as I understand, we can merge this PR. But I don't have write access. 
How should we manage this?
Thanks

PR #852, opened 2017-11-15
Add track selection (iOS, Android) and useTextureView prop for ExoPlayer

asdmikk:
This adds the ability to select media tracks during video playback. It is achieved by adding a new `overrideTrack` static method. It works for both iOS and Android (ExoPlayer).
A `useTextureView` prop has also been added to enable smooth transforming of the <Video/> view when using ExoPlayer on Android.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,10 @@ Seeks the video to the specified time (in seconds). Access using a ref to the co
 
 Toggles a fullscreen player. Access using a ref to the component.
 
+`overrideTrack({ rendererIndex: number,  groupIndex: number trackIndex: number })`
+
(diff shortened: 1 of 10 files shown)


acollazomayer:
@asdmikk  This is a great feature, can you give me an example on how to use it. Thanks!

cobarx:
@asdmikk Thanks for submitting this! I've gone ahead and copied your work for `useTextureView` into #1049 so that I can merge it without bringing in the track selection. 

For track selection, I want to make the props a bit simpler and easier to use so you don't have to know the index & renderer numbers. I've already developed a prop for text tracks and will be working on audio tracks soon.

PR #884, opened 2017-12-26
Rn 0.30 exo

tconns:


cobarx:
I'm not clear what this PR is attempting to do, is this just to update ExoPlayer to 2.5.0? I'm going to be updating #983 to use ExoPlayer 2.8.0 shortly, if there are other bits here that are useful, I can see about merging them.

cobarx:
Thank you for submitting this. There is way too much stuff here to merge as one PR. For track selection, I want to look at each type of track individually and come up with an API that makes sense for each type and is user friendly. If you'd like to work with me on this, please open a new PR or issue and we can discuss it.

Issue #534, opened 2017-03-17
this.player.presentFullscreenPlayer() doesn't work on android


wack17s:
hi, button with this.player.presentFullscreenPlayer() handler do nothing on android, ios works good. Any suggestions?

gergob:
Please share more details, like version of the react-native-video component and a code snippet would be good too @wack17s

xavieramoros:
I have the same problem. 

My react-native-video component is wrapped within a InViewPort component to know if it's visible or not. 

Something like:

```
<InViewPort key={'videoInViewPort'} onChange={this._onPlayerVisibilityChanged}>
	<View 
	key={'videoPlayer'}
	<Video 
		key={'video'}
		ref={(ref) => {
			this.player = ref
		}}
		playInBackground={false}
		paused={this.state.paused} 
		style={{flex: 1}}
		resizeMode={"cover"}
		muted={this.state.muted}
		source={{uri: "http://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4"}} 
	/> 
</View>
</InViewPort>
```


I use the reference to this.player to trigger fullscreen mode (this.player.presentFullscreenPlayer())

I'm using react-native 0.40 and react-native-video 1.0.0. I tested in Android 5.1.1 (Samsung Galaxy J5) and Android 6.0.1 (OnePlus 2 with Oxygen OS)

wack17s:
RN 0.40, RN Video 1.0.0

kevinvandijk:
Can confirm. Tested with RN 0.43.1 with RN Video master on multiple Android devices. The presentFullscreenPlayer does not work on Android.

alixeb:
This does not seem to be working on Windows either.
Also considering that `presentFullscreenPlayer` update the property `fullscreen` to be true.
Why cant we directly put `fullscreen={true}` or (false for that matter) directly on `Video` ?

SebT:
It does work on iOS but not android for me too.
RN 0.42, RN Video 1.0.0, android 6.

mcatmos:
Not working on Android

452MJ:
Not working on Android. +1

>  "dependencies": {
>     "react": "16.0.0-alpha.6",
>     "react-native": "0.44.0",
>     "react-native-video": "^1.0.0",
>     "react-navigation": "^1.0.0-beta.9"
>   },

roancoetzee:
Any update on this?

alvelig:
[...]

PR #890, opened 2018-01-04
Update README to document lack of presentFullscreenPlayer support in …

daisy1754:
#534 has been open for 6+ months. Until it's fixed at least doc should be updated.


cobarx:
I've updated the docs now that we have basic fullscreen support for ExoPlayer

PR #912, opened 2018-01-19
refactor the visualization of documentation

enieber:
this way is easier to find the how to install on different platforms


cobarx:
I've made a bunch of changes to the README to make it a lot easier to read. Thank you for suggesting we change this.

PR #922, opened 2018-01-29
Link exoplayer with cookies

brandonmoon:
Currently, React Native creates a separate instance of OkHttpClient for every module. It does not link up this copy of OkHttpClient to the shared cookies. As a result, video playback fails if it requires cookies for authentication. 

I added code to link up the cookies with our instance of OkHttpClient when created in DataSourceUtils.java so that cookie based auth will still work for video playback.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/DataSourceUtil.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/DataSourceUtil.java
@@ -1,7 +1,11 @@
 package com.brentvatne.exoplayer;
 
 import android.content.Context;
+import android.content.ContextWrapper;
 
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.modules.network.CookieJarContainer;
+import com.facebook.react.modules.network.ForwardingCookieHandler;
 import com.facebook.react.modules.network.OkHttpClientProvider;
 import com.google.android.exoplayer2.ext.okhttp.OkHttpDataSourceFactory;
 import com.google.android.exoplayer2.upstream.DataSource;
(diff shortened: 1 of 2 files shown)


cobarx:
@brandonmoon This is a great feature and I'm looking forward to getting it merged. I'd like to test this before merging it, do you by any chance have a server that uses cookie based auth that I can test it against?

brandonmoon:
@cobarx Thanks! I think I can set you up with something to test with. I'll get back to you in a bit.

brandonmoon:
OK, here's the test case:

1. Make a `GET` request to `https://app.devhv.com/api/v1/videos/188445/` with the header `Authorization: Bearer c11AEGPPhVybEjIhOyZAEdkIlh04abOM836LAqZyfa8=`. It will return a JSON payload and will also set several cloudfront cookies on the device.
2. In the JSON payload returned, extract the `hlsUrl` property and pass that into your video component as the source.
3. With my changes, the video player should send those same cookies with all future requests, which authorizes it to get the video data.

The cookies only last for 15 minutes or so, so you'll need to make sure you make that initial request before playing the video back (it's a rather ugly video of me, sorry!). The bearer token should be valid for 30 days. Let me know if you run into any issues!

Also, for convenience:

```
// Makes initial auth request (saves cookies) and returns the HLS url 
// to hand to the Video component
function getVideoUrl() {
  const url = 'https://app.devhv.com/api/v1/videos/188445/';
  return fetch(url, {
    headers: {
      Authorization: 'Bearer c11AEGPPhVybEjIhOyZAEdkIlh04abOM836LAqZyfa8=',
      'Content-Type': 'application/json'
    }
  })
    .then(response => response.json())
    .then(json => json.hlsUrl);
}
```

cobarx:
Thanks so much for setting that up for me. Everything works great. You've got yourself a merge!

brandonmoon:
Thank you!

PR #926, opened 2018-01-31
add ipod-library support

ymuzi:
should support ipod-library protocol for iPod library

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -732,10 +732,7 @@ static NSString *const timedMetadata = @"timedMetadata";
 - (void)removePlayerLayer
 {
     [_playerLayer removeFromSuperlayer];
-    @try{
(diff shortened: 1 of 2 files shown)


cobarx:
@yasiryang Thanks for this submission! I'm guessing this is an iOS only feature.

How do I go about testing this? I want to provide examples for each of the source types in the README, so some usage instructions would help a lot.

ios/RCTVideo.m
```diff
@@ -732,7 +732,10 @@ - (void)setProgressUpdateInterval:(float)progressUpdateInterval
 - (void)removePlayerLayer
 {
     [_playerLayer removeFromSuperlayer];
-    [_playerLayer removeObserver:self forKeyPath:readyForDisplayKeyPath];
+    @try{
```
cobarx: What's triggering this crash? I'd like to fix the issue where we remove an observer that doesn't exist rather than putting a bandaid over it with try / catch.

PR #943, opened 2018-02-17
Fix playableDuration attribute of onProgress event

paramaggarwal:
Currently we emit `currentTime` and `playableDuration` values on `onProgress` event of the player, but we are incorrectly sending `getBufferedPercentage` when we should be sending `getBufferedPosition` as reflected in the`playableDuration` name.

Reference: https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer2/SimpleExoPlayer.html


cobarx:
@paramaggarwal We need to calculate the playableDuration based on the buffered percentage and seekableDuration. I'm updating your code and also adding `seekableDuration` as a field we return. Thanks for discovering and reporting this bug!

PR #945, opened 2018-02-21
Use setSeek instead of unreliable seek on reference

kentos:
I was expecting repeat to be endless. However the AVPlayerItemDidPlayToEndTimeNotification wasnt fired after the second run. So all this does is to use the class method setSeek instead of the seek on the notification object reference.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -485,8 +485,8 @@ static NSString *const timedMetadata = @"timedMetadata";
   }
 
   if (_repeat) {
-    AVPlayerItem *item = [notification object];
-    [item seekToTime:kCMTimeZero];
+    [self setSeek:0.0];
+    // Not sure this is needed:
     [self applyModifiers];
   }
 }


ios/RCTVideo.m
```diff
@@ -485,8 +485,8 @@ - (void)playerItemDidReachEnd:(NSNotification *)notification
   }
 
   if (_repeat) {
-    AVPlayerItem *item = [notification object];
-    [item seekToTime:kCMTimeZero];
+    [self setSeek:0.0];
```
masarusanjp, line 488: I think that you need to do a comparison of `notification.object` and `_player.currentItem` here, or call `[notificationCenter removeObserver:self name:AVPlayerItemDidPlayToEndTimeNotification object:_player.currentItem]` at here https://github.com/kentos/react-native-video/blob/62ba9f18da218e87fb3d10ec7a6f421eb39f35f2/ios/RCTVideo.m#L285 in order not to process notifications of other videos

PR #952, opened 2018-02-28
Can not play videos from local file system on Android 4.1.x and 4.2.x

LukaszC:


LukaszC:
I forgot to comment: as far as I remember this simple hack will break exoplayer, then it seems that it has to be fixed in the android component itself.

PR #959, opened 2018-03-07
Allow other audio recorders in the same app to run nicely while the video player is playing.

iMagdy:
Allow other audio recorders in the same app to run nicely while the video player is playing.

The problem is without the option `AVAudioSessionCategoryOptionMixWithOthers` to allow mixing with others, the video player will automatically interrupt any other audio session (recording activity for example).

This PR will let the player play, and if the app need to record something, it will be able to do without interruption.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -529,9 +529,9 @@ static NSString *const timedMetadata = @"timedMetadata";
     [_player setRate:0.0];
   } else {
     if([_ignoreSilentSwitch isEqualToString:@"ignore"]) {
-      [[AVAudioSession sharedInstance] setCategory:AVAudioSessionCategoryPlayback error:nil];
+      [[AVAudioSession sharedInstance] setCategory:AVAudioSessionCategoryPlayback withOptions:AVAudioSessionCategoryOptionMixWithOthers error:nil];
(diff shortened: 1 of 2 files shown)


cobarx:
@iMagdy I'm not sure this will have the intended effect. The Apple docs suggest that setting this flag will allow any other audio apps to mix their audio with the player. If a customer has music playing from another app, it would keep playing rather than pausing when you start the video.

This needs to be a configurable setting using a prop, since in most cases the video should have exclusive access to audio.

iMagdy:
@cobarx You're right it'll mix with other players/recorders. What's needed here it to mix with other audio recorders.

I agree this should be a configurable prop, maybe in another PR.

evanjmg:
@iMagdy can you make this configurable? If not I'll try to fork and PR

zi6xuan:
When a dynamic background is needed, mp4 is obviously better than gif. In this case, mp4 is displayed as a background or dynamic picture, and does not require sound, and does not need to interrupt the audio being played, such as listening. When the song opens, the app with dynamic background mp4 is opened.

zi6xuan:
It seems that someone has already proposed pr, I don’t know when it will merge.

PR #972, opened 2018-03-15
add exoplayer subtitles attributte

sungjkim34:
When using the exoplayer, the `subtitle` boolean attribute will now toggle on or off the subtitles.  
This works especially great with a HLS with embedded subtitles.


huanghanzhilian:
How to use it, please

sungjkim34:
Just add a boolean to the subtitle prop.
`<Video subtitle={true} />`

sungjkim34:
No unfortunately this PR is specifically only for subtitles attached directly to the HLS stream.
I would possibly recommend attaching a WebVTT subtitle to your HLS stream if you want to use the code from this fix.

cobarx:
I need support for external subtitles as well and should be able to put together a PR for ExoPlayer support shortly. It would make a lot of sense to have a unified API to control both embedded & external subtitles. Trying to stay close to the HTML5 video tag (https://developer.mozilla.org/en-US/Apps/Fundamentals/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video), I came up with:

```
<Video source={{ uri: "https://example.com/video.mp4" }}
  tracks={[
    { label: "English", type: "vtt", uri: "https://example.com/video.en.vtt" },
    { label: "Spanish", type: "srt", uri: "https://example.com/video.es.srt" }
  ]}
  // One of "none", "embedded", or a label. Defaults to none. Maybe should be called selectedCaption
  selectedTrack="English" 
  />
```

I think we want to be able to switch between tracks without reloading the video or loading subtitles on the fly, so this approach would make it easier. Looping in a team member @akshaygore06 for thoughts.

akshaygore06:
@cobarx: I like your approach of passing the list of available tracks with the video source. I think this will be a good approach to solve the subtitles sideloading issue.

cobarx:
@petrbela Not seeing anything else we need to support at the moment, we can always add more attributes later.

Does anyone know how to control the order that the props are passed to the native module? ExoPlayer only supports loading subtitles before the video starts and React Native is passing the source prop before the track prop.

cobarx:
I have a pretty clear idea of what's required to implement robust text track support for both iOS & ExoPlayer. I've documented my ideas in #1044, please provide feedback. I should have be able to get embedded caption support done in about a week.

cobarx:
My implementation is now ready in #1049.

cobarx:
Going to close this since we have a more robust implementation in the other PR. I'll open another PR in the next few weeks to support sidecar subtitles.

PR #974, opened 2018-03-19
Fixes #963

### LINKED ISSUE #963: Crashed(ios): to call setSrc twice
Related pr: https://github.com/react-native-community/react-native-video/pull/600 

Crash occured at here: 
https://github.com/react-native-community/react-native-video/blob/master/ios/RCTVideo.m#L735

Called from: 
https://github.com/react-native-community/react-native-video/blob/master/ios/RCTVideo.m#L277

Because `readyForDisplayKeyPath` observer had been removed in `removePlayerTimeObserver`.
https://github.com/react-native-community/react-native-video/blob/master/ios/RCTVideo.m#L271

So, when calling `removePlayerLayer`, an exception occurred by removing `readyForDisplayKeyPath` in multiple.

CC: @awitherow

wildseansy:
Fixes #963 
Bug happens when uri is changed for a video that is already rendered. This is because `[_playerLayer removeObserver:self forKeyPath:readyForDisplayKeyPath]` is called twice because the `_playerItemObserversSet` variable isn't properly set in sequence.

## The fix:
Remove player layer before addPlayerItemObservers so `_playerItemObserversSet` is still set to `NO` if observers have already been removed.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -140,8 +140,8 @@ static NSString *const timedMetadata = @"timedMetadata";
 - (void)dealloc
 {
   [[NSNotificationCenter defaultCenter] removeObserver:self];
-  [self removePlayerItemObservers];
   [self removePlayerLayer];
+  [self removePlayerItemObservers];
   [_player removeObserver:self forKeyPath:playbackRate context:nil];
 }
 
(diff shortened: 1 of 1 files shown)


masarusanjp:
Thanks for fixing!

Could you please remove these lines too?

https://github.com/wildseansy/react-native-video/blob/bfa2969c5f8a295b4abb060b59e8b585580a3d96/ios/RCTVideo.m#L255-L257

I think these lines are no longer needed by this pr.

wildseansy:
@masarusanjp - updated.  Also re-ordered [these lines](https://github.com/react-native-community/react-native-video/pull/974/files#diff-0fe2a211e3c9f287e96b339e47c5bd1aL142) so `removePlayerItemObservers` consistently comes after `removePlayerLayer`

PR #983, opened 2018-04-03
Upgrade ExoPlayer to 2.7.2

peacechen:


cobarx:
Can we merge this? I haven't run into any issues with this in my testing.

Winglonelion:
exoplayer has some conflict with React Native Fetch Blob

cobarx:
@Winglonelion What sort of conflict are you running into? I tested with React Native Fetch Blob and didn't encounter any issues. Can you attach an error log so I can try and track it down.

Winglonelion:
Build is ok but on runtime App crash with a NoClassDefFoundError of RNFetchBlobPackage. May be by my buildTarget and buildTool is set to api 26. Everything ok with AndroidMediaPlayer

cobarx:
Hmm, I have no idea what would cause that. We are compiling with SDK 25 and targeting 22 in our project, but I believe react-native-video will override the project settings and use 23.

Can you include the full stack trace that may help. It also might be worth opening an issue with react-native-fetch-blob to see if they have any ideas.

peacechen:
Could be a Google Play Services mismatch.  Similar issue reported on RN maps in the past. 

https://github.com/react-community/react-native-maps/issues/894

Winglonelion:
Additional, ```NoClassDefFoundError``` only happen  in api 22 and lower. I will send you Log after bit of time late. I am having lunch now.

peacechen:
Updated Exoplayer version.  Haven't had a chance to test it yet.

cobarx:
@Winglonelion Can you send over that log, I'd really like to get this merged.

cobarx:
@Winglonelion Thank you for sending this over. When you say that you're using SDK 22, do you mean the defaults like React Native uses?
compile SDK 23
target SDK 22

cobarx:
@Winglonelion I'm at a loss as to how to fix it. I've run the app on both a 4.4.4 and 5.1.0 emulator and it works fine on both of them.

There are a couple things you could try.

add:
```
    dexOptions {
        jumboMode = true
    }
```
to the android section of your build.grade

Change the exoplayer build.grade to use the same SDK & build tool versions as the rest of the project.

If neither of those works, I think it would be worth opening a ticket with react-native-fetch-blob as there's nothing in react-native-video or exoplayer that should be causing this issue.

Winglonelion:
@cobarx, do you tried it with build tool 26 and target 26 ?  i just tried the official `react-native-video` repo, and get the same error. I think it start from my build config. 

The exoplayer already use same SDK and build tool version with main project.

cobarx:
@Winglonelion Yes I tried with target & compile SDK 26 on both the project & exoplayer. Also tried with a real device on 5.0.2. Still no issues.

I think you're right, something is going on with your build config.

I'm going to go ahead and merge this. If you figure out what happened, please let me know and we can add a fix.

Winglonelion:
@cobarx Sure, thanks a lot

PR #871, opened 2017-12-07
Add requiresMainQueueSetup method

habovh:
Since RN 0.49, `requiresMainQueueSetup` needs to be defined if the module overrides `constantsToExport`.

Closes #820

PR #998, opened 2018-04-24
Update RCTVideoManager.m

anhtuank7c:
Module RCTVideoManager requires main queue setup since it overrides `constantsToExport` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.


alexpareto:
👍 also getting this Yellowbox

cobarx:
This has been fixed with #871

PR #1015, opened 2018-05-15
Support for android assets

roadster01:
assets://path.to.assets.in.src.main


ascorbic:
React Native supports the asset:// scheme (via Fresco) rather than assets://, so it might be better to use that (or support both). It would also be good to support file:///android_asset/ as used in webview.

cobarx:
Going to close this since we already have `asset://` support. Please re-open if there's a reason we need this.

PR #1021, opened 2018-05-18
React is missing as peer dependency in package.json

futhrevo:
React should be listed as peer dependency else getting the error ```Unknown module 'react' from haste map```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,11 @@
     },
     "dependencies": {
         "keymirror": "0.1.1",
-        "prop-types": "^15.5.10"
     },
+    "peerDependencies": {
+        "react": "^15.0 || ^16.0",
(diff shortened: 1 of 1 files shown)


cobarx:
When do you see this error, I haven't encountered it.

futhrevo:
Steps for reproduction
1. Clone the repo
2. go to example folder and run `react-native start` packager and build with android studio
3. Error saying "Unknown module 'react'

cobarx:
Thanks for pointing this out. Even with this PR, I still get an error about the `react` module being missing in the Haste map. I'm not sure what's going on.

I'm going to try doing a clean project that contains the example files with the latest versions of React Native & React and see if I can get it to work.

PR #1027, opened 2018-05-22
windows - send same seek event playload as on android

mrousal:


cobarx:
I'm new to react-native-windows, so I'm not sure how doing events is supposed to work. When is it appropriate to use RunOnDispatcher vs. not?

I looked and on the Net46 version, MediaPlayer doesn't support a SeekCompleted event. Can you update the code to handle this, you can probably fake it by passing in the values from the `Seek` function to `OnSeekCompleted`. You can also remove:
https://github.com/mrousal/react-native-video/blob/c157da0d4e40d8a7be42e218d3602e39e97dedb9/windows/ReactNativeVideo.Net46/ReactVideoView.cs#L160
https://github.com/mrousal/react-native-video/blob/c157da0d4e40d8a7be42e218d3602e39e97dedb9/windows/ReactNativeVideo.Net46/ReactVideoView.cs#L183

cobarx:
@mrousal Any chance you can update the WPF code? It would be great to get this merged.

cobarx:
@mrousal Wanted to send you another ping in case you didn't see the early message. The WPF code won't work as written and needs to be updated.

mrousal:
Hi, sorry I was offline a bit. What exactly is the problem for WPF part, I can't build it easily atm? I have removed that RunOnDispatcher call as it should not be needed in this case, but I don't see any other issue.

PR #1037, opened 2018-05-29
Make ExoPlayer send onEnd when repeat is on

cobarx:
ExoPlayer doesn't generate a state change when the end of a video is reached if repeat mode is turned on. So we need to use onPositionDiscontinuity to detect it as described here in ExoPlayer issue https://github.com/google/ExoPlayer/issues/3252


dfattlar:
@cobarx it seems that this fix no longer works? I'm not seeing onEnd fire when repeat is true. I've tested the scenario with the "basic" example app as well.

PR #1039, opened 2018-05-29
adds audioOnly option that always displays poster image (if provided)…

shlokamin:
… on top of player


shlokamin:
I use this component to load multiple types of media formats (MPEG4, AACAudio, MP3, etc) all in the same screen. It is convenient to be able to just have a poster image be displayed on the player when playing an audio format. 

Let me know your thoughts!

cobarx:
Ok, that seems like a reasonable use case. Merged!

PR #1043, opened 2018-05-31
Add: Custom styles for Poster via a posterStyles prop

tomchinery:
I needed to be able to style the Poster image so I've added a `posterStyles` prop. This allows you direct control over the styles of your poster image. 

Let me know your thoughts?


cobarx:
Hi @tomchinery, can you give me a little idea of where this comes in handy? I haven't used the poster feature that much myself.

PR #805, opened 2017-10-02
added optional request headers for remote assests (android & ios)

emrahdk:
This pull request adds support for setting arbitrary `Request Headers` for resolving remote media. A use case could be to add an `Authorization` header for media that needs authentication. Therefore this PR should be able to close #475.

It can be used as demonstrated in the snippet below, by adding a `headers` object in the `source` property of the `Video` component.

```jsx
<Video
  source={{
    uri: videoUrl,                        .
      headers: {
        Authorization: 'bearer some-token-value',
        'X-Custom-Header': 'some value'
      }
  }}
</Video>
```

It has been tested on iOS and Android. 

_Note: It was currently not possible to add the same functionality for Windows. `System.Windows.Media.MediaPlayer` only supports a single `Uri` class for remote/local media._

PR #1046, opened 2018-06-01
Add headers back to android remote assets

djw27:
In the following commit, headers were removed from the source in Android, meaning that cookies are not set with the data source on Android (non-exoplayer).

The commit also changed the user of `parsedUrl` for `uriString` which is less safe.

I'm not sure why this commit was originally accepted! 

For context, our use case is accessing a HLS video on CloudFront which uses signed parameters to keep the assets private. We're required to set the signing parameters as cookies so that the HLS stream plays.


cobarx:
Hi @djw27, I went through the commit log and found that it was changed with:
https://github.com/react-native-community/react-native-video/pull/537

It sounds like reverting the commit will have issues as well since the ContentProvider doesn't get initialized. I'm not too familiar with this part of the code, do you have any ideas how we fix the other issue?

djw27:
@cobarx sorry I meant to include that commit hash in my PR!

I'll look into it a little further, although I can say that we've been using this in production for 6 months now and had no issues.

cobarx:
No idea what `An IOException was constantly being thrown from the android Media player classes, because instead of only a URL, the media player, for a dataSource was expecting a whole ContentProvider initialised with context`. I'm guessing a specific URL had issues. We support a lot of different source types, maybe it's one of the obscure ones that I haven't tested yet.

@gorjanz Hi, do you happen to know which source you were using that caused the IOException? Removing that code causes us to lose the ability to specify HTTP headers.

cobarx:
This is going to get returned back to the old behavior in #805. We'll take our changes on re-introducing the issue and make sure to fix it right away if someone reports a problem.

cobarx:
Ok found a URL that runs into a similar error:
https://raw.githubusercontent.com/jwplayer/jwdeveloper-demos/master/demos/basic/audio-metadata/assets/index.m3u8

It throws a FileNotFound exception due to no content provider. However it's not fatal and the media still plays. So hopefully everything keeps working...

djw27:
Good find @cobarx. Thanks for the progress with this

PR #1048, opened 2018-06-02
Fix partial swipe on video in fullscreen shows a black screen

Ahmed-Emad:
In iOS 11, Apple added a feature to close the Video Player via a swipe gesture. So when partially swiping, the viewWillDisappear function gets called but viewDidDisappear is not called which shows a black screen.


cobarx:
Thank you, going to merge this.

PR #1058, opened 2018-06-08
ExoPlayer TextureView support

cobarx:
Provide support for selecting outputting to a TextureView rather than the default SurfaceView. This is useful in certain situations where you want to animate the video view or need to layer multiple videos on top of each other.

Based on code from #852 by @asdmikk


ma-pe:
this is great - thank you!

PR #1071, opened 2018-06-18
prevent crash fixes #907

### LINKED ISSUE #907: Error UI block RCTVideo 0x123e3f990
I getting two error when I switch m3u8 link in iOS, Android Work Perfectly.

`Exception thrown while executing UI block: Cannot remove an observer <RCTVideo 0x123e3f990> for the key path "readyForDisplay" from <AVPlayerLayer 0x1c4625060> because it is not registered as an observer.`

`Exception thrown while executing UI block: 'parentNode' is a required parameter`

n1ru4l:
Fixes the crash described in #907

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -732,7 +732,9 @@ static NSString *const timedMetadata = @"timedMetadata";
 - (void)removePlayerLayer
 {
     [_playerLayer removeFromSuperlayer];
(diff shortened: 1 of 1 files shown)


cobarx:
Thanks for submitting this. I've created #1075 to handle this a bit more cleanly.

n1ru4l:
Thank you for taking care of this issue!

PR #1072, opened 2018-06-19
Fix(ios): crash in removePlayerLayer keyPath observer (#907)

ahmetardal:


cobarx:
I've created #1075 to handle this a bit more cleanly.

Issue #1067, opened 2018-06-14
HLS stream UI block when loading a broken URL (iOS)


HeinzStaschen:
### Current behavior
Screen doesn't render until the onError callback gets triggered.

### Reproduction steps
Set the Video component source uri to a broken https m3u8 URL.

### Expected behavior
Screen needs to render normally so I can display a loader and then handle the onError event if needed.

### Platform
* iOS

### Video sample
https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8

I've investigated RCTVideo.m and noticed you guys use AVURLAsset (which I suspect might be the cause of the issue):
```
if (isNetwork) {
    NSArray *cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookies];
    AVURLAsset *asset = [AVURLAsset URLAssetWithURL:url options:@{AVURLAssetHTTPCookiesKey : cookies}];
    return [AVPlayerItem playerItemWithAsset:asset];
  }
```

Refer to link for possible solution:
https://stackoverflow.com/questions/38777384/avurlasset-urlassetwithurloptions-blocks-main-thread-with-remote-url

danielmcmillan:
Do you still get this issue if you don't specify any one of the props 'ignoreSilentSwitch', 'rate', 'volume' or 'muted'?
If not than my pull request #1086 may fix the issue. Potentially the cause is what you have referenced and this just happens to work around it though.

namgk:
I have the same problem, but I have 'volume' prop on. Will try to remove it.

treemore:
use charles.  start throttling . the ios ui will block util buffed success

cobarx:
Are people still having this issue? 

I tested the PR in #1113 and it has the side effect of causing the video to show up if you start the video paused. I think the solution @HeinzStaschen proposed is the right way to handle this, but it will take a fair bit of work to implement that. So lmk if that rework is needed.

StuartMorris0:
This issue still exists for me on iOS.

I only have props for source, onError, style and controls. The video locks up the main UI thread when loading the video, once the video is loaded it is released.

BaderSerhan:
@StuartMorris0 are you starting the videos paused?

StuartMorris0:
I think they autoplay but I cannot remember 100%.

muganwas:
I'm using version 5.0.2 and I'm still experiencing this issue. Is there a solution?

Issue #577, opened 2017-04-19
Application Freeze when loading the video


janakact:
If we are trying to play a video from a web URL, the whole application stay stuck (About 2-3 seconds until ) the a portion of the video is buffered. Seems like the whole UI rendering thread in blocked. Is there a solution for this or is it a bug?

janakact:
Tried different versions. Currently using,

```
    "react-native": "^0.38.0",
    "react-native-video": "github:react-native-community/react-native-video#9bcf70cbb513402622bc114f93da751c28f15303"
```

dquak:
Same here!
all UI is feezing for about 5-10 secounds before showing the video.
Tried @janakact solution with no luck:
`
"react-native": "0.42.3",
    "react-native-video": "github:react-native-community/react-native-video#9bcf70cbb513402622bc114f93da751c28f15303",
`

kamalk333:
I faced the same Issue. Application freezes up because the video loading takes place on the main thread.
Try degrading the version of react-native-video to 0.9.0

It will work.

dquak:
@kamalk333  Thanks!! This fixed this issue, - BTW version 1.0.0 works as well for me.
When setting some mp4 file in the uri, it seem that the player is not buffering, but waiting for the video to fully loaded before start playing it, any idea of how to fix it?

janakact:
Solved. If you are using an older version, 
try changing the following file.  Change `prepare(this);` => `prepareAsync(this);` 
https://github.com/ttdat89/demo-video-streaming/blob/master/node_modules/react-native-video/android/src/main/java/com/brentvatne/react/ReactVideoView.java#L301

Thanks [Tien Dat](https://www.facebook.com/TranTienDat?fref=ufi) for the help.

dquak:
@janakact Amazing, would really like to know how you got to this solution, as I would really like to fix the issue on the latest version

mesbahsabur:
+1

janakact:
@itay9001 
Got it from [FB React Native Community group. You should contact [Tien Dat](https://www.facebook.com/TranTienDat?fref=ufi)
 Hope it will be helpful. 
[https://www.facebook.com/groups/react.native.community/permalink/971188333016745/](https://www.facebook.com/groups/react.native.community/permalink/971188333016745/)

carbureted:
I'm having this issue on iOS (javascript FPS drops to 0 for a second or two when the video starts) - are there any fixes floating around for it?

[...]

PR #1086, opened 2018-06-23
Fix UI block on iOS during initial load

danielmcmillan:
Possibly related issues: #577 #1067.
I'm experiencing an issue where, if any prop that causes the applyModifiers function to be called has a value set, then the UI will block during the initial video load from a network location. This is the 'ignoreSilentSwitch', 'rate', 'volume' and 'muted' props.
I can reproduce this by pointing the player to a remote file such as 'https://www.sample-videos.com/video/mp4/720/big_buck_bunny_720p_10mb.mp4' and setting, for example, volume to 1. If the player is mounted when the app starts, then no UI will render until the video loads. Removing the volume prop prevents the issue.

I fixed the issue by adding a check that the video is ready in the applyModifiers function. The function is called again when the video state changes to ready so I don't think this causes any problems.

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -639,6 +639,11 @@ static NSString *const timedMetadata = @"timedMetadata";
 
 - (void)applyModifiers
 {
+  // Skip applying modifiers if the video is not ready to avoid freezing
+  if (_playerItem.status != AVPlayerItemStatusReadyToPlay) {
+    return;
+  }
+
   if (_muted) {
     [_player setVolume:0];
     [_player setMuted:YES];


cobarx:
Thanks for suggesting this. I did a fair bit of testing around it and this has the effect of changing the styling both while it's loading and afterwards since the resizeMode modifier doesn't get applied.

What I'd suggest is filtering out the modifiers that cause it to hang so we only avoid apply those ones.

danielmcmillan:
@cobarx Okay. Thanks for the suggestion, I'll look into it and update the PR when I can.

iloveyumiforever:
@cobarx @danielmcmillan , I have created a new PR #1113, I adopt your suggestion to filter out specific modifiers to fix this issue, and also I didn't get the meaning of "effect of changing the styling" you mentioned clearly, so I haven't find any side-effect, could you check my code and give me more specific details of the side-effect you mentioned or response/suggestion if there's exist any problem actually , otherwise if my code without any problem , could you merge this update and release a new version as soon as possible, because this bug cause so many bad user experience that our team is sticked in this problem, thanks for your helper, I am looking forward your response.

cobarx:
@danielmcmillan Pretty sure you're right, that PR makes it so that all the props get set before it attempts to set the source and start loading the video. Not sure why that fixes it, but we can take the win! Thank you for letting me know, I've released 3.1.0 with that change and am closing this.

BaderSerhan:
Hi. I have been facing the exact same issue and working on my app for the past 3 days thinking it was a structural problem only to have it fixed by removing the volume/muted attributes! This saved my day, but now the problem that arises is when will this be merged cause volume is crucial for my app!

zek:
Why this issue closed? I've got the same issue on controls prop.
I applied same patch on
~~~
- (void)setControls:(BOOL)controls
{
    // Skip applying modifiers if the video is not ready to avoid freezing
    if (_playerItem.status != AVPlayerItemStatusReadyToPlay) {
        return;
    }
....
~~~

That fixes the issue.

PR #1088, opened 2018-06-25
Prevent iOS onLoad event during seeking

cobarx:
onLoad should only fire when the video is loaded, however it was firing at other times on iOS like when you performed a seek.


laurent22:
Is there any plan to merge this? The onLoad event is indeed firing every time when seeking the video, which can be a major issue when your rely on this event. In our case, we seek during the onLoad event, which works fine in Android but results in a loop in iOS.

PR #1109, opened 2018-07-04
iOS Side loading for captions and offline support

ashnfb:
Adds support for side-loading captions in iOS for both streaming and offline resources.


cobarx:
@ashnfb Would you mind merging in the latest master? There's quite a bit that's changed in the main tree.

ashnfb:
@cobarx I've merged upstream into this fork.  Should be good to go!

cobarx (changes requested):
Here are some initial notes. I haven't gotten it to work quite yet.

Let me know on any of these if you need me to make the changes.

cobarx:
Thank you for all your hard work on this!

Ok, I started testing this and it's getting too late to figure out what I'm missing.

A couple questions:
Do you know which formats are supported? The ExoPlayer versions currently supports VTT, TTML, and SRT. If not, I have test files for all 3 and can figure it out.

Is it feasible to support more than one mutable text track? It's a pretty common situation where you have a menu and can select between English, French, Spanish, etc.

Eventually, I'd like to see about making it possible to support sideloaded and streaming captions at the same time. I'm not sure it's super useful so it's not important atm. I can do the work to handle both at some point in the future.

ios/RCTVideo.m
```diff
@@ -284,53 +286,76 @@ - (void)setSrc:(NSDictionary *)source
...
+  NSString *subtitleUri = self.selectedTextTrack[@"uri"];
   NSString *type = [source objectForKey:@"type"];
   NSDictionary *headers = [source objectForKey:@"requestHeaders"];
-  
-  NSURL *url = (isNetwork || isAsset) ?
-    [NSURL URLWithString:uri] :
-    [[NSURL alloc] initFileURLWithPath:[[NSBundle mainBundle] pathForResource:uri ofType:type]];
+
+  AVURLAsset *asset;
+  AVURLAsset *subAsset;
```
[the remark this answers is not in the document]
ashnfb: fixed

ios/RCTVideo.m
```diff
@@ -710,23 +835,30 @@ - (void)setSelectedTextTrack:(NSDictionary *)selectedTextTrack {
...
+- (void)setTextTracks:(NSArray*) textTracks;
+{
+  _textTracks = textTracks;
+}
+
 - (NSArray *)getTextTrackInfo
 {
+  
+  // if sideloaded, textTracks will already be set
+  if (_textTracks) return _textTracks;
```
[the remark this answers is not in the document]
ashnfb: I'm unclear how this works, so I've not attempted to do this.

ashnfb:
Hi @cobarx, I've pushed fixes in the PR for iOS and Android.

Replies to some of your questions:

> Do you know which formats are supported? The ExoPlayer versions currently supports VTT, TTML, and SRT. If not, I have test files for all 3 and can figure it out.

It seems like only VTT.  I also tried .txtt and that didn't work.

>Is it feasible to support more than one mutable text track? It's a pretty common situation where you have a menu and can select between English, French, Spanish, etc.

From what I read, AVMutableCompositionTrack doesn't support AVMediaSelectionOption, which is what is used to switch between languages for Streaming video (AVMediaSelection also has built-in UI support, so that's a bummer we can't use it).   We would have to build our own UI for switching tracks, and swap out the Mutable track for one language with another.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -103,7 +105,7 @@
     private boolean loadVideoStarted;
     private boolean isFullscreen;
     private boolean isInBackground;
-    private boolean isPaused;
+    private boolean isPaused = true;
```
cobarx: This looks like a change that got wiped out during the merge. Please revert.

ios/RCTVideo.m
```diff
@@ -284,53 +288,76 @@ - (void)setSrc:(NSDictionary *)source
...
+    
+    _player = [AVPlayer playerWithPlayerItem:_playerItem];
+    _player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
+    
+    [_player addObserver:self forKeyPath:playbackRate options:0 context:nil];
+    _playbackRateObserverRegistered = YES;
+    
+    [self addPlayerTimeObserver];
+
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
cobarx: We are already in a delayed run loop, so I don't think we need a second one.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -762,7 +764,26 @@ public void setSelectedTextTrack(String type, Dynamic value) {
             trackIndex = value.asInt();
         } else { // default. invalid type or "system"
             trackSelector.clearSelectionOverrides(index);
```
cobarx: Do we need this or the setSelectionOverride anymore? If we are SDK <= 18 or no groups, the trackIndex will be C.INDEX_UNSET and we'll clear the overrides further down.

cobarx:
I have a few hours free this afternoon, so I'm going to start working on the text track selection for iOS piece. I wanted to make sure we don't duplicate effort. I'll post what I have at the end of the day.

ashnfb:
@cobarx thanks for the heads up.  I'll comment on some of the iOS changes above, and I'll look into the Android changes then.

ios/RCTVideo.m
ashnfb: Agreed, the extra dispatch can be removed.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
ashnfb: @cobarx we need the setSelectionOverride in cases where captions were previously enabled, and are now explicitly disabled (or the system settings are disabled).  I've moved this up further in the code for clarity.

cobarx:
Ok I have the selection part working in 93ce4d6. Can you merge that into this PR.

I am seeing one issue which is that if you specify `textTracks` it screws up HLS playlists since they can contain video, audio, and text tracks. I'm looking to see if there is a way to fix that.

cobarx:
@ashnfb Ok found a Stack Overflow that suggests it's not possible to use HLS streams in a mutable composition since tracks can be added & removed throughout playback:
https://stackoverflow.com/questions/8318422/inserting-an-http-stream-into-a-avmutablecomposition

So I think this is ready to merge as soon as you pull in my commit and remove the redundant dispatch call.

cobarx:
Pretty sure it all works, but would appreciate a review on my changes as well.

cobarx:
I will go ahead and document everything after we're done.

ashnfb:
Ok, reviewing and merging now.

ashnfb:
@cobarx I've merged your changes, but unfortunately sideloading isn't working for me. It looks like _player is null when setSideloadedText is being called, which results in no text being enabled?

cobarx:
What I was seeing is that setSelectedTrack was called multiple times. At first, _player is null but eventually everything is set, applyModifiers gets called, and then it enables the right stuff.

cobarx:
Is it feasible to send me a copy of the code you use on the React side to create the player? Props get set in the native code in different orders depending on how you pass the props in. If not, I can probably trial and error it until I get the right combination.

ashnfb:
@cobarx invite sent to the repo.  Thanks for offering to look into the problem.  With the app, turn on Captions through device settings, and use the Search button to find film "Am Stram" (not all films have captions).  In the React code, the props are sent by FilmDetail.js

cobarx:
@ashnfb Got it, thanks. This is helpful.

I found one issue. I have English set on my simulator and the captions are French. It was setting the wrong default index when there wasn't a caption matching the system language. Once I made the change, the captions showed up. I've committed my fix to this PR.

I checked the code and setSelectedTextTrack is called as part of applyModifiers once the onLoad event fires, so I'm guessing the issue that I encountered is the same as yours and that _player being nil is not the problem. Please test against and let me know if there's still a problem.

cobarx (approved):

ashnfb:
@cobarx thanks for investigating and solving the problem!  works fine here too, Ready for merge 👍

cobarx:
Awesome, it's great to have full subtitle support!

cvaldezscse:
@cobarx Thanks for the hardwork, but i haven't understood where do we put the Url of the subtitle file, i mean, could you make an example or demo, of how does it work? or the correct steps to implement this feature

PR #1113, opened 2018-07-08
fix ios loading freeze

iloveyumiforever:
fix ios loading freeze in #1086, adopting suggestion of filtering out the modifiers that cause it to hang of cobarx

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -548,7 +548,7 @@ static NSString *const timedMetadata = @"timedMetadata";
 - (void)setIgnoreSilentSwitch:(NSString *)ignoreSilentSwitch
 {
   _ignoreSilentSwitch = ignoreSilentSwitch;
-  [self applyModifiers];
+  [self applyModifiersWithFilter];
 }
 
 - (void)setPaused:(BOOL)paused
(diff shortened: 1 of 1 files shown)


treemore:
this work for me . 
Using nginx
Limit_rate 1k; This problem can be reproduced in the development environment; this pull request can solve our problem. No other impact has been found yet.

brjennin:
I was seeing the same issue and this seems to have fixed it. Would be great to see this merged in to master.

Thanks

danielmcmillan:
I think the issue no longer exists on master. Can anyone else confirm that?

cobarx:
Tested this and if you have the video set to start paused, it has the effect of causing the video to appear on screen whereas before it didn't. I also simplified the code, what I was looking for was that since paused, rate, volume, and ignoreSilentSwitch are the only offenders, we don't apply those ones.

Is this still an issue as @danielmcmillan mentioned?

@HeinzStaschen found a StackOverflow post that suggests the problem is likely that AVURLAsset isn't async by default so it blocks the UI:
https://stackoverflow.com/questions/38777384/avurlasset-urlassetwithurloptions-blocks-main-thread-with-remote-url
That's the right approach since it doesn't have side effects like preventing props from being loaded. However, it's a fair bit of work to refactor the code to do everything async. I'll get around to it eventually, but if someone can do it sooner that would be great.

Dumidu:
I m facing the same issue on iOS. Is there any workaround ?

PR #1117, opened 2018-07-10
Android - Keep screen on

geriux:
Even though `setScreenOnWhilePlaying` is set to `true` in the MediaPlayer, on Android the screen will dim depending on the time the user has set on their phone. 

It was already reported in this issue https://github.com/react-native-community/react-native-video/issues/744.

This will prevent the screen to turn off while a video is playing and it only affects the Android platform.


fluiddot (approved):
:100:

cobarx:
Thanks for pointing this out and creating a fix.

I haven't gotten to test this, so please excuse if I'm not up to speed. I think the ideal behavior is that the screen stays on while you're playing the video, but if you pause the video the screen should be allowed to go to sleep as that's what other video apps do. If we set this flag, will the screen go to sleep while paused?

Also, is this a problem on ExoPlayer as well? If so, would you mind creating a PR for that.

geriux:
You're right @cobarx the screen should be allowed to go to sleep if the video is paused so I changed the code to take that into account as well :) let me know what you think.

As for the ExoPlayer I can't check it at the moment but I'll try to soon.

Thanks!

cobarx:
Thanks for tackling the pause conditions. Is there a reason we need to check the current state of the flag rather than applying the appropriate value any time we play or pause? It adds quite a bit of code & complexity unless there's a compelling need for it (like performance).

cobarx:
@geriux When we added support for sideloaded captions, it also reworked how the video setup happens. From https://github.com/react-native-community/react-native-video/pull/1086#issuecomment-404681326, @danielmcmillan is no longer experiencing the issue. I've gone ahead and released 3.1.0 with that change, can you test again and see if you're having the problem?

geriux:
Hi! I had a comment in my e-mail about this being fixed in another PR but i've just tested it and it **is still happening** with the latest version. Also I don't see your comment here :D do you still want me to continue with this PR? Let me know!

cobarx:
@geriux I posted in the wrong PR, that comment was meant to go in the one about the iOS hanging while loading.

Yes, I would like to get this included. Can you simplify the logic so that it's more straightforward? Unless there's a good reason to do otherwise, it's much simpler to set and clear the flag every time rather than tracking the current state.

djw27:
See here for how it could potentially be simplified: https://github.com/corbt/react-native-keep-awake/blob/master/android/src/main/java/com/corbt/keepawake/KCKeepAwake.java

geriux:
@cobarx alright! I'll simplify it and update the PR =)

android/src/main/java/com/brentvatne/react/ReactVideoView.java
```diff
@@ -457,6 +459,42 @@ public void setControls(boolean controls) {
...
+
+        if ((flags & WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON) == 0) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public void setPreventScreenFromDimmingFlag(final boolean state) {
+        if (!mMediaPlayerValid && mThemedReactContext == null) {
```
cobarx: This should be `!mMediaPlayerValid || mThemedReactContext == null`

android/src/main/java/com/brentvatne/react/ReactVideoView.java
```diff
@@ -457,6 +459,42 @@ public void setControls(boolean controls) {
...
+            return;
+        }
+
+        final boolean isFlagOn = isPreventScreenFromDimmingFlagOn();
```
cobarx: I would remove this check and simply apply the flag every time.

android/src/main/java/com/brentvatne/react/ReactVideoView.java
```diff
@@ -457,6 +459,42 @@ public void setControls(boolean controls) {
         this.mUseNativeControls = controls;
     }
 
+    public boolean isPreventScreenFromDimmingFlagOn() {
+        int flags = mThemedReactContext.getCurrentActivity().getWindow().getAttributes().flags;
+
+        if ((flags & WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON) == 0) {
```
cobarx: If we need to keep this check for performance or other reasons, it can be written as:
`return flags & WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON == 0`

cobarx (changes requested):
I've added a fix for one bug and also included some examples of how to simplify the code.

geriux:
Thank you for the suggestions @cobarx ! I've fixed the bug and removed the check of the flag. Let me know what you think =)

cobarx:
Sorry for the delay in getting this merged. I kept forgetting to bring along my cable when going to coffee shops and there's no way to test without using a real device #smh  

I went ahead and switched this over to setKeepScreenOn since it's much simpler and matches what we're using on the ExoPlayer side. We also need to check for when the video ends and allow the screen to sleep if the video isn't on repeat. I tested and it works great. 

I'm glad you brought this up, this is pretty important behavior, I have no idea how people got around this. Maybe everybody was using other modules to control sleeping.

tkstang:
Is there a prop I need to pass in for this fix to work? I am still having this issue.

auseika:
Would be great to have a way to disable this, since sometimes video is just a background.

sadikyalcin:
> Is there a prop I need to pass in for this fix to work? I am still having this issue.

This is still an issue for me too.

mehthabux:
any update for exoplayer ?

mezalejandro:
Hello,
This still doesn't work in version 6.0.0-beta.5,
any solution?

PR #1119, opened 2018-07-11
Document iOS sidecar text tracks

cobarx:


ashnfb:
@cobarx we have a problem with AVMutableComposition and Airplay -- the combination does not work, and video will not airplay.  This affects textTracks currently.  I spent several hours trying to address the issue. Stack overflow users have reported the issue with no solutions unfortunately.  My interim step, is to try and detect Airplay, and disable textTracks for the player.

cobarx:
@ashnfb Well that's super annoying. If you disable text tracks before switching to AirPlay, does that resolve the problem? We already have a hook to detect when the switch is happening, it's fired as onExternalPlaybackChange, so if that works, you could disable them there.

It would really suck if we have to disable text tracks any time we even see an AirPlay target on the network prior to the switch.

ashnfb:
hi @cobarx.  Tried onExternalPlaybackChange ... it doesn't trigger a call if the video is fullscreen, and then airplay-ed.

PR #1124, opened 2018-07-12
Allow ExoPlayer media to continue to play after being detached from view

shlokamin:
I am currently building an app where I need audio/video to continue playing after I push a new screen onto the navigation stack. This currently works out of the box in iOS, but not in android.

For context of a use case, think about an app like Spotify, where you can navigate away from the media player and continue to control that media being played in the background as you browse different screens.

I added a prop called "doNotDetach" (I'm not sure how I feel about this naming) that allows a consumer to control whether they would like this behavior or not in Exo Player. 

Let me know your thoughts/suggestions!

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -342,6 +342,7 @@ Video.propTypes = {
   playWhenInactive: PropTypes.bool,
   ignoreSilentSwitch: PropTypes.oneOf(['ignore', 'obey']),
   disableFocus: PropTypes.bool,
+  doNotDetach: PropTypes.bool,
   controls: PropTypes.bool,
   audioOnly: PropTypes.bool,
   currentTime: PropTypes.number,
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -120,6 +120,7 @@ class ReactExoplayerView extends FrameLayout implements
     private float mProgressUpdateInterval = 250.0f;
     private boolean playInBackground = false;
     private boolean useTextureView = false;
+    private boolean doNotDetach = false;
     private Map<String, String> requestHeaders;
     // \ End props
 
@@ -193,7 +194,9 @@ class ReactExoplayerView extends FrameLayout implements
     @Override
     protected void onDetachedFromWindow() {
         super.onDetachedFromWindow();
-        stopPlayback();
+        if(!doNotDetach) {
+          stopPlayback();
+        }
     }
 
     // LifecycleEventListener implementation
@@ -842,6 +845,10 @@ class ReactExoplayerView extends FrameLayout implements
 
     public void setDisableFocus(boolean disableFocus) {
         this.disableFocus = disableFocus;
+    }
+
+    public void setDoNotDetach(boolean doNotDetach) {
+        this.doNotDetach = doNotDetach;
     }
 
     public void setFullscreen(boolean fullscreen) {
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
@@ -40,6 +40,7 @@ public class ReactExoplayerViewManager extends ViewGroupManager<ReactExoplayerVi
     private static final String PROP_RATE = "rate";
     private static final String PROP_PLAY_IN_BACKGROUND = "playInBackground";
     private static final String PROP_DISABLE_FOCUS = "disableFocus";
+    private static final String PROP_DO_NOT_DETACH = "doNotDetach";
     private static final String PROP_FULLSCREEN = "fullscreen";
     private static final String PROP_USE_TEXTURE_VIEW = "useTextureView";
 
@@ -185,6 +186,11 @@ public class ReactExoplayerViewManager extends ViewGroupManager<ReactExoplayerVi
     @ReactProp(name = PROP_DISABLE_FOCUS, defaultBoolean = false)
     public void setDisableFocus(final ReactExoplayerView videoView, final boolean disableFocus) {
         videoView.setDisableFocus(disableFocus);
+    }
+
+    @ReactProp(name = PROP_DO_NOT_DETACH, defaultBoolean = false)
+    public void setDoNotDetach(final ReactExoplayerView videoView, final boolean doNotDetach) {
+        videoView.setDoNotDetach(doNotDetach);
     }
 
     @ReactProp(name = PROP_FULLSCREEN, defaultBoolean = false)


cobarx:
I haven't encountered this before, so I'm interested in talking through it. What I saw with react-navigation is that pushing a new route keeps the previous screen mounted so you don't encounter this.  It seems pretty worthwhile though if other routers don't have the same behavior.

It would be great to have consistent behavior between platforms, so I'm wondering what keeps iOS from having the same issue. I'm guessing that unmounting the component and detaching are the same in many cases.

My other concern would be, how do you later re-attach? My apologies if running the code would answer this, I'm a bit short on time today and would need to do a fair bit of setup to test it.

shlokamin:
So we use Wix's react-native-navigation, which behaves like you just said - pushing to a new screen will keep the previous one mounted. For some reason, onDetachedFromWindow() gets called in ReactExoplayerView when we push a new screen onto the navigation stack.

As per the Android docs, onDetachedFromWindow() is called when the main window associated with the activity has been detached from the window manager. 

Given this, the behavior here in Android Exo Player makes some sense to me.

As for why this works out of the box in iOS, I'm not sure. I've been looking into RCTVideo.m, and there are not any lifecycle methods that indicate to me that the player would get detached (as it does in Android).  

I'm not sure what you mean by "how do you later re-attach?". The player will continue to exist in the background. When we want to go back to it, we pop the navigation stack to revisit the screen. 

Let me know if I can clarify anything though. This is definitely an interesting use case for this component. But having the ability to have our player exist in the background is a very powerful feature for us.

johnsiah:
This happened to me when using the tab navigator in Android. 

When we're switching from one tab to another, the current screen will get detached and the video in it will call `stopPlayback()`. But when we switch back to the video tab, all videos in it are unpaused and play simultaneously regardless of states. 

I think this happened is because the `onAttach()` is called when it re-attach as the `initializePlayer()` will be called. The same thing happens on MediaPlayer. I think the `doNotDetach` support for the MediaPlayer is also needed as some user might using it.

cobarx:
Finally got a chance to test this in detail. If you push a new route to a stack navigator, the initial route stays mounted so detach never gets called. However, if you switch to a different tab or unmount the component by replacing the Video with a null element, both cause a detach. Unmounting calls stopPlayback a second time for onDropViewInstance / cleanUpResources. On iOS, switching between tabs (on React Navigation at least) doesn't have any effect. On Android MediaPlayer, it restarts the video from the beginning. 

Wherever we can, we want the platforms to behave the same. I leaning towards getting rid of the detach behavior and letting the developer handle it as part of their routing transitions like they have to do on iOS or with the stack navigator. The drawbacks to this are, we'll be using a bit more memory keeping the video instance mounted and if you are using SurfaceView on ExoPlayer (the default option) it possibles the video will draw on top of everything else, though I'm not seeing that currently.  @n1ru4l and @ashnfb may have some thoughts on this. We would also need to do a major version bump to 4.0.0 since we are changing behavior.

@johnsiah Can you talk more about the situation you encountered with multiple videos. When I tried pausing a video, switching to a different tab, then switching back the video didn't auto-play. For MediaPlayer, I want to get ExoPlayer to a point where people no longer need MediaPlayer. I'm swamped just keeping up with iOS & ExoPlayer, other people will need to put forward patches keeping that going.

shlokamin:
Yeah letting the developer handle routing transitions makes sense to me. I'm not sure how much of a concern memory will be. I can try to do some testing on my end to see what happens when you push multiple videos onto the navigation stack, but I don't imagine it's too intensive.

Since we need the player to continue to exist in the background I'll use my fork until this behavior change is released in the next version update. 

Thanks for taking the time to look into this and your investigative work!

johnsiah:
@cobarx After a few investigations, I found out that this line is causing the auto-play issue after player being released in `onDetach`. 

https://github.com/react-native-community/react-native-video/blob/fd6333f8ac83848b6a0d6295ef5070cdf75b36f5/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L323

`getPlayerWhenReady()` will return true if playing, and false if paused. But this line is saving the opposite value causing the pause flag to be true when previous state is playing, and vice versa.

When the player call `onAttach` after switching back to the tab. The player is initialising with this wrong value. The paused state is being reversed and causing the auto-play issue occured when paused.

I tried to fix to `isPaused = !player.getPlayWhenReady()` and everything now works nicely.

nicer00ster:
Currently experiencing this as well, building a music player with this library and so far it's been awesome! Nice work. When I set playInBackground={true} I can minimize the app and my music will still play, but if I change routes within the application, no matter whether its a Stack or Tab navigator, it'll pause the media, then continue to play whenever I go back to that route. Any solution to this? 😄

shlokamin:
@nicer00ster we ended up having so much trouble implementing a media player with a stack based navigator that we ended up ditching the navigator all together. 

Our media player screen is now part of the top level screen, and we just present/hide via styling as necessary. The logic was actually pretty simple and we have found this approach to make life much simpler for us. 

The concept of a stack based navigator just doesn't really seem to make much sense for a media player that needs to exist in the background. In addition to the problem posed in this thread, when you pop, the video screen (and therefore the player) are gone. This means you have to use a workaround via push, and have somewhat convoluted logic to pass all props that are needed to continue to use the navigator as intended.

We really tried hard to make it work using Wix's react native navigation, but it ended up being more trouble than it was worth. 

If anyone has been able to do this using a navigator, I would love to hear how!

nicer00ster:
@shlokamin I ended up doing the exact same thing. Resolved the issue by moving the Video component outside of the routes within my navigator. While doing this it inspired me to move all my methods over to redux. I stumbled upon one issue though that I can't seem to resolve (It's probably obvious) but with the .seek() method, to my understanding you NEED to have a "ref" on the Video component. Yet I'm unable to figure out how to pass the ref down to a child component that's nested within react-navigation. I understand this issue is unrelated but I figure someone else might happen to come across this as well.

I've tried setting ```ref={ref => this.audio = ref}``` on my video component, then using ```onRef={this.audio}``` on my root navigator. It passes the ref to navigator itself but doesn't send it along to all the routes within the root navigator.

Any insight on how I can achieve this would be wonderful. Thanks! 😄

shlokamin:
@nicer00ster you can control the state of the video player by passing a method into your child component that controls the state.

For example, let's say you wanted to be able to toggle pause/play from a child component. 

In your parent screen:

onProgress = (data) => {
    this.setState({
      currentTime:data.currentTime
    });
  }

togglePausePlay = (paused) =>{
    this.setState({paused})
  }

//not sure the use case for seek, but...

fastForwardFiveSeconds = () => {
 this.player.seek(this.state.currentTime + 5)
}

render(){
 return (
   <View>
     <Video
           ref={(ref) => {
           this.player = ref
           }}
           onProgress={this.onProgress}
           paused={this.state.paused}
        />
   <ChildComponent togglePausePlay={this.togglePausePlay} fastForwardFiveSeconds={this.fastForwardFiveSeconds}/>
 )
}

In your child screen you can then call this.props.togglePausePlay() and the player's state will change. You can do something similar for seek as well.

I'm not sure if this addresses your problem or not, but this is the pattern I use to control the video player from other screens.

nicer00ster:
@shlokamin Mmm well that's the thing, the Video component is now being mounted outside of my screens. It's on the same level where my root router is being initiated. Here's an example. https://github.com/nicer00ster/lofimedia/blob/master/src/Root.js

Sorry it's currently messy, trying to figure out what's going to work with this so I have a lot of commented code lol

cobarx:
Ok, I've gone ahead and turned off stopping playback on detach in #1287. Since iOS doesn't stop playback it doesn't make sense to make this configurable. If people want to pause, they can use the react-navigation hooks.

In my limited testing everything works properly, but let me know if you see anything that breaks.

klodha:
@cobarx Is this still the behaviour (i.e. not stopping on navigating to other screen)? Because in my case when running on Android 7 and 9, both devices stop playback immediately upon navigating to different screen, while I want playback to continue. 

In java code, I can see that stopPlayback is commented inside `ReactExoplayerView.java` file on detach, so not getting how it is still causing video playback to stop.

mrousavy:
@cobarx In my case popping a screen (i.e. detaching it from a window, destroying the context) does not destroy the video (iOS). The video continues to play (I can hear sound) even if the screen is already removed/popped/destroyed. Is this intentional?

I'm using react-native-navigation.

PR #1131, opened 2018-07-16
Document onAudioBecomingNoisy & add iOS support

cobarx:
Provides support for the onAudioBecomingNoisy event which is fired when headphones are disconnected so that developers can choose to pause the player to avoid blasting noise out of the internal speaker.


paddy57:
> Provides support for the onAudioBecomingNoisy event which is fired when headphones are disconnected so that developers can choose to pause the player to avoid blasting noise out of the internal speaker.

Hii @cobarx 
My video lags a frame then plays smooth if i use video which has audio,
but if i used the video which does not have audio then it does not lag. can you help me with that?
I'm using react-native-video@3.0.0
react-native 0.55.4

PR #1145, opened 2018-07-24
little improvement over too many "def"

yeomann:
just added a function which acts like a ternary with fallback option. Hence, less cluttered "def" variables
great to see that lib is supporting already SDK 27

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,20 +1,16 @@
 apply plugin: 'com.android.library'
 
-def _ext = rootProject.ext
-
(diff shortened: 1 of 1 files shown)


cobarx:
Hi @yeomann, that's a nice cleanup. Thanks for submitting! We were fortunate to get a PR to update the SDKs from a guy who submitted this to a bunch of the big React Native projects.

yeomann:
my pleasure @cobarx There are lots of packages with cluttered def, agreed!  I have already sent bunch of pull req along react native package.

PR #1157, opened 2018-07-30
Fix sideload text track issues on iOS

### LINKED ISSUE #1155: TextTrack gets picked up on Android but not on iOS
### Current behavior
The text track from my HLS videos get picked up on Android but not on iOS. The strange part is that before iOS was working fine (nothing changed on my code).

### Reproduction steps
I am looking at the data that's coming back when onLoad triggers, this is how I was able to see the tracks for the same video coming back on Android but not on iOS.

### Expected behavior
TextTracks get picked up both on iOS and Android.

### Platform
* iOS

cobarx:
This PR is designed to fix #1155. 

While investigating this, I discovered the code would cause text tracks to be displayed even if the type was set to "disabled". I've refactored the code so that we only look for text tracks matching the device language if the type is "system".

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -13,6 +13,8 @@ static NSString *const readyForDisplayKeyPath = @"readyForDisplay";
 static NSString *const playbackRate = @"rate";
 static NSString *const timedMetadata = @"timedMetadata";
 
+static int const RCTVideoUnset = -1;
+
 @implementation RCTVideo
 {
   AVPlayer *_player;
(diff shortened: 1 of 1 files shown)


cobarx:
@ashnfb I can't figure out how to make it so all text tracks get disabled. The code to loop through all the tracks and update the enabled status works fine as long as we enable at least one track. But if I setEnabled:NO for all of them, the text for the most recently selected track gets stuck on the screen and never disappears. Any ideas how we can tackle this? I've looked all over Stack Overflow and can't find anything.

Here's the relevant code:
```
  for (int i = firstTextIndex; i < _player.currentItem.tracks.count; ++i) {
    BOOL isEnabled = NO;
    if (selectedTrackIndex != RCTVideoUnset) {
      isEnabled = i == selectedTrackIndex + firstTextIndex;
    }
    [_player.currentItem.tracks[i] setEnabled:isEnabled];
  }
```

cobarx:
@ashnfb Ok, thanks man. I've asked a friend who's pretty familiar with AVPlayer how to tackle this, but I haven't heard back from him yet.

cvaldezscse:
@cobarx Does someone looking at this issue? i can help you if you tell me where to start from

PR #1160, opened 2018-07-31
Expose DefaultLoadControl parameters Android ExoPlayer

bryanvanwijk:
This makes it possible to adjust the parameters of the DefaultLoadControl in the Android ExoPlayer: minBufferMs, maxBufferMs, bufferForPlaybackMs and bufferForPlaybackAfterRebufferMs.


bryanvanwijk:
@cobarx The props can now be passed to the video instead of the separate config. The reason why I had a separate config before was because the parameters are needed during the initialization of the video player and cannot be changed like the other video properties. So now I just create a new video player when the load control parameters are set.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -931,4 +939,14 @@ public void setFullscreen(boolean fullscreen) {
     public void setUseTextureView(boolean useTextureView) {
         exoPlayerView.setUseTextureView(useTextureView);
     }
+
+    public void setLoadControl(int newMinBufferMs, int newMaxBufferMs, int newBufferForPlaybackMs, int newBufferForPlaybackAfterRebufferMs) {
+        player.release();
```
cobarx: This will generate a null pointer exception if setLoadControl gets called before setSrc. There is no guarantee to the order, it's somewhat random depending on what order you specify props in. I'm not sure if it's a good idea to call `player.release()` without doing all the cleanup that happens in `releasePlayer()` so I would suggest calling `stopPlayback()` which will do the null pointer check and cleanup.

Also, you can remove the `this` since all the variable names are different it doesn't achieve anything.

cobarx:
Thanks, yeah I was wondering how you were going to handle that, good catch. I left you a note about the implementation.

Sorry to ask for another round of changes, but I think it would be better to call the prop bufferConfig or bufferSettings. If someone is looking at the list of props, something that mentions buffer will be a lot more obvious than having to read through everything and then discover that you can change buffer settings via loadControl. I would also mention in the docs that you need to apply the prop when loading the video, since AFAIK you can't adjust it after the video is already loaded.

cobarx:
Thanks for making the updates! I tested this, made a few tweaks and everything is ready to go. Thanks again for implementing.

Issue #1143, opened 2018-07-23
iOS Sidecar VTT TextTrack not working for longer videos (~1hr)


shlokamin:
### Current behavior
When loading in sidecar text track for long videos (~1hr) in iOS, the video will not start. When pressing play, the player will simply just stay idle. 

I have had no issues like this with shorter videos (~10-20 mins)

### Reproduction steps

Using this video: https://publish.dvlabs.com/democracynow/360/dn2018-0720.mp4
Using this VTT: https://integration.democracynow.org/resources/captions/shows/5874/English.vtt

i.e.

source={{uri:https://publish.dvlabs.com/democracynow/360/dn2018-0720.mp4}}
selectedTextTrack={{
                type: "title",
                value: "English"
              }}
textTracks={[{
                  title: "English",
                  language: "en",
                  type: TextTrackType.VTT,
                  uri: https://integration.democracynow.org/resources/captions/shows/5874/English.vtt
                }]}

### Platform
Which player are you experiencing the problem on:
* iOS

cobarx:
@ashnfb Can you take a look into this? I will also see if I can figure anything out.

ashnfb:
@cobarx I'll take a look and see if I can figure it out.  I have a film not playing too, also when captions are enabled via the system.

cobarx:
Ok I did some tests and it appears that AVPlayer doesn't like the server that's hosting the subtitles. I downloaded them to my server and they work fine with no other modifications:
http://hamptonmaxwell.com:8080/DemNow.vtt

The video works fine whether I don't have subtitles, use my server, or use other subtitles. If I use integration.democracynow.org, onError shows:
```
{ code: -11800, domain: "AVFoundationErrorDomain" }
```
and the error in the native code is:
```
Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSLocalizedFailureReason=An unknown error occurred (-12783), NSLocalizedDescription=The operation could not be completed, NSUnderlyingError=0x600000249a80 {Error Domain=NSOSStatusErrorDomain Code=-12783 "(null)"}}
```

Looking at AVError.h, -11800 AVErrorUnknown. I've also seen that error when trying to load unsupported subtitle formats on iOS like SRT, so it's not very helpful.

I'd suggest reaching out to the Democracy Now team and seeing if they have had anyone else encounter this issue.

ashnfb:
[...]

PR #1162, opened 2018-08-01
Fix for #1150 and improvements for #1143

### LINKED ISSUE #1150: iOS Video not displaying in release mode from bundled assets
### Current behavior
Video does not display in 'Release' Build Configuration' in iOS

### Reproduction steps

```
<Video style={styles.landingVideo} muted repeat resizeMode="cover" source={require('../../assets/videos/bg.mp4')} />
```

Build in iOS a choose 'Release' build configuration by changing the scheme cmd + shift + <

### Expected behavior
Video to display

### Platform
* iOS

Package version 3.1.0
react-native 0.56.0

ashnfb:
@cobarx @ahmetabdi
Fixes #1150 
Improves #1143 (onLoad -> textTracks returns only valid sideloaded vtt)

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -308,26 +308,26 @@ static int const RCTVideoUnset = -1;
   [self removePlayerItemObservers];
   
   dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
-    
+
     // perform on next run loop, otherwise other passed react-props may not be set
     _playerItem = [self playerItemForSource:source];
     [self addPlayerItemObservers];
-    
+
     [_player pause];
     [_playerViewController.view removeFromSuperview];
     _playerViewController = nil;
-    
+
     if (_playbackRateObserverRegistered) {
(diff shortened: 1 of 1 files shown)


cobarx:
@ashnfb Thanks for all your contributions!

Did you by any chance figure out how to disable sideloaded tracks? I'm still waiting to hear back from my contact.

ashnfb:
@cobarx I gave disabling tracks a try.  I'm a little frustrated with the lack of AVPlayer docs.  I'll try and spend a bit more time on it.

cobarx:
@ashnfb Yup, AVPlayer is great when it works, when it doesn't you're just out of luck.

I was thinking about this last night and I don't think we want the fallback code if it can't find the requested track. The way that code works right now is it will only use the fallback if the system Captions preference is enabled, so the behavior will be inconsistent. I think it's totally reasonable to say that if you request an invalid track nothing happens. Really it should throw an error, but there's no framework for it (onError has already been documented as only being for load errors). Also, we don't use the fallback code on Android, so the behaviors are different. Do you have an objection to me reverting that part of the code?

ashnfb:
@cobarx no objection to reverting that code!

PR #1165, opened 2018-08-03
Fixing fullscreen mode for tvOS

cdeust:
I'm made modification in the way the player is created when opening for fullscreen mode.
I've created a new method to fetch the AVURLAsset using the var _player. 
I'm then assigning the AVURLAsset video to a new AVPlayer instanciated when going fullscreen and then passing the video player to RCTVideoPlayerViewController.

Working fine on tvOS with video around 1h30 min

### DIFF
--- a/ios/RCTVideo.m
+++ b/ios/RCTVideo.m
@@ -15,6 +15,9 @@ static NSString *const timedMetadata = @"timedMetadata";
 
 static int const RCTVideoUnset = -1;
 
+static NSNumber *currentTime;
+static NSNumber *timescale;
+
 @implementation RCTVideo
 {
   AVPlayer *_player;
@@ -149,7 +152,13 @@ static int const RCTVideoUnset = -1;
   __weak RCTVideo *weakSelf = self;
   _timeObserver = [_player addPeriodicTimeObserverForInterval:CMTimeMakeWithSeconds(progressUpdateIntervalMS, NSEC_PER_SEC)
                                                         queue:NULL
-                                                   usingBlock:^(CMTime time) { [weakSelf sendProgressUpdate]; }
+                                                   usingBlock:^(CMTime time) {
+                                                       if (! [weakSelf getFullscreen] && !_playerViewController) {
+                                                           currentTime = [NSNumber numberWithFloat:(Float64)_playerItem.currentTime.value/(int32_t)_playerItem.currentTime.timescale];
+                                                           timescale = [NSNumber numberWithFloat:(int32_t)_playerItem.currentTime.timescale];
+                                                           [weakSelf sendProgressUpdate];
+                                                       }
+                                                   }
                    ];
 }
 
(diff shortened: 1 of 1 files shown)


cobarx:
I appreciate you looking into this. I tested this a bit and don't think this is going to be an optimal approach. It will work ok if you start out in fullscreen, but in many cases people are going to want to switch between non-fullscreen & fullscreen mode. By creating a new player instance, the video has to be reloaded which adds a significant delay on some videos I tested.

If we do go down this route, there needs to be added code when switching to pause the initial player instance, grab the current position, and apply all the modifiers like volume & rate to the new player. We also need to apply those modifiers to the new player instance whenever they change.

What I'd like to do next is see if the fullscreen code works in the Expo Video component and if it does, figure out how to use their code here. There are also some examples on Stack Overflow that might work like:
https://stackoverflow.com/questions/30817667/how-to-make-a-avplayerviewcontroller-go-to-fullscreen-programmatically/33785005
If neither of those works, we can add the extra code as described above.

cdeust:
Actually the problem that has been present is coming from the source that is passed to the player. So I’ll keep looking in the same direction and provide another version with a smooth transition between non full screen player and full screen player (I mean without loosing the progress of video and audio).

I don't think trying implementing selector that are part of apple private api as a fix is the best solution, either the users will get their app refused, or they will get crashs when iOS 12 gets out.

cdeust:
@cobarx / @JulienKode Could you please review the modification I've made.

You can now continue to play in fullscreen a video you started with the embedded player, and audio track is normally being recalculated to match the video marker position.

JulienKode:
The coding style looks good to me, 
I didn’t test it yet, but it looks like a good solution 

I agree with @cdeust using private API is not recommanded

inlinecoder:
I can confirm that videos don't stuck now.

But, I've noticed one thing — application crashes once you want to `presentFullscreenPlayer` before `onReadyForDisplay` is invoked. 

Before this PR, I could invoke `presentFullscreenPlayer ` even in `onLoad` event.

```
2018-09-16 02:21:58.473678+0300 ASD-tvOS[22819:1511904] [aqme] 755: MEMixerChannel::EnableProcessor: failed to open processor type 0x705f6571
2018-09-16 02:21:58.887432+0300 ASD-tvOS[22819:1511776] Unbalanced calls to begin/end appearance transitions for <RCTVideoPlayerViewController: 0x7fd98b003e00>.
2018-09-16 02:21:58.906909+0300 ASD-tvOS[22819:1511776] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[AVPlayerItem seekToTime:toleranceBefore:toleranceAfter:completionHandler:] Seeking is not possible to time {INVALID}'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010e782a56 __exceptionPreprocess + 294
	1   libobjc.A.dylib                     0x000000010d0f7031 objc_exception_throw + 48
	2   AVFoundation                        0x000000010da06bbb -[AVPlayerItem seekToTime:toleranceBefore:toleranceAfter:completionHandler:] + 1961
	3   AVFoundation                        0x000000010da06409 -[AVPlayerItem seekToTime:completionHandler:] + 146
	4   AVFoundation                        0x000000010da06304 -[AVPlayerItem seekToTime:] + 49
	5   ASD-tvOS                            0x000000010c3a2278 __26-[RCTVideo setFullscreen:]_block_invoke + 808
	6   UIKit                               0x0000000110f1af4a -[UIPresentationController transitionDidFinish:] + 1346
	7   UIKit                               0x000000011114da98 -[_UICurrentContextPresentationController transitionDidFinish:] + 44
	8   UIKit                               0x0000000110f1edba __56-[UIPresentationController runTransitionForCurrentState]_block_invoke.430 + 183
	9   UIKit                               0x00000001119caeda -[_UIViewControllerTransitionContext completeTransition:] + 102
	10  UIKit                               0x0000000110f17d36 -[UITransitionView notifyDidCompleteTransition:] + 251
	11  UIKit                               0x0000000110f179ad -[UITransitionView _didCompleteTransition:] + 1397
	12  UIKit                               0x0000000110f19fe4 -[UITransitionView _transitionDidStop:finished:] + 104
	13  UIKit                               0x0000000110e4d499 -[UIViewAnimationState sendDelegateAnimationDidStop:finished:] + 343
	14  UIKit                               0x0000000110e4dadc -[UIViewAnimationState animationDidStop:finished:] + 293
	15  UIKit                               0x0000000110e4db90 -[UIViewAnimationState animationDidStop:finished:] + 473
	16  QuartzCore                          0x00000001107c1d07 _ZN2CA5Layer23run_animation_callbacksEPv + 323
	17  libdispatch.dylib                   0x0000000110c5f848 _dispatch_client_callout + 8
	18  libdispatch.dylib                   0x0000000110c6a92b _dispatch_main_queue_callback_4CF + 628
	19  CoreFoundation                      0x000000010e745509 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
	20  CoreFoundation                      0x000000010e709716 __CFRunLoopRun + 2342
	21  CoreFoundation                      0x000000010e708b7b CFRunLoopRunSpecific + 635
	22  GraphicsServices                    0x00000001152a8a73 GSEventRunModal + 62
	23  UIKit                               0x0000000110dbd214 UIApplicationMain + 159
	24  ASD-tvOS                            0x000000010c0d5fd0 main + 112
	25  libdyld.dylib                       0x0000000112c4bd99 start + 1
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

PR #1167, opened 2018-08-04
Remove flash between poster and video

### LINKED ISSUE #1128: (Android) Black Screen for a second between loading and video playback
There is an always around 2 seconds of black screen when moving from onLoad which stop my loading spinner component and the start of the video playback in my code below. I am using ExoPlayer.
Any idea what i am doing wrong, or how can i improve my code to get rid of the black screen?!

My Code:
```
render() {
    let loading = null;
    let waitingVideo = null;
    if (!this.state.waitingVideoURL) {
      loading = (
        <View>
          <Text style={styles.loading}>Loading...</Text>
          <ActivityIndicator size="large" color="#2C3942" />
        </View>
      );
    } else {
      waitingVideo = (
        <Video
          source={{
            uri: this.state.waitingVideoURL,
          }}
          resizeMode="cover"
          style={styles.backgroundVideo}
          repeat
          onLoad={this.stopLoading}
        />
      );
    }

    return (
        <StatusBar hidden />
        {waitingVideo}
        {loading}
      </View>
    );
  }
}
```

cmmartin:
This resolves https://github.com/react-native-community/react-native-video/issues/1128

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -248,35 +248,21 @@ export default class Video extends Component {
       onAudioBecomingNoisy: this._onAudioBecomingNoisy,
     });
 
-    if (this.props.poster && this.state.showPoster) {
-      const posterStyle = {
-        position: 'absolute',
-        left: 0,
-        top: 0,
-        right: 0,
-        bottom: 0,
-        resizeMode: this.props.posterResizeMode || 'contain'
-      };
-
-      return (
-        <View style={nativeProps.style}>
-          <RCTVideo
-            ref={this._assignRoot}
-            {...nativeProps}
-          />
-          <Image
-            style={posterStyle}
-            source={{uri: this.props.poster}}
-          />
-        </View>
-      );
-    }
+    const posterStyle = {
+      ...StyleSheet.absoluteFillObject,
+      resizeMode: this.props.posterResizeMode || 'contain',
+    };
 
     return (
-      <RCTVideo
-        ref={this._assignRoot}
-        {...nativeProps}
-      />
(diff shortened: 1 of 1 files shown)


n1ru4l:
Could you please explain why this prevents the flashing?

cmmartin:
If you put a breakpoint in the `dealloc` method of `RTCVideo.m`, it will be fired once the video loads and the poster is removed. So, the current implementation allocates a native video player, renders it behind a poster, then once the video loads, it de-allocates the original instance and creates a new instance. That re-instantiation causes the flash. After applying my changes, `dealloc` is not called when the video loads and the flash is gone.

I don't know enough about the internals of React to say for sure why it is not re-using the native video player upon video load, but I would guess the fact that the `RCTVideo` component changes parents.

In other words, the switch from...

```
 <View style={nativeProps.style}>	
        <RCTVideo	
            ref={this._assignRoot}	
            {...nativeProps}	
        />
       ...
</View>
```

to simply...

```
<RCTVideo	
   ref={this._assignRoot}	
   {...nativeProps}	
/>
```

leads to a creation of a new `RCTVideo` instance

n1ru4l (approved):
@cobarx Lets merge this!

cobarx:
@n1ru4l Ok, feel free to approve any of these open PRs and I'll merge them. I've been short on time to do the extensive testing to make sure that many of these work.

Before approve anything, please review and make sure the code is readable, test it, and make sure any new functionality is documented.

Issue #1166, opened 2018-08-03
[Android] Update Exoplayer to 2.8.X


Chubacca:
Can we update Exoplayer to use 2.8.X? We currently have a hybrid native/react-native app and would like to use the newest Exoplayer, and the current react-native-video code is not compatible with 2.8.X.

Specifically 2.8.X removed the method "clearSelectionOverrides" in [this commit](https://github.com/google/ExoPlayer/commit/67cde97a70f60e9de6bd7fe9804fe48a7366c16f#diff-03150d30292d64a476a92dc35441f0e1). However, react-native-video calls the method [here](https://github.com/react-native-community/react-native-video/blob/91194edbc9a6352326a8bf00d390cf58419ff99c/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L819). There will potentially be other incompatibilities, but this was the exception we ran into first.

### Platform
Which player are you experiencing the problem on:
* Android ExoPlayer

cobarx:
I'll do this pretty soon and handle the API changes. There were some outstanding PRs I wanted to get merged first to avoid conflicts, but I think those are in now.

cobarx:
Done in #1170. I've handled all text track stuff. I didn't hit any issues with other features not working post upgrade, but testing would be welcome in advance of the release. You can grab the update off the master branch.

PR #1170, opened 2018-08-06
Update ExoPlayer to version 2.8.2

cobarx:
Updates us to the latest ExoPlayer version. This version requires a newer Android SDK version to compile, so it will be a BC (breaking change) release.


cobarx:
Fixes #1166

PR #1172, opened 2018-08-08
Prevent errors on iOS when using text tracks with caching

cobarx:
This PR is intended to fix a crash that happens when you attempt to mix in text tracks. Also fixes issues where the right error code wasn't returned when the file couldn't be cached.

@n1ru4l I did some refactors to make it a bit easier to read. IMO, putting all of the caching code into a separate function so you don't have to parse all the ifdefs is a lot easier to digest. For future PRs, I'd suggest breaking it into pieces. Giant PRs are make it more challenging to figure out what changed and also are a bit harder to merge since everything needs to be tested and evaluated all at once.

I did some investigation into the text track issue and have it mostly isolated. In the text track code https://github.com/react-native-community/react-native-video/blob/86d655c3d1c8a81f6b841d7379ba624903440b32/ios/Video/RCTVideo.m#L393 it is bombing at the line for:

`AVAssetTrack *videoAsset = [asset tracksWithMediaType:AVMediaTypeVideo].firstObject;`

Which throws an error in the debugger of:
```
error: warning: got name from symbols: AVMediaTypeVideo
error: cannot initialize a parameter of type 'id' with an lvalue of type 'void *'
```

I was thinking maybe DVURLAsset didn't subclass AVURLAsset and was missing the `tracksWithMediaType` method and manually created the AVURLAsset. However, that didn't fix it and it turns out it does subclass it. 

I ran out of time today to finish investigating but will see if I can figure out that error this evening or tomorrow.

@n1ru4l A few other questions for you. Can we also cache the audio & text tracks? Also, I'm assuming that if you watch a video for a second time and part of the video is in the cache, it will pull any other parts that aren't cached from the network, right? Just wanted to be sure.

### DIFF
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -9,11 +9,13 @@ The cache is backed by [SPTPersistentCache](https://github.com/spotify/SPTPersis
 # How Does It Work
 
 The caching is based on the url of the asset.
-SPTPersistentCache is a LRU ([last recently used](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU))) cache.
+SPTPersistentCache is a LRU ([Least Recently Used](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU))) cache.
 
 # Restrictions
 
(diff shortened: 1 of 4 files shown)


n1ru4l:
@cobarx I will try to take a look when I got some spare time, currently having text tracks is unfortunately not a requirement of our app 😕.

cobarx:
That’s fine, worst case scenario we disable caching for text tracks and
come back to it later.

cobarx:
I have no idea how to fix this so I'm going to disable caching if text tracks are specified and get this released. Glad we're finally getting to ship this!

Hopefully some day your company will need to subtitles and we can get them fixed :)

PR #1183, opened 2018-08-12
Use FileDataSource for file:/// Uris for better handling of in-progress I/O (eg. playback during download)

akinwale:
This allows an incomplete media file loaded from the local filesystem to continuously play while it is being written to (in our use case, playback of a file while it is being downloaded). https://github.com/lbryio/lbry-android/issues/191 provides some additional context for why this was done.


cobarx:
@akinwale Great feature! So if I wanted to test this, I'd need to setup a download task and play the file at the same time. And the file should be in the form file:///storage/Downloads/myfile.mp4 or something like that. Correct?

akinwale:
Yes, this is correct.

cobarx:
Do you have some code that you could send over for performing download? I don't have anything like that put together, so I'd have to research it. I should be able to merge this quite quickly once I test it.

PR #1219, opened 2018-09-06
Support RN 0.57 w/ gradle 3

fiznool:
React Native 0.57 will [scaffold new projects using Gradle 3, instead of Gradle 2](https://github.com/facebook/react-native/commit/6eac2d4e2f5f697043b495002872bfac3e8595cb).

Users who install this library with Gradle 3 will see the following error:

```
Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html
```

This PR fixes this warning.

Please note: this will cause any users who are still on Gradle 2 to see an error:

```
> Could not find method implementation() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

This is because `implementation` is only available for users of Gradle 3.

It may therefore be worth bumping the major version of this lib to indicate a breaking change for some users.


cobarx:
Thanks for suggesting & implementing this!

I had investigated this before and Gradle 2 didn't like the new compileOnly and implementation directives. I'm kinda nervous to force everyone to move to RN 0.57 ASAP or miss out on react-native-video fixes. Is there any way to get this to be backward compatible? If not, I'd like to wait a few months until most apps have updated their RN.

fiznool:
I’ve yet to find a way to make this backwards compatible, but I’m no expert
in grade!

I understand the concerns. FWIW I think the RN community generally expects
that to get the latest features and fixes in a third party module, the
latest version of RN needs to be used.

cobarx:
If we leave things the way they are, does everything work properly on RN 0.57? If so, I think we should leave it that way for a month or two. There's no point in forcing people to upgrade just to avoid warnings.

In my experience, .0 RN releases are often unstable, so it's a bad idea to upgrade before there have been a few weeks to sort the bugs out.

fiznool:
Yes - things will continue to work, but users on RN 0.57 will see a warning
from Gradle.

According to the warning, this change will need to be in place by the end
of 2018.

n1ru4l (changes requested):
I quickly scanned the Gradle API Docs. Apparently it seems to possible to include two `build.gradle` files, where one has a different name like `build-new.gradle`.

https://docs.gradle.org/current/javadoc/org/gradle/api/initialization/ProjectDescriptor.html#setBuildFileName-java.lang.String-

I quickle tested it by copying the `build.gradle` file and adding throw Statement into the original `build.gradle`. It was possible to override the path like this:

```gradle
project(':react-native-video').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-video/android-exoplayer')
project(':react-native-video').setBuildFileName('build-1.gradle')
```

I would recommend doing it that way, until 0.57 is "stable".

cobarx:
I'm going to push out a 3.3.0 release that adds all the latest fixes but doesn't require the Gradle 3 plugin. That gives people a good solid version they can use if they can't update to RN 0.57 right away.

In order to upgrade to ExoPlayer 2.9.0, it turns out you need to have Gradle 3 installed. So we can do this change and the ExoPlayer upgrade for 4.0.

Issue #1234, opened 2018-09-13
Error setting property 'fullscreen' of RCTVideo


svegan:
### Current behavior
After updating from 0.47 react version and plugin version 1.2.0 can't launch video in full screen.
When executing presentFullscreenPlayer() function I'm getting an error:
```Exception thrown while executing UI block: Application tried to present a nil modal view controller on target <UIViewController: 0x7ffe3c305600>.``` Screenshot attached.
<img width="257" alt="screenshot 2018-09-13 16 23 43" src="https://user-images.githubusercontent.com/13061553/45493089-a5c56080-b776-11e8-9719-a0dfdfd01be7.png">

At this time xcode console writes next:
"[error][tid:main][RCTUIManager.m:1107] Error setting property 'fullscreen' of RCTVideo with tag #619: Exception thrown while executing UI block: Application tried to present a nil modal view controller on target <UIViewController: 0x7fc94e541e50>."

### Reproduction steps
Simply try to open a video in full screen on iOS.

### Expected behavior
Opens the video in full screen.

### Platform
Which player are you experiencing the problem on:
* iOS

fossage:
I am experiencing the same thing.  It appears to be an issue on the Objective-C side where the player can potentially be uninitialized at the point that `presentFullScreenPlayer` is called.  This is true in the case where you call it right after getting the ref to the video component, or if you try to use the old api and pass `fullscreen` as a prop.

The issue is the player is being instantiated asynchronously, so it may be nil at the point we are trying to fullscreen it.  As a kind of shitty hack to get around this in the meantime, you can wrap your call to `presentFullscreenPlayer()` in a `setTimeout` with a value of something like 100.  It pretty gross, but might carry you over until this gets fixed.  I'll try to get in a PR for it at some point, but my Objective-C skills are somewhat mediocre so it might take me a little bit.

svegan:
Thanks a lot @fossage for response. That's what I have accidentally discovered yesterday as well, but had not time to write down. In my case it worked even with 0 delay, so I made assumption - it's just necessary to add in the end of event loop queue. So went a bit futher and start using [requestAnimationFrame](https://facebook.github.io/react-native/docs/timers) which works pretty well for me as well. But that's of course just workaround and Objective-c solution should be found.

PR #1244, opened 2018-09-18
Fixes bug that causes player to crash when fullscreening at initialization (#1234)

fossage:
## Overview

Because the `setSrc` method now runs async, it is possible for the player to not be initialized at the point a user calls the `setFullscreen` method via `presentFullscreenPlayer`. A good example of this is when a user of the library wants to fullscreen a video as soon as the ref for the video component is captured. 

To prevent this problem, in the case where the user is trying to fullscreen the video but the player has not yet been initialized, this will set a flag letting the system know that the user would like to fullscreen the video as soon as the player is initialized. Upon initialization, that flag will be read, the player will be fullscreened, and the flag will be set back.

See (#1234)

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -32,6 +32,7 @@ static int const RCTVideoUnset = -1;
   BOOL _playerLayerObserverSet;
   RCTVideoPlayerViewController *_playerViewController;
   NSURL *_videoURL;
+  BOOL _pendingFullScreenRequest;
   
   /* Required to publish events */
   RCTEventDispatcher *_eventDispatcher;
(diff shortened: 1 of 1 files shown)


dozel:
Is this going to be merged at any point? Would love this fix

n1ru4l (approved):

n1ru4l:
@fossage could you please rebase on master?

fossage:
@n1ru4l , branch has been updated.

kcalmes:
When will this pull request be accepted?

nmalancea:
Any plans to get this merged? `fullscreen` property is not working for me.
Getting:
```
Exception thrown while executing UI block: Application tried to present modally a view controller <RCTVideoPlayerViewController> that has a parent view controller <UIViewController>
```

PR #1246, opened 2018-09-19
Instead of crashing, show a warning and return

niklassaers:
I was having an issue with iOS crashing in RCTVideo.m line 446 because sometimes my data wasn't as I thought it should be. pathForResource is nullable, yet initFileURLWithPath cannot take a null-parameter. This patch makes sure we don't get a null at all on bad input.


niklassaers:
(and provides a warning about it)

cobarx:
It's ok to have an empty type. I think we can just check the uri and be good. Haven't had a chance to test it though.

cobarx (approved):

PR #1253, opened 2018-09-27
Support react-native-dom

cobarx:
This adds support for using react-native-video using the [react-native-dom](https://github.com/vincentriemer/react-native-dom) platform. react-native-dom is a complete implementation of React Native for the web.

Most props are implemented & working.

Remaining todo items:
* Add event handlers
* Document supported props
* Test on browsers other than Chrome
* If possible, make controls hide & appear by clicking on the video
* Add fullscreen support

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,31 @@ using System.Collections.Generic;
 ```
 </details>
 
+<details>
+  <summary>DOM</summary>
+
+Make the following additions to the given files manually:
+
+**dom/bootstrap.js**
+
+Import RCTVideoManager and add it to the list of nativeModules:
+
+```javascript
+import { RNDomInstance } from "react-native-dom";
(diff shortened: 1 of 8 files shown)


cobarx:
Ok, all the common props and events are there so I'm merging this! 

The big missing piece is fullscreen support which is tricky because react-native-dom only has an async event handler model. To request fullscreen in a browser you must do it in a synchronous response to a user gesture like a key press or mouse click. I'm hoping I can get the react-native-dom maintainer to create a method to work around this since fullscreen is critical for web apps.

cobarx:
This is available on master right now. I have a number of PRs to review and merge for 4.0, probably end of next week.

Be aware that this is a first pass, so some stuff may be buggy - like I need to handle browsers that don't allow autoplay.

PR #1283, opened 2018-10-12
Make seek() throw an exception when using NaN value

laurent22:
When `seek()` is used with a `NaN` value, a very hard to debug error is generated somewhere deep in React Native, which makes it hard to find the cause of the error (especially when the seek value is the result of a calculation that goes wrong in some cases but not others):

image

This pull request improves this by throwing an error early, which makes the issue clearer and provides a stack trace in user space:

image

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -51,6 +51,8 @@ export default class Video extends Component {
   }
 
   seek = (time, tolerance = 100) => {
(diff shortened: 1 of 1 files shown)


cobarx:
Ok this makes a lot of sense, going ahead and merging this is a big improvement over the arcane error. Thanks.

PR #1302, opened 2018-10-24
Fix ExoPlayer: avoid stretching of HLS when aspect ratio changes

brianephraim:
The ExoPlayer implementation was not properly handling anamorphic HLS playlists, i.e. videos that change size or aspect ratio during playback.  iOS already handled this complexity fine.

Here is an HLS that exhibits this anamorphic behavior:  
https://hls.ted.com/talks/93.m3u8

## before fix
https://hls.ted.com/talks/93.m3u8
stretched-video

## after fix
fixed-stretch-video

Thanks for opening a PR!

#### Update the documentation

#### Update the changelog

#### Describe the changes

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -205,9 +205,7 @@ public final class ExoPlayerView extends FrameLayout {
             layout.setAspectRatio(height == 0 ? 1 : (width * pixelWidthHeightRatio) / height);
 
             // React native workaround for measuring and layout on initial load.
-            if (isInitialRatio) {
-                post(measureAndLayout);
-            }
+            post(measureAndLayout);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


nbennink:
Thanks it works for me. Don't forget to remove the (now useless) comment and variable in the code.

NrqhEIcsRi004:
Good PR :)

NrqhEIcsRi004:
Can a ratio change be used for a callback?

egor-bukharov:
Hi! The PR is pending for 2 years, does this fix work now?

yegorius:
We have recently hit this bug in production and it affects a lot of our customers. Can someone please merge it?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -201,13 +201,8 @@ public void onCues(List<Cue> cues) {
 
         @Override
         public void onVideoSizeChanged(int width, int height, int unappliedRotationDegrees, float pixelWidthHeightRatio) {
-            boolean isInitialRatio = layout.getAspectRatio() == 0;
             layout.setAspectRatio(height == 0 ? 1 : (width * pixelWidthHeightRatio) / height);
-
-            // React native workaround for measuring and layout on initial load.
-            if (isInitialRatio) {
```
freeboub, line 208: With this patch you will have strange behavior.
If you are playing content with ABR, onVideoSizeChanged will be called on every time playlist change.
you need to check the aspect ratio have a real change,
something like this:
abs(oldAspectRatio - newAspectRatio) > aspectRatioChangeThreesold

PR #1304, opened 2018-10-25
disable airplay when sideloaded captions

ashnfb:
For sideloaded captions on iOS, we have to use an AVMutableComposition.  Unfortunately AVMutableComposition is not playable on Airplay.  This code turns off video playback (sets external playback to false).

Additional check on textTracks also returns the default AVPlayerItem if textTracks is not null, but is empty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -603,6 +603,8 @@ uri | URL for the text track. Currently, only tracks hosted on a webserver are s
 
 On iOS, sidecar text tracks are only supported for individual files, not HLS playlists. For HLS, you should include the text tracks as part of the playlist.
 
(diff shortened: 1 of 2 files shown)


cobarx:
Looks good. I would add a note in the textTracks docs covering this limitation so developers know to expect it.

ashnfb:
@cobarx note for textTracks issue with airplay added to readme.  PR ready for merge.

cobarx:
Thanks for making the change, I made a few slight tweaks to the wording to make it a bit more clear. Glad you got this fixed!

PR #1310, opened 2018-10-29
Support for maximum bit rate adaptive streaming

ashnfb:
iOS and Android support maximum bit rates for HLS streams; and this can be helpful when a user wants to limit bandwidth consumption (example may include limiting content on cellular devices).

This PR implements this prop for both platforms

#### Example to test:
specify adaptive streaming without maximum bitrate limit (this is default when prop not specified)
maximumBitRate={0}

specify adaptive streaming with maximum bitrate limit
maximumBitRate={2500000}

[2500000 is an example medium-quality encoded 720P h264 video, but it will vary by the encoder and source]


cobarx:
Hi @ashnfb, great feature.

I went ahead and made some updates to the documentation and code. I've renamed it to maxBitRate to keep it in the style of bufferConfig.

In the docs, I made a couple of changes:
* Added a link to the maxBitRate entry in the list of props
* Added platform support info
* Removed note that this is for HLS streams. On ExoPlayer, this will work on DASH streams with multiple video profiles as well
It's critical we document these correctly, as someone reading the docs will be looking to them to understand the capabilities of the feature.

Looking at the iOS code, if you set the maxBitRate after the video is loaded, nothing will happen because there is nothing in applyModifiers to apply it. This is a pretty common use case if you want to emulate YouTube and change the bitrate mid-playback. Please update the PR to handle it.

Otherwise looking very good. I am going to wait for the change and will test then. Do you by any chance have a sample stream I can use to test?

c1ngular:
it is possible to make video player to automatically switch between different bitrates/resolutions according to current internet connection speed/quality at real time ?

ashnfb:
@c1ngular, the code permits chaining the maximum bitrates; however two caveats:

1. on iOS, it's not clear whether Apple supports changing bandwidth during playback
2. I don't think is the responsibility of react-native-video to make decisions about internet connection and accordingly change bitrate.  In my project, we do that in our own codebase.

c1ngular:
@ashnfb thanks !

cobarx:
@ashnfb Tested this on iOS tonight and it's looking good. I'll get to ExoPlayer tomorrow and merge it! Thanks for making the changes.

cobarx:
@ashnfb It just occurred to me that we should document how to clear the value by setting it to zero. If you get to it before I review tomorrow, can you add a note about that.

Also, looking at the Android code you are setting max rate to 0 when initializing the video. That's not what we want since it wants MaxInfinity to indicate auto-selection, I'd imagine it will cause it to select the minimum bitrate stream. I would instead use a conditional like this:
```
            trackSelector.setParameters(trackSelector.buildUponParameters()
                            .setMaxVideoBitrate(maxBitRate === 0 ? Integer.MAX_VALUE : maxBitRate));
```

ashnfb:
@cobarx I'll make the change for Android in the next hour!

ashnfb:
@cobarx thanks for making the changes

cobarx:
@ashnfb No problem, I didn't see your message till just now. Have a little time this morning to do open source work, so I hopped right into it.

ivanzhang:
@cobarx  i getting build error in android ...
image

coucoseth:
> @c1ngular The underlying player will automatically change bitrates to what it feels is the appropriate stream/resolution. I believe there is a PR in the works to support choosing a specific resolution similar to what YouTube offers.

Was this feature worked on, if yes, how can i activate/implement

PR #1325, opened 2018-11-12
Implement picture in picture for iOS

aalzanki:
Test Plan:
 - Run on ipad
 - test out `onIsPictureInPictureSupported`, `onIsPictureInPictureActive`, `restoreUserInterfaceForPictureInPictureStop`, `startPictureInPicture`, `stopPictureInPicture`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -867,7 +867,7 @@ Payload: none
 Platforms: Android ExoPlayer, Android MediaPlayer, iOS
 
 #### onIsPictureInPictureActive
-Callback function that is called when picture in picture becames active on inactive.
+Callback function that is called when picture in picture becomes active or inactive.
 
 Property | Type | Description
 --- | --- | ---
--- a/README.md
+++ b/README.md
@@ -298,6 +298,8 @@ var styles = StyleSheet.create({
 * [onFullscreenPlayerDidPresent](#onfullscreenplayerdidpresent)
 * [onFullscreenPlayerWillDismiss](#onfullscreenplayerwilldismiss)
 * [onFullscreenPlayerDidDismiss](#onfullscreenplayerdiddismiss)
+* [onIsPictureInPictureActive](#onispictureinpictureactive)
+* [onIsPictureInPictureSupported](#onispictureinpicturesupported)
 * [onLoad](#onload)
 * [onLoadStart](#onloadstart)
 * [onProgress](#onprogress)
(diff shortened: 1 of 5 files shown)


cobarx:
This is a very cool feature, thanks for submitting and providing all the documentation!

I'd like to see if we can condense the interface so that it's as simple as possible.

onIsPictureInPictureSupported - React Native only supports iOS >= 9.0 so I don't think we need this.
onIsPictureInPictureActive - This gets called whenever the state changes, so it should be onPictureInPictureChanged or onPictureInPictureStatusChanged. The payload should be similar to other events: `isActive` or `isPictureInPictureActive`
restoreUserInterfaceForPictureInPictureStop - Given that this is mandatory to be called, how about just calling this automatically when we exit picture in picture
startPictureInPicture / stopPictureInPicture - I'm trying to use props wherever possible and avoid functions unless absolutely necessary. I'd suggest changing this to a boolean prop like `pictureInPicture`.

aalzanki:
@cobarx I made 4 of the 5 changes you mentioned. Please let me know what you think about the restoration piece. I tried to mirror Apple's restoration function with: `onRestoreUserInterfaceForPictureInPictureStop` and `restoreUserInterfaceForPictureInPictureStopCompleted`.

aalzanki:
@cobarx any feedback about the updated code?

cobarx:
Sorry for the delay in getting to review this. This is working great, I'm going to merge it.

Two things I noticed:
Do we need the true/false argument to `restoreUserInterfaceForPictureInPictureStopCompleted`? If not, you could just pass true automatically and remove the argument from the README / JS function. If it is needed, please document when it's appropriate to use true vs. false.

Also, I'm not seeing events when I put the video in fullscreen and then engage PiP from there. If possible, can you add a PR to capture those using the same framework you've done here.

cobarx:
Btw this works great. Thanks so much for putting it together, great feature.

It would be wonderful to be able to support this on ExoPlayer as well.

danielmarino24i:
I think this PR breaks compatibility with tvOS, as PiP is only on AVKit (ios)

Probably is can be fixed with a solution like this (probably some condition has to be added also on AVPictureInPictureControllerDelegate delegate's methods ) or `__TVOS_PROHIBITED`:
```
#if __has_include(<react-native-video/RCTVideoCache.h>)
@interface RCTVideo : UIView <RCTVideoPlayerViewControllerDelegate, DVAssetLoaderDelegatesDelegate, AVAssetResourceLoaderDelegate>
#elif TARGET_OS_TV
@interface RCTVideo : UIView <RCTVideoPlayerViewControllerDelegate>
#else
@interface RCTVideo : UIView <RCTVideoPlayerViewControllerDelegate, AVPictureInPictureControllerDelegate>
#endif
```

danielmarino24i:
Also it fails if initial value is set to true (it should be UI/JS triggered) of it is set as initial value, trigger the function on `onLoad` or sth like this

FullstackJack:
@cobarx As noted by @danielmarino24i, this created a regression on tvOS. We'll need to get a PR up to hotfix it.

FullstackJack:
I've spun up a work branch to complete this work for Android, we'll see how it goes.

PR #1327, opened 2018-11-14
Fix: volume should not change on onAudioFocusChange event

ddudek:
#### Update the documentation
Not a new feature

#### Update the changelog
Updated.

1. Set muted property to "true" or volume to anything other than 1.f
2. Play the video, make the audio focus change (e.g. other notification sound playing)
Result: volume is set to 0.8f for a while and then resets to 1.0f
Expected: volume should set to 0.8 of current volume, and then get back to previous volume

#### Describe the changes
I'm saving current volume in a field, and make audio focus change listener use the current volume value instead of harcoded levels.


cobarx:
Well done! Nice diagnosis and fix. 

When muting, we should save the current volume and restore it when unmuting, but that can be a separate fix. Feel free to push a separate PR for that.

PR #1346, opened 2018-11-24
Manual Installation (iOS) - Update onVideoSaved

ajsmth:
this missing proptype was throwing an error with the packager, not sure if this is because I elected to manually install instead of taking the cocoa pods route. in either case, there is an onVideoSaved method declaration in RCTVideoManager that I cant seem to find anywhere else (in the docs or Video.js)


ajsmth:
looks like this was because I was using the wrong package version for my RN version (0.55). downgrading to 3.2.1 works fine 👍 I hope this helps someone else in the future

cobarx:
It looks like this was left over from some of the work done on the `save()` code. I checked and that event isn't used anywhere, so I'm going to remove it. Thanks for pointing this out.

PR #1351, opened 2018-11-28
Fire onSeek after seek completes on Android, document onSeek

cobarx:
The implementation for onSeek was incorrect on both Android MediaPlayer & ExoPlayer. The code would generate a seekEvent before performing the seek, which is pretty much useless. I've updated the code to wait until we receive a seek complete event from the player and then generate the event. That way we can get the actual position post-seek. This matches the behavior for iOS.

This event wasn't documented, so I've added that as well.


dylancom:
Thanks!

PR #1354, opened 2018-11-28
fix: useTextureView default to true

gcottrell1:
This PR makes implements defaulting `useTextureView` to true as per the documentation


cobarx:
Ok I tested and the setUseTextureView function never got called. It makes more sense to handle this in the Java code, so I've updated the default value there and it's working properly for me. I've merged those changes in #1383.

I really appreciate you discovering and reporting this. It's hard to believe I missed it, but I should have done more thorough testing.

PR #1355, opened 2018-11-28
ExoPlayer - Add possibility to hide shutterView (black screen while loading)

goroshevsky:
The purpose of this PR is similar to https://github.com/react-native-community/react-native-video/pull/783, it adds possibility to remove black screen in ExoPlayer while it's rendering first frame, but I made this option configurable via `hideShutterView` prop


cobarx:
Making this configurable is a much better approach. Many thanks for getting this implemented.

PR #1360, opened 2018-12-02
Only apply filter if filter is set

### LINKED ISSUE #1348: iOS: RCTVideo setSrc: will block mainthread for long due to RCT setFilter:
### Current behavior
When I instantiate a RCTVideo, setSrc is called. It in turn calls 
[self playerItemPrepareText:asset assetOptions:assetOptions withCallback:handler];
which calls
handler([AVPlayerItem playerItemWithAsset:asset]); (because I have no text tracks)
which in turn calls 
[self setFilter:_filterName];
setFilter takes between 0.3 and 2.8 seconds on my iPhone 7 Plus.

This all happens on the main thread, and thus I have a main thread block. In my case, since I'm loading multiple videos, I have a real bad hang.

### Reproduction steps
Load a video

### Expected behavior
Immediate return, anything that takes long will be done async

### Platform
iOS 12.1 
react-native-video/Video 4.0.1

nickgzzjr:
This is an attempt to fix [Issue 1348](https://github.com/react-native-community/react-native-video/issues/1348)

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -1265,35 +1265,29 @@ static int const RCTVideoUnset = -1;
 }
 
 - (void)setFilter:(NSString *)filterName {
-
     _filterName = filterName;
-
     AVAsset *asset = _playerItem.asset;
-
-    if (asset != nil) {
-
-        CIFilter *filter = [CIFilter filterWithName:filterName];
-
-        if (filter != nil) {
-
-            _playerItem.videoComposition = [AVVideoComposition
-                    videoCompositionWithAsset:asset
-                 applyingCIFiltersWithHandler:^(AVAsynchronousCIImageFilteringRequest *_Nonnull request) {
-
-                   CIImage *image = request.sourceImage.imageByClampingToExtent;
-
-                   [filter setValue:image forKey:kCIInputImageKey];
-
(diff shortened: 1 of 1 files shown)


ios/Video/RCTVideo.m
```diff
@@ -1274,27 +1274,23 @@ - (void)setFilter:(NSString *)filterName {
 
         CIFilter *filter = [CIFilter filterWithName:filterName];
 
-        _playerItem.videoComposition = [AVVideoComposition
-                videoCompositionWithAsset:asset
-             applyingCIFiltersWithHandler:^(AVAsynchronousCIImageFilteringRequest *_Nonnull request) {
+        if (filter != nil) {
```
niklassaers: Actually, need to test for an empty filter name too. Having set no filter, I get "" here, not nil

niklassaers:
👍 Thank you. 🙏

nickgzzjr:
@niklassaers I had specifically made it to test `filter != nil` only and not "" because on some videos (with rotated orientations) the transformation matrix does not get applied correctly to the `videoComposition` if doesn't start playing with a filter. I think what @cobarx had suggested in [Issue 1348](https://github.com/react-native-community/react-native-video/issues/1348) to add `filterEnabled` is the way to go. I'll try to get a pull request in later this evening.

PR #1387, opened 2018-12-15
Videos Without Text Tracks Should Be Cached

brendangkchan:
The documentation and code logs both say that caching should be bypassed if there are text tracks, however the code says the opposite. This PR fixes such that caching works as expected when there are no text tracks.

In the code the if statement checks that there are NO text tracks:
https://github.com/react-native-community/react-native-video/blob/master/ios/Video/RCTVideo.m#L483

However the debug log inside the if statement says there ARE text tracks:
https://github.com/react-native-community/react-native-video/blob/master/ios/Video/RCTVideo.m#L488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 ## Changelog
 
-### Version 4.2.1
-* Fix cache being bypassed when no text tracks are passed.
-
 ### Version 4.2.0
 * Don't initialize filters on iOS unless a filter is set. This was causing a startup performance regression [#1360](https://github.com/react-native-community/react-native-video/pull/1360)
 * Support setting the maxBitRate [#1310](https://github.com/react-native-community/react-native-video/pull/1310)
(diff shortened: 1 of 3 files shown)


cobarx:
@n1ru4l Can you take a look at this. It looks like the code in has been updated, but I'm the logic is throwing me off a bit...should we be checking for _textTracks set or not set?

cobarx:
@n1ru4l Any thoughts on this?

n1ru4l:
@cobarx Yes it makes sense to only cache if `_textTracks` is empty or nullish.

danielmarino24i:
So if I understand correctly, with new code from master it should be:
```
if (shouldCache && _textTracks &&_textTracks.count) {
```
right?

danielmarino24i:
So if I understand correctly, it is already working on master

```
if (shouldCache && (!_textTracks || !_textTracks.count)) {
```
`playerItemForSourceUsingCache:uri ` is only being called if shouldCache = true and has no textTracks

cobarx:
@danielmarino24i I'm going to close this then. I think this problem got addressed by the change from @n1ru4l that added that code to master.

PR #1395, opened 2018-12-22
Fix iOS video not displaying after switching source

### LINKED ISSUE #1394: Change source uri dynamically
### Current behavior
I am trying to change the video uri dynamically, but it doesn't reload the video when you change it. Does anybody know how to run it?

### Reproduction steps
Change the video uri dynamically.

### Expected behavior
Automatically reload the video component when the uri change. 

### Platform
* iOS

cobarx:
Fixes #1394 

When switching videos on iOS the video loads but does not display. I tracked this down to:
https://github.com/react-native-community/react-native-video/pull/1310/files#diff-42e8804735be64e84d6dd1fb103210a1R861
where applying modifiers while setting the source causes us to lose the player layer and not display the new video. 

@ashnfb Please do your best to do extensive testing around switching sources & different video types before submitting PRs. We need to do everything we can to avoid regressions creeping into releases as this breaks apps for numerous developers. It ended up taking about an hour and a half to track this down which would be best spent on reviewing PRs & fixing bugs.

@n1ru4l heads up on regression

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -858,7 +858,7 @@ static int const RCTVideoUnset = -1;
 
 - (void)setMaxBitRate:(float) maxBitRate {
   _maxBitRate = maxBitRate;
-  [self applyModifiers];
+  _playerItem.preferredPeakBitRate = maxBitRate;
 }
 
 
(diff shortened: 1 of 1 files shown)


Wangzihao5325:
When i switching videos on iOS if i set controls={true} everything is ok ,but if i set controls={false} audio is ok but screen will not change.  version is 4.2.0. sorry for my poor english - -!

alencrowdbotics:
> When i switching videos on iOS if i set controls={true} everything is ok ,but if i set controls={false} audio is ok but screen will not change. version is 4.2.0. sorry for my poor english - -!

The reverse will happen in my case. When i switch source and controls={true} the video freezes audio plays on the background. When controls={false} video reloads.

5.0.2 and 5.1.0-alpha8

alencrowdbotics:
For anyone else facing this issue try to add

        maxBitRate={anyvalue}

For our case either the controls on or off froze the video. Somehow it affects the bitrate maybe!

PR #1420, opened 2019-01-10
Fixed a mistake in the documentation about the Headers.

rinonb:
### A small change in the documentation regarding headers.

The "headers" property was listed as:
```
headers={{
  Authorization: 'bearer some-token-value',
  'X-Custom-Header': 'some value'
}}
```
Which makes it look like it's a prop, but in reality, is just a property of the source object.
```
source={{
  uri: "https://www.example.com/video.mp4",
  headers: {
    Authorization: 'bearer some-token-value',
    'X-Custom-Header': 'some value'
  }
}}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,15 +420,18 @@ Platforms: iOS
 Platforms: iOS
 
 #### headers
-Pass headers to the HTTP client. Can be used for authorization.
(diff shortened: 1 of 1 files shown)


n1ru4l (approved):
Thank you for your contribution!

PR #1446, opened 2019-01-24
Add maxResolution support for iOS

am33t:
Added maxResolution support for iOS. This will allow to limit bandwidth consumption if multiple streams are available in hls.

Example:
```
maxResolution={width:360, height: 180}
```


danielmarino24i (approved):

cobarx:
@am33t I started testing this and it looks like setting either height or width to 0 will completely disable the maxResolution limit. Do you know a way to check the currently selected resolution so I can verify whether this is the case and make sure the documentation is as exact as possible.

It would be good to have a resolution change event that we can use for testing & diagnosing things like this.

cobarx:
By verifying, I mean do you know what function I could call in Xcode to print the current resolution?

danielmarino24i:
I think that's the desired behaviour on AVPlayer
https://developer.apple.com/documentation/avfoundation/avplayeritem/2867324-preferredmaximumresolution

```
Defaults to CGSizeZero, which indicates there is no limit on the video resolution. Any other value indicates a preferred maximum video resolution. This property only applies to HTTP Live Streaming assets.
```

danielmarino24i:
@cobarx maybe with
https://developer.apple.com/documentation/avfoundation/avassettrack
`naturalSize`

Pseudo: `let naturalSize = player?.currentItem?.tracks.first?.assetTrack?.naturalSize`

cobarx:
@danielmarino24i It would be nice to be able to override the default behavior. For example, to force 480p or lower, you'd want to set height 480 but leave width unset or change it to 0. So then in the Objective C code we'd need to remap 0 to some super large number so that it would get ignored. It would work but is kinda hacky.

I'll give that code you suggested a shot when I have a chance.

flexgrip:
I'm looking at this for background playback. My HLS manifest is a demuxed list of video and audio tracks.

On android, it works great. I set the video track to disabled when the app is in the background, then just the audio plays. But on iOS, I have no ability to disable the video track.

PR #1448, opened 2019-01-24
Feature Implementation: Recovery from transient internet failures

sridhard:
Thanks for opening a PR!

#### Update the documentation
Documentation updated

#### Update the changelog
Will update the changelog

Send failureRetryCount={10} in props. When video play is active just turn-off the internet/wifi for 1-2 seconds and  turn-on immediately. Video play should continue after the wifi is turned-on

ok

#### Describe the changes
This is a new feature to recover from transient internet issues. In mobile network the connection will not be stable and in case of transient internet issues video play should recover and continue to play once the connection is ok. Exoplayer has a feature to configure the retry count. This PR uses the exoplayer feature to recover from transient internet issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Fix text not appearing in release builds of Android apps [#1373](https://github.com/react-native-community/react-native-video/pull/1373)
 * Update to ExoPlayer 2.9.3 [#1406](https://github.com/react-native-community/react-native-video/pull/1406)
 * Add video track selection & onBandwidthUpdate [#1199](https://github.com/react-native-community/react-native-video/pull/1199)
+* Recovery from transient internet failures and props to configure the custom retry count [#1448](https://github.com/react-native-community/react-native-video/pull/1448)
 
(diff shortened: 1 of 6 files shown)


sridhard:
@cobarx please review and merge this feature. I have tested in low bandwidth and unstable 3g/4g/wifi connections. This feature will retry and autoreconnect when the internet connects again.

sridhard:
@cobarx can you please check and merge this PR

cobarx:
Thanks for handling all the guidelines in the PR checklist, that makes my life a lot easier. 

I made a slight change the naming to bring it closer to what ExoPlayer calls it. I also reverted your change to the INDEX_UNSET code for video.  I'm guessing based on the code that you were just trying to remove a nested if. However, it has a slight change to the behavior - if there are no groups but it's a video, it would drop into the second case. I'm not sure if that will cause problems...I spent quite a bit of time on that code and it might be necessary for it to work that way, unless you've tested it quite a bit I'd rather not risk it.

PR #1468, opened 2019-02-06
Fix: Can't disable subtitle on iOS when using type: disabled

### LINKED ISSUE #1144: iOS Sidecar VTT TextTrack toggling does not work
### Current behavior 
When attempting to toggle captions on and off via the selectedTextTrack prop, the video stops and freezes. I can no longer play the video after attempting to toggle. This happens for all videos I have tried.

Note, toggling captions works beautifully on Android.

### Reproduction steps
Currently I have a state variable for captions. I control the toggle through a button and set the type of selectedTextTrack to "disabled" or "title". 

i.e.
In the video component:
selectedTextTrack={this.state.captions}
textTracks={[
  {
    title: "English",
    language: "en",
    type: TextTrackType.VTT, // "text/vtt"
    uri: "https://bitdash-a.akamaihd.net/content/sintel/subtitles/subtitles_en.vtt"
  },

In a button press event:
toggleCaptions = () => {
    this.setState({
      captions:{
        type: this.state.captions.type === "disabled" ? "title" : "disabled",
        value: "English"
      }
    })
  }

### Platform
* iOS

nbennink:
This is a PR to fix #1144. 

### The original issue
When you select a textTrack it starts playing the video with subtitles. To disable these subtitle you can give the `setSelectedTextTrack={{type:'disabled'}}` to the Video component. This works correctly on Android, but on iOS it "freezes" the last known subtitle until the end of the video.

### How does this PR fix this?
There's been a discussion about how to fix this problem. A solution was to generate an empty VTT file with a package like RNFetchBlob. I think however that it's a better solution to generate this file in the Video component itself (since that's where the issue comes from). 

This PR adds an VTT file to the array of TextTracks. When the selectedTextTrack gets `type: disabled` it will select this track which makes the "old" track disappear.

Nothing changes in how you use texttracks. Something to note is that you can't load a completely empty VTT file. So there is a dot at the 98 hour timemark (basically never). 

Edit: I only load this empty VTT file when you actually have other texttracks as well. So you don't have the issue where it breaks Airplay.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -58,7 +58,7 @@ static int const RCTVideoUnset = -1;
   BOOL _paused;
   BOOL _repeat;
   BOOL _allowsExternalPlayback;
-  NSArray * _textTracks;
+  NSMutableArray * _textTracks;
   NSDictionary * _selectedTextTrack;
   NSDictionary * _selectedAudioTrack;
   BOOL _playbackStalled;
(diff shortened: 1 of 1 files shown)


danielmarino24i (approved):

cobarx:
@nbennink This is a very clever and intelligent solution. I haven't had a chance to test this yet. My only concern would be whether the empty VTT track appears in the list of text tracks reported by onLoad. If so, we should filter it out.

nbennink:
@cobarx Yes you are right. My latest commit removes the subtitle file from the onLoad event. So you can't see it in your front-end and the only way to select it, is by using `{type: "disabled"}` (just like Android).

danielmarino24i:
Can this change conflict with this?
https://github.com/react-native-community/react-native-video/blob/d8df6fe01f70ab8bdb0dcc79a64f38b3f54be304/ios/Video/RCTVideo.m#L510

nbennink:
I don't think that it's a problem, here in the [playerItemPrepareText method](https://github.com/react-native-community/react-native-video/pull/1468/files#diff-42e8804735be64e84d6dd1fb103210a1R410) the function gets returned when you don't give the video player text tracks using the `textTracks` prop. So the empty subtitle is not available when you don't give the player at least one text track.

This was done to prevent an issue where you can't Airplay a video that contains text tracks. I'm not using caching for my project right now, but I assume that it works the same here.

nbennink:
So I recently discovered that this is not working on tvOS. It appears that [this](https://github.com/react-native-community/react-native-video/blob/d8df6fe01f70ab8bdb0dcc79a64f38b3f54be304/ios/Video/RCTVideo.m#L467) line returns nil so it skips the empty VTT file. I think that this is limitation of the tvOS file system or AVPlayer on tvOS. I'm looking for a solution to fix it. Let me know if someone has suggestions.

CHaNGeTe:
@nbennink Just a quick question, are you using mp4? or adaptative streaming?
I have detected some issues with texttracks on Adaptative ones.

I was thinking in another approach for setting subtitles.

nbennink:
@CHaNGeTe I am using mp4 files for subtitles and still have this issue. I'm not sure what the best option is to replace the current approach for subtitles. I'd like to fix the problem without the empty VTT file. 

Edit: I recently came back to this issue and tried again to come up with solution for tvOS but did not get it working. Anyone a suggestion?

kotomono:
same problem here. why isnt this merged? tested and it works

PR #1475, opened 2019-02-09
Add support for setting audio attributes in ExoPlayer

hjames9:
<Video
  audioUsage={'alarm'} //Sets audio attribute usage to "USAGE_ALARM" (default is USAGE_MEDIA)
  audioFlags={'audibility_enforced'} //Sets audio attribute flag to "FLAG_AUDIBILITY_ENFORCED" (default is none)
  audioContentType={'speech'} //Sets audio attribute content type to "CONTENT_TYPE_SPEECH" (default is CONTENT_TYPE_UNKNOWN)
/>

Set the "audioUsage" to "alarm", mute the main "music" audio stream on Android, and watch as you're able to hear the audio from the video via a different stream.

#### Describe the changes
This PR allows you to set the various audio attributes on an ExoPlayer instance via the <Video> element.  Currently, this library lacks the ability to do so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### next
+* Add support for setting audio attributes in ExoPlayer [#1475](https://github.com/react-native-community/react-native-video/pull/1475)
+
 ### Version 4.4.0
 * Fix runtime warning by replacing `UIManager.RCTVideo` with `UIManager.getViewManagerConfig('RCTVideo')` (and ensuring backwards compat) [#1487](https://github.com/react-native-community/react-native-video/pull/1487)
 * Fix loading package resolved videos when using video-caching [#1438](https://github.com/react-native-community/react-native-video/pull/1438)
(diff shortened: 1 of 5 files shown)


hjames9:
@cobarx Please let me know if these changes make sense.  Thanks!

cobarx:
Yes it makes sense. Ideally, we need a common API that will work on iOS & Android, so I'm hesitant to merge something that's very specific to the way Android handles things and then have a completely different API down the road for iOS.

I did a writeup covering what I found out how iOS & Android handle audio configuration here:
https://github.com/react-native-community/react-native-video/issues/1103
I still think that API makes sense. However, we can add the configurable options for the type of the content to supplement it. I also curious if we need to handle audio focus.

We also need to think about if there are multiple media items playing at the same time. Expo does an extremely good job of this, they evaluate all the items with audio and give the item that needs the most permissions priority.

Would you be open to spending a bit of time hashing out a cross platform API (with me) before we start looking the PR. Also, do you have time to tackle this for iOS as well. I'd like to get this all in place at once if possible. This is probably the most requested feature, so it's very important we get this right the first time.

hjames9:
@cobarx I'll be happy to help hash out a different API that would be more cross platform, but unfortunately I don't have the gear to do the iOS work.  Will gladly do the Android changes if that helps though

madandrija:
Also, we've forked the repo since we needed the changes in this PR on Android only (on iOS the defaults work as we would want them) and it's working great out of the box so if anyone else needs until it's merged, I definitely recommend.

PR #1478, opened 2019-02-11
Throw an error instead of crashing app when video source is empty

laurent22:
If an undefined source is accidentally loaded in react-native-video a fatal error SIGABRT will happen and crash the application. This is very difficult to debug since there's apparently no relation between this crash and react-native-video. So this change checks if the URI is empty and, if it is, it throws an error, allowing the user to find out early about the issue.

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -219,6 +219,8 @@ export default class Video extends Component {
     if (uri && uri.match(/^\//)) {
       uri = `file://${uri}`;
     }
+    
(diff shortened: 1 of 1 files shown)


danielmarino24i (approved):

cobarx (approved):

cobarx:
I thought about this a little more and I think it makes sense to combine this with #1246 where we handle it in the native code so it's the same as ExoPlayer. I'm going to make this a warning.

PR #1487, opened 2019-02-17
replaced UIManager.RCTVideo >  UIManager.getViewManagerConfig('RCTVideo')

nitishxyz:
#### Describe the changes
just updated  UIManager.RCTVideo >  UIManager.getViewManagerConfig('RCTVideo') to remove warnings from the debug app.

Closes #1466

PR #1488, opened 2019-02-19
fix: access on undefined variable

n1ru4l:
This should fix the issues introduced in #1487


n1ru4l:
@cobarx could you please test if this works? Did it on my phone 😄

cobarx:
Well done sir, your phone IDE skills are impressive :)

Working great for me on RN 57, hoping it will be ok for 58 as well.

PR #1493, opened 2019-02-22
Call playbackRateChange onPlay and onPause

entropitor:
This already happens on iOS, I'm also doing this on Android now

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -585,6 +585,7 @@ class ReactExoplayerView extends FrameLayout implements
     @Override
     public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
         String text = "onStateChanged: playWhenReady=" + playWhenReady + ", playbackState=";
+        eventEmitter.playbackRateChange(playWhenReady && playbackState == ExoPlayer.STATE_READY ? 1.0f : 0.0f);
         switch (playbackState) {
             case ExoPlayer.STATE_IDLE:
                 text += "idle";


danielmarino24i:
I think they way it is implemented right now it will fire multiple playbackRateChange = 0 event when f.e. the player transitions from STATE_PREPARING to STATE_READY

entropitor:
Would you advise/prefer to keep the old state so we can check this?

danielmarino24i:
I am trying to think on the big picture, and when people will use this event.

Also we have to check if setting a playbackRate higher than 1, pause, and play again Exoplayer restores the previous speed or sets it to 1 (if I have the chance to test it I will let you know)

Also following the title of the PR (call the event when pause / play) maybe it makes more sense to emit the event on `setPlayWhenReady`

What do you think @cobarx ?

cobarx:
@danielmarino24i Good questions. 

I'm fairly certain that a custom rate will be preserved through play & pause. It's good to check though. Along those lines, this PR needs to add code to emit an event when the rate is changed to a value other than 1.

As far as the events go, I would key it off of `onPlayerStateChanged` like @entropitor is doing. The reason for this is you want base events off of changes to state rather than functions that result in a change to state - if the function fails or takes a long time, the event will not be accurate. In addition using the state allows you to capture all kinds of state changes like idle, ended, buffering, etc. that cause rate to go to 0. 

To prevent the duplicate events, I would keep a var with the last reported rate. Whenever the player state changes, calculate the current rate, compare to last reported, and then update the var & issue the event if it changed.

cobarx:
It would also be good to document this event as part of this PR. I had no idea it existed lol! Let's make sure we note that the player buffering or pausing will report a rate change.

richardwu:
Is there something I can help with to merge this so one can listen to play/pause events? This is also a highly upvoted issue in #2478

hueniverse:
If we can get quick review we can merge this.

PR #1499, opened 2019-02-26
Adding compileOptions Java 1.8 to build.gradle

### LINKED ISSUE #1464: Add Complie options in build.gradle
Hi , I am trying to automate the react native project using detox and everytime when i try to build using detox it gives an error 

We have solved this by adding compile Options in build.gradle as 
compileOptions {
        targetCompatibility 1.8
        sourceCompatibility 1.8
    }

Now the issue is that everytime when i try to install node modules it needs to be added again and also its failure on CI 

Can we do something like this that it picks from the app/build.gradle or adding the above

reime005:
#### Issue description
Android build for Detox E2E tests does not work (see #1464). This PR should fix that.

#### Versions for reproducing
react-native 0.57.8 
detox 10.0.10

Fixes #1464

### DIFF
--- a/android-exoplayer/build.gradle
+++ b/android-exoplayer/build.gradle
@@ -7,6 +7,11 @@ def safeExtGet(prop, fallback) {
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 27)
     buildToolsVersion safeExtGet('buildToolsVersion', '27.0.3')
+ 
+    compileOptions {
+        targetCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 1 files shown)


yogeshthanvi:
any updates for this merge

yogeshthanvi:
Hi @reime005  any idea when it will be updated

cobarx:
Merged. If we add this I take it we no longer need to add this to the app's build.gradle and can remove those notes from our README?

yogeshthanvi:
@cobarx  Working fine checkout out the changes, thanks a lot for help . When we going to release new version with this update ?

PR #1501, opened 2019-02-26
Changed iOS check for video landscape/portrait

Southporter:
Any video taken in Portrait should do the trick. Access this video using react-native-video. Hook up an `onLoad` listener. Before the change, naturalSize would be of the shape 
```javascript
naturalSize: {
  orientation: 'landscape',  // Occasionally it would correctly report as 'portrait'
  width: landscapeWidth,  // Typically the large of the two values
  height: landscapeHeight,
}
```
After the change, naturalSize for portrait should be of the shape
```javascript
naturalSize: {
  orientation: 'portrait',
  width: portraitWidth, // Should be the smaller of the two values
  height: portraitHeight, // Should be the larger of the two values
}
```

Landscape videos should still accurately report as landscape with the correct dimensions

#### Describe the changes
iOS returned `natrualSize` differently from Android. This PR seeks to get these two in alignment and to accurately detect if the video is portrait or landscape.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -617,17 +617,26 @@ static int const RCTVideoUnset = -1;
           width = [NSNumber numberWithFloat:videoTrack.naturalSize.width];
           height = [NSNumber numberWithFloat:videoTrack.naturalSize.height];
           CGAffineTransform preferredTransform = [videoTrack preferredTransform];
-          
-          if ((videoTrack.naturalSize.width == preferredTransform.tx
-               && videoTrack.naturalSize.height == preferredTransform.ty)
-              || (preferredTransform.tx == 0 && preferredTransform.ty == 0))
+
+          // Landscape transform is of the shape:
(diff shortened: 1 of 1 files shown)


cobarx:
Is there any risk that ht

cobarx:
I tested this with two portrait mode videos and they report as being landscape:
http://www.exit109.com/~dnn/clips/RW20seconds_1.mp4
http://hamptonmaxwell.com/videos/treadmill.mp4

I'm not too familiar with the video metadata for determining orientation and rotation. Should these also be detected as portrait? If so would you mind adding a check that covers them.

cobarx:
@ssedrick Can you take a look at the feedback I left and see if it's appropriate to what you've done with this PR.

Southporter:
@cobarx Sorry for not getting back to you. I changed jobs, and no longer have access to the repo that was using this. I'll try to get a setup to where I can test this and look into those two files. It may take a while though.

Issue #1417, opened 2019-01-08
Audio not playing on Android (v 4.3.1)


jmrhvl:
### Current behavior
Video plays fine, but audio files (mp3, m4a) produces the following error:

`
E/ExoPlayerImplInternal: Internal runtime error.
    java.lang.ArrayIndexOutOfBoundsException: length=0; index=-1
        at com.google.android.exoplayer2.source.TrackGroupArray.get(TrackGroupArray.java:65)
        at com.google.android.exoplayer2.trackselection.DefaultTrackSelector.selectTracks(DefaultTrackSelector.java:1264)
        at com.google.android.exoplayer2.trackselection.MappingTrackSelector.selectTracks(MappingTrackSelector.java:394)
        at com.google.android.exoplayer2.MediaPeriodHolder.selectTracks(MediaPeriodHolder.java:160)
        at com.google.android.exoplayer2.MediaPeriodHolder.handlePrepared(MediaPeriodHolder.java:141)
        at com.google.android.exoplayer2.ExoPlayerImplInternal.handlePeriodPrepared(ExoPlayerImplInternal.java:1549)
        at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:318)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:154)
        at android.os.HandlerThread.run(HandlerThread.java:61)
`

### Reproduction steps
Try to play back mp3 or m4a files.

### Expected behavior
The audio would start playing.

### Platform
Which player are you experiencing the problem on:
* Android ExoPlayer

### Video sample
https://telavaag.sfnm.build/en/post4intro.mp3

RN is at version 0.57.8.

Edit: This might be related to #1408, but the error message is not 100% the same and version 4.3.1 seems to have that fix.

Also, before the error previously mentioned I see these in LogCat:

`
I/VideoCapabilities: Unsupported profile 4 for video/mp4v-es
I/OMXClient: MuxOMX ctor
I/ACodec: codec does not support config priority (err -2147483648)
`

supagu:
I have the same problem 4.2.0 works
I think this is the same bug as mentioned in the description.

dkoo761:
+1, downgrading to 4.2.0 until this is fixed. I don't see any errors in the logs but my MP3s don't start playing at all, they just get stuck at time=0.

kesha-antonov:
[...]

PR #1529, opened 2019-03-19
Fix Android audio crash fixes Issue #1417

Mrjaco12:
Fix Issue #1417

Why:

It appears that in [this commit](https://github.com/react-native-community/react-native-video/commit/2d016e7e6ab0bbe4b4ac765093ce607668c991bd) the check added to ensure group lengths are greater than 0 still allows for a case in which the groupIndex is -1 but the group.length == 0 so we are never breaking for the invalid groupIndex.

This change addresses the need by:

* Don't chain check for INDEX_UNSET to previous conditional

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -984,7 +984,9 @@ class ReactExoplayerView extends FrameLayout implements
                     tracks[j] = j;
                 }
             }
-        } else if (groupIndex == C.INDEX_UNSET) {
+        } 
+
+        if (groupIndex == C.INDEX_UNSET) {
             trackSelector.setParameters(disableParameters);
             return;
         }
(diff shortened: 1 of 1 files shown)


rborn:
@Mrjaco12 ❤️

I can confirm the fix works, tested on android 9 (nokia 7+) with `audioOnly` and an mp3 file
It doesn't work without the fix

wangliang1124:
Good job.

rborn:
@cobarx hey, any chance to have a look at this?

Also any chance to reply to @kelset  related to the React Native Core Contributors org we're trying to setup (well, he's doing all the work)? 🤗 https://github.com/react-native-community/meta/blob/master/MAINTAINERS.md

kesha-antonov:
Nice!

cobarx:
Ok sorry for the delay on this, I haven't been as active lately. 

Yes, this will work. It took me a bit of looking but you were right on the money. The nested check whether there are video groups or not through me for a loop on why unchaining worked. So I moved that into the main if check to make that more readable and added some comments.

I'll reach out to @kelset tonight.

PR #1558, opened 2019-04-17
add stop video function

kbourchbane:
Thanks for lib!
i add a stop video function because for android 6 the app crash when the user navigate or change page without stopping video

### DIFF
--- a/README.md
+++ b/README.md
@@ -1080,6 +1080,15 @@ this.player.presentFullscreenPlayer();
 
 Platforms: Android ExoPlayer, Android MediaPlayer, iOS
 
+#### 
+`stopVideoPlayer()`
+Stop video, in componentWillUnmount for example
+
+Example:
+```
+this.player.stopVideoPlayer();
+```
+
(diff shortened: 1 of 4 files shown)


danielmarino24i:
Maybe it is something to have in consideration for `componentWillUnmount`

sasaluo:
how to stop or remove the video? I have the same problem now,when changing pages from A to B and  both pages have video, the app could not  work

sasaluo:
i tried the this.player.stopVideoPlayer() method,however app crashed in android tv

jensellfors:
What is the use case for this? If you want stop the video you can either pause it or remove it from your view, that should stop and deallocate it.

If we introduce a stop function then it needs to be available on all platforms, not just ExoPlayer. It should also probable be a method and not a prop.

siokstoksanoks:
@jenshandersson  when component is unmounting that function stop is needed.
Wouldn't be setting state for unmounting component to change prop for this component.

PR #1612, opened 2019-06-10
Fix the code that causes unnecessary re-rendering

ifsnow:
As you all know, unnecessary re-rendering can cause performance issues.

- Using `PureComponent`
It's generally more efficient than not.

- Changed the logic
Rendering occurs multiple times even if we don't use a `poster`.

Please consider it positively. I hope this helps.

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -1,4 +1,4 @@
-import React, {Component} from 'react';
+import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {StyleSheet, requireNativeComponent, NativeModules, View, ViewPropTypes, Image, Platform, findNodeHandle} from 'react-native';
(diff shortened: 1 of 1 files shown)


CHaNGeTe:
Sounds good for me.
The only caveat I see is nested objects, for people who has the library already integrated and using object refs and mutating them, but I guess it is not so common.

CHaNGeTe:
Analyzing it deeper I found possible problems on:
- `source` can have nested props, `headers` is nested right now.
- array props as `textTracks` and `audioTracks` can lead to problems.

Maybe is not a big problem because normally these props are changed in block: I mean, you set a whole new source object, and usually you don't mutate these textTracks and audioTracks array.

But I wanted to put it here to have them in consideration.

cobarx:
We need to keep `headers` nested as sometimes they are required to authenticate the source.

PR #1625, opened 2019-06-18
Replace RCTBubblingEventBlock  by RCTDirectEventBlock to avoid event name collision with Expo AV

CHaNGeTe:
#### Update the changelog
Pending, will do tomorrow.

Test case with expo is reported here: https://github.com/react-native-community/react-native-video/issues/1618

#### Describe the changes
Replaces RCTBubblingEventBlock for RCTDirectEventBlock.
As bubbling events are recommended for DOM-like events (scroll, focus, ...) that need to propagate.
This avoids collision on event names with the expo av library. As `DeviceEventEmitter` is shared among all libraries.

Ref: https://gist.github.com/chourobin/f83f3b3a6fd2053fad29fff69524f91c#ios
```Notes: Bubbling events are like DOM events so that a parent component can capture an event fired by its child. Generally these are UI-related, like "the user touched this box". Direct events are not bubbled and are intended for more abstract events like "this image failed to load".```

Closes https://github.com/react-native-community/react-native-video/issues/1618

Tomorrow I will test that events are firing correctly, but I open the PR so people can test it easily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 
 ### next
+* Replaced RCTBubblingEventBlock events by RCTDirectEventBlock to avoid event name collisions [#1625](https://github.com/react-native-community/react-native-video/pull/1625)
 * Added `onPlaybackRateChange` to README [#1578](https://github.com/react-native-community/react-native-video/pull/1578)
 
 ### Version 4.4.1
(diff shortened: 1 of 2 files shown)


jensellfors:
I've tested and it works. We should change the block types in RCTVideo.h as well though.

danielmarino24i:
Updated with changes on RCTVideo.h as well
+ Changelog

CHaNGeTe:
+ Tested with few events (onLoad, onProgress, onBuffer, ...) and it is working.

jensellfors (approved):

PR #1627, opened 2019-06-19
Fixes bug where poster and video was displayed simultaneously

jensellfors:
When using poster image and not using absolute positioning of the Video element you end up with both video and poster elements. This was introduced in #1167. My fix to revert some changes but keep the benefits of no flashing.

This fix relies on `onReadyForDisplay` callback to hide the poster. It's been tested on iOS and ExoPlayer. I added it to DOM as well but haven't been able to test...

Fixes [1509](https://github.com/react-native-community/react-native-video/issues/1509)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### next
 * Added `onPlaybackRateChange` to README [#1578](https://github.com/react-native-community/react-native-video/pull/1578)
+* Added `onReadyForDisplay` to README [#1627](https://github.com/react-native-community/react-native-video/pull/1627)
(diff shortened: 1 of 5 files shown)


jensellfors:
@CHaNGeTe I'm happy with this one now, but need help testing. Especially on android and the ExoPlayer. We need to verify that `onReadyForDisplay` is called for all platforms, otherwise the poster won't be removed. What do you guys think of this solution?

jensellfors:
@CHaNGeTe I just pushed a commit that hides the poster in `onLoad` as `onReadyForDisplay` is not implemented. (A good example why maintaining several platforms is hard ;p)

CHaNGeTe (approved):

n1ru4l (approved):
Legit!

IamMasterWayne:
"react-native": "0.61.5"
"react-native-video": "^5.0.2"

React Native Android Black Screen For Android

PR #1720, opened 2019-08-20
Removing the call to observeValueForKeyPath:ofObject:change:context: …

nmarisi:
Removing the call to observeValueForKeyPath:ofObject:change:context: on super from RCTVideo.

If the super class is not actually observing the key, the app will crash. Checking to see if
the super class responds to this selector doesn't solve this issue.

react-native-video GitHub issue: https://github.com/react-native-community/react-native-video/issues/1515

Discussion about this particular problem: https://stackoverflow.com/questions/6574714/whats-wrong-with-this-observevalueforkeypathofobjectchangecontext-implement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 ## Changelog
 
+### Version 5.0.1
+* Fix crash when RCTVideo's superclass doesn't observe the keyPath 'frame' (iOS) [#1720](https://github.com/react-native-community/react-native-video/pull/1720)
+
 ### Version 5.0.0
 * AndroidX Support
 
 ### Version 4.4.4
 * Handle racing conditions when props are setted on exoplayer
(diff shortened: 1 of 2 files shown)


CHaNGeTe (approved):

zek:
When can we merge this PR?
I got a lot of error reports from crashlytics

<RCTVideo: 0x14c91d650; frame = (0 0; 320 568); layer = <CALayer: 0x2834f3760>>: An -observeValueForKeyPath:ofObject:change:context: message was received but not handled. Key path: status Observed object: <AVPlayerItem: 0x283706bf0, asset = <AVURLAsset: 0x2834577a0, URL = >> Change: { kind = 1; } Context: 0x0
-[RCTVideo observeValueForKeyPath:ofObject:change:context:]

PR #1722, opened 2019-08-22
Update exoplayer to 2.10.4

florianalbrecht:
This PR updates expoplayer to 2.10.4 and replaces deprecated methods for creating media sources.


benoitdion:
This was done in #1753! We'll release the change in a 5.1-alpha

waheedakhtar694:
@benoitdion confliction is been still there. React-native-video conflicts with react-native-track-player because both used exo-player and as a result video is not shows on android only audio is been playing.

PR #1740, opened 2019-09-05
[iOS] RCTVideoDelegate

sidferreira:
Adding a delegate will allow developers to extend `react-native-video` with other libraries (including ads).

We're using it to create a Google IMA SDK.

Notes:
- Android changes coming soon
- See this as an open discussion / WiP

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -364,53 +364,67 @@ static int const RCTVideoUnset = -1;
     
     // perform on next run loop, otherwise other passed react-props may not be set
     [self playerItemForSource:self->_source withCallback:^(AVPlayerItem * playerItem) {
-      self->_playerItem = playerItem;
-      _playerItem = playerItem;
-      [self setPreferredForwardBufferDuration:_preferredForwardBufferDuration];
-      [self addPlayerItemObservers];
-      [self setFilter:self->_filterName];
-      [self setMaxBitRate:self->_maxBitRate];
-      
-      [_player pause];
-        
-      if (_playbackRateObserverRegistered) {
-        [_player removeObserver:self forKeyPath:playbackRate context:nil];
-        _playbackRateObserverRegistered = NO;
+      bool delegateShouldSetup = false;
+      if (self->_rctVideoDelegate != nil &&
+        [self->_rctVideoDelegate respondsToSelector:@selector(shouldSetupPlayerItem:forSource:)]) {
+        delegateShouldSetup = [self->_rctVideoDelegate shouldSetupPlayerItem:playerItem forSource:source];
       }
-      if (self->_isExternalPlaybackActiveObserverRegistered) {
-        [self->_player removeObserver:self forKeyPath:externalPlaybackActive context:nil];
-        self->_isExternalPlaybackActiveObserverRegistered = NO;
-      }
-      
-      self->_player = [AVPlayer playerWithPlayerItem:self->_playerItem];
-      self->_player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
-      
-      [self->_player addObserver:self forKeyPath:playbackRate options:0 context:nil];
-      self->_playbackRateObserverRegistered = YES;
-      
-      [self->_player addObserver:self forKeyPath:externalPlaybackActive options:0 context:nil];
-      self->_isExternalPlaybackActiveObserverRegistered = YES;
-      
-      [self addPlayerTimeObserver];
-      if (@available(iOS 10.0, *)) {
-        [self setAutomaticallyWaitsToMinimizeStalling:_automaticallyWaitsToMinimizeStalling];
-      }
-
-      //Perform on next run loop, otherwise onVideoLoadStart is nil
-      if (self.onVideoLoadStart) {
-        id uri = [self->_source objectForKey:@"uri"];
-        id type = [self->_source objectForKey:@"type"];
-        self.onVideoLoadStart(@{@"src": @{
-                                    @"uri": uri ? uri : [NSNull null],
-                                    @"type": type ? type : [NSNull null],
-                                    @"isNetwork": [NSNumber numberWithBool:(bool)[self->_source objectForKey:@"isNetwork"]]},
-                                @"drm": self->_drm ? self->_drm : [NSNull null],
-                                @"target": self.reactTag
-                                });
+      if (!delegateShouldSetup) {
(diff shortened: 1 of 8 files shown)


CHaNGeTe:
I see this as a very interesting feature, I kinda hacked it to allow Verimatrix streams, but like your delegate's approach.

Can you add readme with instructions and changelog?

ios/Video/RCTVideo.m
```diff
@@ -350,35 +350,52 @@ - (void)setSrc:(NSDictionary *)source
...
-      [self addPlayerItemObservers];
-      [self setFilter:_filterName];
-      [self setMaxBitRate:_maxBitRate];
-      
-      [_player pause];
-        
-      if (_playbackRateObserverRegistered) {
+        if (self->_rctVideoDelegate == nil ||
+            ![self->_rctVideoDelegate respondsToSelector:@selector(willSetupPlayerItem:forSource:)] ||
+            ![self->_rctVideoDelegate willSetupPlayerItem:playerItem forSource:source])
```
jensellfors: I don't really get this IF-statement. If `willSetupPlayerItem:playerItem` returns true you don't want to call `setupPlayerItem:playerItem`?

jensellfors:
@sidferreira thanks for this, I too like this approach 👍. Is there a similar approach we can do for Android?

ios/Video/RCTVideo.m
sidferreira: No, it's like:
- if no delegate
- or delegate doesn't have method
- or delegate's method returns false

Then call the usual code

jensellfors: Sure, that's what I wrote (but other way around). Are you doing the actual set up in `willSetupPlayerItem:playerItem`? I thought `willSetupPlayerItem:playerItem` was just a callback letting the delegate know it was about to set a new playerItem...

Shouldn't `setupPlayerItem` always be called? Otherwise we don't set up all the observers. Looks like `onVideoLoadStart` will not be called if we use your delegate method?

Can you please elaborate on what it is you are doing here, instead of me guessing :)

sidferreira: Oh! indeed @jenshandersson ! It makes sense now!

It seems that the problem is to find the best writing.

I wrote that method to be read like "Will the `delegate` setup the `player`?" If `NO`, then `RCTVideo` will set up it.

Google IMA requires me to make a network request BEFORE having the real `AVPlayerItem` and `AVPlayer`. So I need to "delay" the player setup.

You can see it here:
- [RCGoogle IMA - Checks if has all requirements for ad playback](https://github.com/tedconf/react-native-google-ima/blob/42346b20eebfbc8d4c75420cf3a65b9697e769ca/ios/RNGoogleIMA.m#L96)
- [RCGoogle IMA - Request the ads](https://github.com/tedconf/react-native-google-ima/blob/42346b20eebfbc8d4c75420cf3a65b9697e769ca/ios/RNGoogleIMA.m#L146)
- [RCGoogle IMA - Set up player when all AD content is loaded](https://github.com/tedconf/react-native-google-ima/blob/42346b20eebfbc8d4c75420cf3a65b9697e769ca/ios/RNGoogleIMA.m#L181)
  - I don't like calling line 184 but I can't find a better way yet. Maybe if we force calling it in [RCTVideo.m#370](https://github.com/react-native-community/react-native-video/blob/5289def3ba9921c3f1d9c8adfe2be6578e010225/ios/Video/RCTVideo.m#L370) but I'm not sure the side-effects of it.

Maybe rename to `should` instead? So the `delegate` will tell if `RCTVideo` should set it up?

Also, I made it boolean so I can just "let `RCTVideo` handle everything" instead.

sidferreira:
About Android, I still don't fully check how to proceed. 
There's a big chance that actually someone else will make it (will talk to my superior on that)

It seems to me that we need to add a handling code to both `setDataSource` and `setRawData`, but AFAIK we can use similar delegate approach.

sidferreira:
I'm also wondering a way to set extra parameters in `setSrc`. Google IMA requires 2 or 3 parameters to set up the ads, and I would love to have them in the `source` object.
Any suggestion?

sidferreira:
Although I need to rebase my changes, any position about this?

sidferreira:
I just rebased the changes to master. 

@jenshandersson is there anything I can do to speed up the merge of this PR?

ios/Video/RCTVideo.m
jensellfors: Ok I get it now, but would make more sense to call it `should`.

sidferreira: My reading is 
`will (the delegate) setup the player (later)?` 
but can also be read like 
`delegate should setup the player?` indeed... 

Will make the change

sidferreira:
@CHaNGeTe @jenshandersson

CHaNGeTe:
@sidferreira sorry for late reply.

I think an example of a file implementing the delegate (even if it is kinda empty) will be really helpfull

sidferreira:
@CHaNGeTe would a link to our implementation be an option?

ATM it is here: https://github.com/tedconf/react-native-google-ima/blob/master/ios/RNGoogleIMA.m#L192

sidferreira:
Hey @CHaNGeTe , although I still have to re-merge and test the changes against the newer version, I would like to know how can I move forward with this.
Sadly I don't have a timeline to work on the Android version yet, but still, I would like to have it merged to it reduces the effort to keep it up-to-date and then try to build the Android version of it.

cobarx:
@sidferreira Please resolve the merge conflicts and revert the white space changes.

sidferreira:
@cobarx OMG, took me 6 months to properly address it... Though, we now have the Android implementation.

Please give your input.

deepgosalia1:
Hey @sidferreira , I see there is some development regarding android and ios part for google ima sdk. I am willing to experiment with your repo, could you give a brief about what params do I need to pass to the video component for vast-tag URL?

sidferreira:
@hueniverse what could be done to have it merged?

hueniverse:
@sidferreira The first step is to rebase. Once that is done, we will need 3 people to review and approve the changes. It would also be helpful to expand the documentation a bit to give more context for what this feature does (some less technical background, maybe an example scenario - just English not code).

hueniverse:
Closing for now. If still interested in getting this in, please rebase and resubmit.

zuck:
Hi @sidferreira, great job on this side! Any chance of having this PR rebased in order to be finally merged? Thanks!

sidferreira:
@zuck will do it later today

PR #1778, opened 2019-10-01
added alternativeAudioSource prop to replace audio track if needed (i…

reime005:
I've added the possibility to use an alternative audio, like using a german audio track for a video with an english (or none) sound. 

The prop `selectedAudioTrack` and `audioTracks` only worked on Android, but not iOS. So I've implemented the alternative only for iOS. It's actually pretty simple and works like the `source` property.

Let me know what you think.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -418,7 +418,7 @@ static int const RCTVideoUnset = -1;
   return nil;
 }
 
-- (void)playerItemPrepareText:(AVAsset *)asset assetOptions:(NSDictionary * __nullable)assetOptions withCallback:(void(^)(AVPlayerItem *))handler
+- (void)playerItemPrepareText:(AVAsset *)asset assetOptions:(NSDictionary * __nullable)assetOptions withCallback:(void(^)(AVPlayerItem *))handler withAudio:(AVAsset *)audio
 {
   // AVPlayer can't airplay AVMutableCompositions
   _allowsExternalPlayback = NO;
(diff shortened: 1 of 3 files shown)


ios/Video/RCTVideo.m
```diff
@@ -438,13 +433,28 @@ - (void)playerItemPrepareText:(AVAsset *)asset assetOptions:(NSDictionary * __nu
...
   
-  AVAssetTrack *audioAsset = [asset tracksWithMediaType:AVMediaTypeAudio].firstObject;
-  AVMutableCompositionTrack *audioCompTrack = [mixComposition addMutableTrackWithMediaType:AVMediaTypeAudio preferredTrackID:kCMPersistentTrackID_Invalid];
-  [audioCompTrack insertTimeRange:CMTimeRangeMake(kCMTimeZero, videoAsset.timeRange.duration)
-                          ofTrack:audioAsset
-                           atTime:kCMTimeZero
-                            error:nil];
-  
+  // load alternative audio source, if provided
+  if (audio != nil) {
```
CHaNGeTe, line 437: Maybe I am half blind, but I don't see where this audio var comes from

reime005: oh yea :D sry, added that now

yasinugrl:
hi, @reime005 

Thanks your code. This is solve my problem on IOS part. But in Android part can ı add audio source like your prop(alternativeAudioSource)
Thanks.

alencrowdbotics:
This would be a great feature if implemented. I am going around the RCTVideo.m for checking why the other audio sources do not appear on iOS but appear on Android ExoPlayer.

PR #1779, opened 2019-10-02
Audio Poster issue fix

ghost:
React-native-video support audio only play. In case of audio play we can supply the poster and React-native-video will show the poster. But after the audio has started playing the poster is hidden and there is blank view. This PR is to show the audio poster even after the audio starts playing


react-native-video.podspec
```diff
@@ -9,8 +9,8 @@ Pod::Spec.new do |s|
   s.description    = package['description']
   s.license        = package['license']
   s.author         = package['author']
-  s.homepage       = 'https://github.com/brentvatne/react-native-video'
-  s.source       = { :git => "https://github.com/brentvatne/react-native-video.git", :tag => "#{s.version}" }
+  s.homepage       = 'https://github.com/learnyst/react-native-video'
+  s.source       = { :git => "https://github.com/learnyst/react-native-video.git", :tag => "#{s.version}" }
```
CHaNGeTe: please change this for the PR

CHaNGeTe (changes requested):

react-native-video.podspec
jensellfors: Should we not just change to https://github.com/react-native-community/react-native-video? Don't know why brentvatne is still there

jensellfors (changes requested):

react-native-video.podspec
sridhard: Yes we need to change...I will change it

sridhard:
@CHaNGeTe @cobarx @jenshandersson can you please review and merge the PR

jensellfors (approved):

react-native-video.podspec
bubah: Any idea when this fix will be complete?

benoitdion (approved):

react-native-video.podspec
bubah: Hi, Any idea when a new release  that includes this fix will be made available?

bubah: Hello, My group at work is using react-native-video in one of our apps and this issue is currently one of the blockers keeping us from releasing. Would you be able to provide an ETA on when the next version that includes this fix will be released?

PR #1780, opened 2019-10-02
Feature/android media session

matejpolak:
#### Describe the changes
Added mediaSession for Android ExoPlayer to let device know about currently used player. With this in place you can use some native features like pause/play using HW buttons on any compatible headset connected to the device, displaying playback buttons in PIP window (previous, play/pause, next), etc.. 

Currently using the simplest solution with usage of external ExoPlayer mediaSession extension library, which takes care of all needed mediaSession callback, and the session it self. To support more advanced features, full implementation may be needed. 

Since I'm not an Android expert, it's possible that some things, or additional metadata are missing. Please consider this as MVP which is currently implemented mainly to support PIP player controls. Feel free to suggest any change or pick deeper implementation if needed.

Also added `onExternalPauseToggled ` event prop, which notifies client about new play/pause state, so JS can perform needed actions based on this information. For example change UI state.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -63,7 +63,6 @@ import com.google.android.exoplayer2.upstream.DataSource;
 import com.google.android.exoplayer2.upstream.DefaultAllocator;
 import com.google.android.exoplayer2.upstream.DefaultBandwidthMeter;
 import com.google.android.exoplayer2.util.Util;
-import com.google.android.exoplayer2.ui.PlayerControlView;
 import com.google.android.exoplayer2.ext.mediasession.MediaSessionConnector;
 
 import java.net.CookieHandler;
@@ -658,7 +657,7 @@ class ReactExoplayerView extends FrameLayout implements
                  * external play/pause state change.
                  */
                 if (playWhenReady == isPaused) {
-                   eventEmitter.externalPauseTriggered(playWhenReady);
+                   eventEmitter.externalPauseToggled(playWhenReady);
                 }
                 break;
             case Player.STATE_ENDED:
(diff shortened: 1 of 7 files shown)


android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -62,6 +63,8 @@
 import com.google.android.exoplayer2.upstream.DefaultAllocator;
 import com.google.android.exoplayer2.upstream.DefaultBandwidthMeter;
 import com.google.android.exoplayer2.util.Util;
+import com.google.android.exoplayer2.ui.PlayerControlView;
```
CHaNGeTe: this import is duplicated

CHaNGeTe (changes requested):

android-exoplayer/src/main/java/com/brentvatne/exoplayer/VideoEventEmitter.java
```diff
@@ -193,6 +197,12 @@ void ready() {
         receiveEvent(EVENT_READY, null);
     }
 
+    void externalPauseTriggered(boolean isPlaying) {
```
CHaNGeTe: the name of the func should'nt be toggled as well?

CHaNGeTe (approved):

AurangzaibRamzan (approved):

rogerkerse:
Hi, what is the state with this pull request? 😊 Would be very grateful if it got merged. We are using picture in picture on Android through another library, but since current version of react-native-video doesn't have media session, there are no playback buttons :(

EyMaddis:
Just a heads up if this ever gets merged:
`onExternalPauseToggled` is not necessary, there is already the prop `onPlaybackRateChange` that should be able to do the same. I just stumbled onto this PR, I did not test it.

PR #1787, opened 2019-10-07
Fix/avplayer release

CHaNGeTe:
Release AVPlayer instance. Fixes some errors about: Decoder busy, etc.

PR #1799, opened 2019-10-18
fix: use AVPlayerLooper to repeat videos on iOS

leethree:
The original `repeat` implementation on iOS leaks memory for some reason and it will cause app to become slower and slower each time it loops. This PR replace the implementation with [AVPlayerLooper](https://developer.apple.com/documentation/avfoundation/avplayerlooper) which would solve the issue.

This PR is based on #1787.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Changelog
 
 ### next
-* Fix AVPlayer release
+* Fix - Release AVPlayer instance[#1787](https://github.com/react-native-community/react-native-video/pull/1787)
 
 ### Version 5.1.0-alpha1
 * Fixed Exoplayer doesn't work with mute=true (Android). [#1696](https://github.com/react-native-community/react-native-video/pull/1696)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### next
+* Fix AVPlayer release
+
 ### Version 5.1.0-alpha1
 * Fixed Exoplayer doesn't work with mute=true (Android). [#1696](https://github.com/react-native-community/react-native-video/pull/1696)
 * Added support for automaticallyWaitsToMinimizeStalling property (iOS) [#1723](https://github.com/react-native-community/react-native-video/pull/1723)
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -24,8 +24,9 @@ static int const RCTVideoUnset = -1;
 
 @implementation RCTVideo
 {
-  AVPlayer *_player;
+  AVQueuePlayer *_player;
   AVPlayerItem *_playerItem;
+  AVPlayerLooper *_playerLooper;
   NSDictionary *_source;
   BOOL _playerItemObserversSet;
   BOOL _playerBufferEmpty;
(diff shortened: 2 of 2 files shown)


FansPro:
Has this merge to master ? I have repeat video on ios, it get slower and slower..

adamkoch:
I was running into this memory leak and this PR seems to fix things. However, it seemed to cause another issue. In my app, I display up to 8 videos at the same time (all looping). After this patch, a random number of those would just render empty/blank. I'm not familiar with Objective C and iOS development but after commenting out this line it seems to have fixed it for me:
https://github.com/react-native-community/react-native-video/pull/1799/files#diff-42e8804735be64e84d6dd1fb103210a1R379

I'm guessing some sort of race condition with how _playerLooper is set but I couldn't work out exactly how or why.

leethree:
@adamkoch thanks for testing out! I think removing the line would cause a problem if `source` is changed for the component. But I haven't been able to figure out how to properly handle it. Would be great if someone more proficient in Objective C could help out 😀

jayasme:
@adamkoch  Hi, can you please print what `_playerItem` is before `_player = [AVQueuePlayer playerWithPlayerItem:_playerItem];`?
And also check `status` & `error` properties of `_playerLooper` after it has been initialized.

adamkoch:
@jayasme 

> can you please print what _playerItem is before _player = [AVQueuePlayer playerWithPlayerItem:_playerItem];

I just did a NSLog directly before this and it shows:
2019-12-20 16:28:37.827398+1100 MyApp-tvOS[13667:125228] _playerItem = <AVPlayerItem: 0x6000000b3f80, asset = <AVURLAsset: 0x60000035fac0, URL = file:////Users/adam/Library/Developer/CoreSimulator/Devices/F0C8BC28-3E5D-4F9A-9172-03357ACE393C/data/Containers/Data/Application/8C499507-2CEE-454B-982F-3B660831FC69/Library/Caches/test.mp4>>

When I load 8 videos on the screen and some show and some don't, they all print out a similar object (different values but valid object).

> And also check status & error properties of _playerLooper after it has been initialized.

I did a NSLog of success/status directly after _playerLooper is initialized (there's 2 locations in the file) but in both cases the values are always null. I'm not sure if maybe those values are populated asynchronously so if printing immediately after they are not yet populated?

jayasme:
> @jayasme
> 
> > can you please print what _playerItem is before _player = [AVQueuePlayer playerWithPlayerItem:_playerItem];
> 
> I just did a NSLog directly before this and it shows:
> 2019-12-20 16:28:37.827398+1100 MyApp-tvOS[13667:125228] _playerItem = <AVPlayerItem: 0x6000000b3f80, asset = <AVURLAsset: 0x60000035fac0, URL = file:////Users/adam/Library/Developer/CoreSimulator/Devices/F0C8BC28-3E5D-4F9A-9172-03357ACE393C/data/Containers/Data/Application/8C499507-2CEE-454B-982F-3B660831FC69/Library/Caches/test.mp4>>
> 
> When I load 8 videos on the screen and some show and some don't, they all print out a similar object (different values but valid object).
> 
> > And also check status & error properties of _playerLooper after it has been initialized.
> 
> I did a NSLog of success/status directly after _playerLooper is initialized (there's 2 locations in the file) but in both cases the values are always null. I'm not sure if maybe those values are populated asynchronously so if printing immediately after they are not yet populated?

status & error will not be set asynchronously at least by my understanding according to Apple's document, I don't know if you could try to make all players link the same one playerItem to see if can load all or part.

ninjz:
Any updates on the status of this?

pribeh:
Any reason why this wasn't merged? This resolves the issue for us.

adamkoch:
FWIW, this is still an issue for me.

hueniverse:
@leethree Can you please rebase so I can merge your fix?

Issue #1436, opened 2019-01-21
Confusing documentation for reportBandwidth


wpinaud:
The reportBandwidth prop:
https://github.com/react-native-community/react-native-video#reportbandwidth

```
false (default) - Generate onBandwidthUpdate events
true - Don't generate onBandwidthUpdate events
```

When reportBandwidth is set to *false* the plugin reports the event bandwidth.
It should be the opposite, or the prop be renamed "dontReportBandwidth"

cobarx:
Is the documentation wrong or does it actually report bandwidth when set to false? Not sure if updating the docs is sufficient.

set0gut1:
The actual behavior is generating events when reportBandwidth is true.
The implementation also seems to intend this behavior.

https://github.com/react-native-community/react-native-video/blob/b97663e09243836fd565e3def149b0bb94dfb82a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java#L221-L224

https://github.com/react-native-community/react-native-video/blob/b97663e09243836fd565e3def149b0bb94dfb82a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L913-L915

https://github.com/react-native-community/react-native-video/blob/b97663e09243836fd565e3def149b0bb94dfb82a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L251-L257

PR #1816, opened 2019-11-08
fix README about reportBandwidth

set0gut1:
In the README, the description of the reportBandwidth prop's value (true or false) is reversed.
Actually, events are generated when reportBandwidth is true.


set0gut1:
related issue #1436

benoitdion (approved):

PR #1862, opened 2020-01-03
[ios] Adaptive fullscreen in landscape by device orientation

r0b0t3d:
Currently, when setting `fullscreenOrientation` to `landscape`, the video will always display in `landscape right`.
This PR allows checking the device's orientation then set the right `UIInterfaceOrientation`.

### DIFF
--- a/ios/Video/RCTVideoPlayerViewController.m
+++ b/ios/Video/RCTVideoPlayerViewController.m
@@ -28,8 +28,7 @@
 
 - (UIInterfaceOrientation)preferredInterfaceOrientationForPresentation {
   if ([self.preferredOrientation.lowercaseString isEqualToString:@"landscape"]) {
-    UIDeviceOrientation currentOrientation = [[UIDevice currentDevice] orientation];
-    return currentOrientation == UIDeviceOrientationLandscapeRight ? UIInterfaceOrientationLandscapeLeft : UIInterfaceOrientationLandscapeRight;
+    return UIInterfaceOrientationLandscapeLeft | UIInterfaceOrientationLandscapeRight;
   }
   else if ([self.preferredOrientation.lowercaseString isEqualToString:@"portrait"]) {
     return UIInterfaceOrientationPortrait;
--- a/ios/Video/RCTVideoPlayerViewController.m
+++ b/ios/Video/RCTVideoPlayerViewController.m
@@ -28,7 +28,8 @@
 
 - (UIInterfaceOrientation)preferredInterfaceOrientationForPresentation {
   if ([self.preferredOrientation.lowercaseString isEqualToString:@"landscape"]) {
-    return UIInterfaceOrientationLandscapeRight;
+    UIDeviceOrientation currentOrientation = [[UIDevice currentDevice] orientation];
+    return currentOrientation == UIDeviceOrientationLandscapeRight ? UIInterfaceOrientationLandscapeLeft : UIInterfaceOrientationLandscapeRight;
   }
   else if ([self.preferredOrientation.lowercaseString isEqualToString:@"portrait"]) {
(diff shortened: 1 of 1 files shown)


sayem314:
Why is this ignored!

ios/Video/RCTVideoPlayerViewController.m
```diff
@@ -28,7 +28,8 @@ - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
 
 - (UIInterfaceOrientation)preferredInterfaceOrientationForPresentation {
   if ([self.preferredOrientation.lowercaseString isEqualToString:@"landscape"]) {
-    return UIInterfaceOrientationLandscapeRight;
+    UIDeviceOrientation currentOrientation = [[UIDevice currentDevice] orientation];
```
jensellfors: Not sure but I think you can just return `UIInterfaceOrientationLandscapeRight | UIInterfaceOrientationLandscapeLeft` no?

r0b0t3d: Yep, tested on simulator and working fine. I am going to update PR

r0b0t3d:
@jenshandersson I got error in ios 12.x

Simulator Screen Shot - iPhone 5s - 2020-06-06 at 20 43 21

The error cause by this line 
```
return UIInterfaceOrientationLandscapeLeft | UIInterfaceOrientationLandscapeRight;
```
Fixed by changing to
```
UIDeviceOrientation currentOrientation = [[UIDevice currentDevice] orientation];
return currentOrientation == UIDeviceOrientationLandscapeRight ?
        UIInterfaceOrientationLandscapeLeft : UIInterfaceOrientationLandscapeRight;
```
Do you have any idea?

jensellfors:
@r0b0t3d ah seems you can't return a mask for `preferredInterfaceOrientationForPresentation`... Let me push a fix for that, thanks

jensellfors:
Hum this was tricker than I thought. I suggest we remove `fullscreenAutorotate` prop. And if you set `fullscreenOrientation="landscape"`we force landscape ie setting `supportedInterfaceOrientations` to `UIInterfaceOrientationMaskLandscape`. Thoughts @r0b0t3d?

jensellfors:
@r0b0t3d I reverted this commit now. And using the snippet you sent won't work well when auto `fullscreenAutorotate` is true. There are a lot of permutations and edge cases we need to test...

r0b0t3d:
@jenshandersson Agreed on point that we could remove `fullscreenAutorotate`.
So with `fullscreenOrientation`
- `all`: rotate by device orientations
- `landscape`: rotate to landscape left or right
- `portrait`

`preferredInterfaceOrientationForPresentation` return the orientation when presenting the view controller. So if we want auto rotate to landscape left or right, we must set `supportedInterfaceOrientations` to `UIInterfaceOrientationMaskLandscape` and `shouldAutorotate` to YES.

But this approach could works only if application support all the orientations. If application only support portrait, it gave this error `Supported orientations has no common orientation with the application, and [RCTVideoPlayerViewController shouldAutorotate] is returning YES`

I am doing another approach by implementation method swizzling `application:supportedInterfaceOrientationsForWindow:`. So we can override it with `UIInterfaceOrientationMaskAll` and do the logic on `supportedInterfaceOrientations`.

What do you think?

jensellfors:
The more I'm thinking about it I feel we should just leave it to `all`... If you want lock or force an orientation, you can use https://github.com/wonday/react-native-orientation-locker (that's what I do). I feel this adds complexity ( and potential issues) to this project...

r0b0t3d:
That sound is good. Doing my approach will make the project more complexity. Let me try with `react-native-orientation-locker`.

PR #1863, opened 2020-01-03
Added `requestOrientationInFullscreen` property

incetarik:
Added `requestOrientationInFullscreen` property to control whether the activity should request landscape/portrait orientation natively, according to the `fullscreen` property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,7 @@ var styles = StyleSheet.create({
 * [fullscreen](#fullscreen)
 * [fullscreenAutorotate](#fullscreenautorotate)
 * [fullscreenOrientation](#fullscreenorientation)
+* [requestOrientationInFullscreen](#requestOrientationInFullscreen)
 * [headers](#headers)
 * [hideShutterView](#hideshutterview)
 * [id](#id)
@@ -641,6 +642,14 @@ Determines how to resize the video when the frame doesn't match the raw video di
 * **"stretch"** - Scale width and height independently, This may change the aspect ratio of the src.
 
 Platforms: Android ExoPlayer, Android MediaPlayer, iOS, Windows UWP
+
+#### requestOrientationInFullscreen
+Determines whether the activity should request landscape orientation or not when the fullscreen property is set to true.
+* **false (default)** - Don't change the activity orientation automatically
+* **true** - Change the activity orientation automatically when the fullscreen is set to true
+
+Platforms: Android
+
 
 #### selectedAudioTrack
 Configure which audio track, if any, is played.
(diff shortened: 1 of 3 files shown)


jensellfors:
Isn't this the same as `fullscreenOrientation` but for Android? Can we reuse the same prop?

incetarik:
> Isn't this the same as `fullscreenOrientation` but for Android? Can we reuse the same prop?
Hi, long time passed and I'm not sure whether this thing had been solved in `react-native-video`. The problem was with Android. Setting `fullscreen` property was not actually making it fullscreen, but just setting some window flags if I'm not wrong, which makes the status bar and navigation bar disappear but not rotating the screen (requesting landscape orientation). This request was just for adding this property to reduce the work needed by the code to make the video full screen on landscape

r0b0t3d:
@incetarik You could try #2073. With those changes, `fullscreenOrientation` can be applied for `ExoPlayer Android`.

incetarik:
@r0b0t3d As you can see that the PR is created pretty long ago, and if your PR contains the things I was tried to fix/add, then of course this PR should be closed in that case, if yours is working. My intention was also adding the actual full screen flags on the Android side. Please tell me whether your PR contains such changes and this PR is already covered there so this could be closed.

r0b0t3d:
@incetarik The solution in that PR is start another activity to show full screen video. It contains code to request for the orientation that you set via `fullscreenOrientation`.
```
String orientation = getIntent().getStringExtra(EXTRA_ORIENTATION);
         if ("landscape".equals(orientation)) {
             setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE);
         } else if ("portrait".equals(orientation)) {
             setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_SENSOR_PORTRAIT);
         }
```

incetarik:
Thank you for replying @r0b0t3d , if you show that video in **another** activity, isn't it causing another navigation event? I'm thinking about whether this could also affect applications that use native navigation libraries.

r0b0t3d:
I was using `react-navigation/native` along with this PR and still not see the problem then. You could give it a try.

incetarik:
I think about `react-native-navigation` -like libraries which cause developers to change their boot/startup code. Because the layout will be based on whatever the library will provide so pushing another activity might cause a corrupted navigation data for them.

r0b0t3d:
Not sure that I got you. But in fullscreen video activity, it just show default controls from ExoPlayer. And you could customize such controls but it will be in native android.

PR #1883, opened 2020-01-24
Added textureViewOpaque prop

lukefanning:
This adds a prop for controlling the opaque setting for a TextureView surface view. It defaults to true which maintains current behaviour of the module. Setting the prop to false is very useful for opacity changes on the <Video /> component. If you decrease opacity on a <Video /> component leaving this on true means that you start to see a black background appear. Setting it to false eliminates the black background and the video can be faded out in a much more pleasing way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### Version 5.1.0-alpha2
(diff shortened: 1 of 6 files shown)


Christian-Athom:
I have tried and tested this PR, but the problem with the black background if you change the opacity of the video still remains. 

Are there any additional steps to make this work?

Issue #1871, opened 2020-01-14
Add React Native for Windows vNext support


stmoy:
# Feature Request

React Native for Windows enables React Native apps to target Windows via UWP. React Native for Windows is under active development by Microsoft. Most recently, the platform has been updated to leverage high-performance C++ (known as "RNW vNext"). Now that the RNW vNext supports native modules as a platform, we have identified several key native modules to add Windows support for (including react-native-video) to help validate that the platform has the right set of capabilities and to bootstrap the community of modules that target Windows.

Microsoft devs are signed up to help build out this implementation, but we wanted to track this work here as well to help provide some transparency that we're doing this work and hoping to contribute back to the repo.

(Note that this is also a topic of conversation on #1628)

## Why it is needed

Microsoft devs are signed up to help build out this implementation, but we wanted to track this work here as well to help provide some transparency that we're doing this work and hoping to contribute back to the repo.

## Possible implementation

This implementation will use the updated React Native for Windows vNext native module infrastructure, documented here: https://github.com/microsoft/react-native-windows/blob/master/vnext/docs/NativeModules.md

On the react-native-windows GitHub, we're tracking this work with the following issue: 
https://github.com/microsoft/react-native-windows/issues/2095

PR #1893, opened 2020-01-31
New project for supporting Video for React Native Windows Cpp/WinRT

ddalp:
This change adds a library project to support video for React Native Windows vNext Cpp/WinRT.
  - The project lives under windows\vNext folder for now.
  - Updated the readme.md to include instructions for building with RNW to support Video.
  - We will look into auto linking and clean-up existing windows support (CS and WPF) projects soon.

Since this is my first time creating PR for this community, please reply if I missed anything, thanks!

Isssue #1871


windows/vNext/ReactNativeVideoCPP/ReactVideoView.h
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
benoitdion: are you intending this file to be licensed separately from the rest of the project?

ddalp: Removed  the copy right text, thanks for the catch.

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
```diff
@@ -0,0 +1,255 @@
...
+  SetMediaPlayer(m_player);
+  m_uiDispatcher = CoreWindow::GetForCurrentThread().Dispatcher();
+  auto mediaOpenedToken = m_player.MediaOpened([ref = get_weak()](auto const &sender, auto const &args) {
+    auto self = ref.get();
+    if (self) {
+      self->OnMediaOpened(sender, args);
+    }
+  });
+
+  auto mediaFailedToken = m_player.MediaFailed([ref = get_weak()](auto const &sender, auto const &args) {
```
stecrain: Do these events get unregistered properly? Can the auto_revoke pattern be used instead?

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
```diff
@@ -0,0 +1,255 @@
...
+    m_player.IsMuted(m_isMuted);
+  }
+}
+
+void ReactVideoView::Set_Controls(bool useControls) {
+  m_useControls = useControls;
+  AreTransportControlsEnabled(m_useControls);
+}
+
+void ReactVideoView::Set_FullScreen(bool fullScreen) {
```
stecrain: When you set the video to fullscreen can you still receive keydown events from the <Video /> in js?
For my private (and less complete :)) video viewmanager I found that I had to handle that specifically as the MediaPlayerElement gets moved into the FullScreenMediaRoot tree, and is no longer a parent of the react root view.
Maybe something else takes care of this or I was missing something, but figured it would be good to ask :D

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
```diff
@@ -0,0 +1,255 @@
...
+using namespace Windows::UI::Xaml::Controls;
+using namespace Windows::UI::Core;
+using namespace Windows::Media::Core;
+using namespace Windows::Media::Playback;
+
+namespace winrt::ReactNativeVideoCPP::implementation {
+ReactVideoView::ReactVideoView(winrt::Microsoft::ReactNative::IReactContext const &reactContext)
+    : m_reactContext(reactContext) {
+  // always create and set the player here instead of depending on auto-create logic
+  // in the MediaPlayerElement (only when auto play is on or URI is set)
```
stecrain: Is this comment supposed to mean that there are cases when the default create logic is used? Or is it suposed to indicate that the mediaplayer is not created until the source is set?
Also I don't see autoplay handled in the updateProperties

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
```diff
@@ -0,0 +1,255 @@
...
+  AreTransportControlsEnabled(m_useControls);
+}
+
+void ReactVideoView::Set_FullScreen(bool fullScreen) {
+  if (m_fullScreen != fullScreen) {
+    m_fullScreen = fullScreen;
+    IsFullWindow(m_fullScreen);
+    auto applicationView = winrt::Windows::UI::ViewManagement::ApplicationView::GetForCurrentView();
+    bool isFullScreen = applicationView.IsFullScreenMode();
+    if (m_fullScreen) {
```
stecrain: Not a big deal to me (my apps always run in FullScreenMode) but I wonder if people expect FullWindow for the video player and FullScreen for the app to be the same thing.

One thing that could get wonky pretty fast, what if I have an app and I already entered FullScreenMode then toggle the FullScreen prop on the video. I believe that this code would cause the app to exit FullScreenMode when the video is no longer full window.
It might be better to have a separate native module that lets people set FullScreenMode independently.

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
```diff
@@ -0,0 +1,255 @@
...
+void ReactVideoView::Set_Position(double position) {
+  m_position = position;
+  if (m_player != nullptr) {
+    std::chrono::seconds duration(static_cast<int>(m_position));
+    m_player.PlaybackSession().Position(duration);
+  }
+}
+
+bool ReactVideoView::IsPlaying(MediaPlaybackState currentState) {
+  if (currentState == MediaPlaybackState::Buffering || currentState == MediaPlaybackState::Opening ||
```
kaiguo: Maybe just 
``` C++
return currentState == MediaPlaybackState::Buffering
|| currentState == MediaPlaybackState::Opening
|| currentState == MediaPlaybackState::Playing;
```

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
ddalp: It seems working fine, but I am going to apply the auto-revoke pattern just to be safe.

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
ddalp: Inside XAML, it has logic to not create underline MediaPlayerElement object if URI is not set yet. For simplicity, I just create the object upfront so we can use it later when properties are set, without needing to cache all properties and wait for URI set.

Also in regarding to autoPlay, I am going to add logic to enable autoplay when "paused" is not set or set to false, to be consistent with other platforms.

benoitdion: relevant: https://github.com/react-native-community/react-native-video/pull/1916

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
ddalp: We need to fix this in RNW code by specializing for MediaPlayerElement. This also impacts touch events too, we need to do similar think as we have done for flyout and popup. I will create an issue in RNW and fix it there.

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
ddalp: Will make fullscreen == fullwindow for now.

windows/vNext/ReactNativeVideoCPP/ReactVideoView.cpp
ddalp: Created RNW issue https://github.com/microsoft/react-native-windows/issues/4157.

windows/ReactNativeVideoCPP/ReactNativeVideoCPP.vcxproj
```diff
@@ -0,0 +1,157 @@
...
+  <PropertyGroup Condition="'$(Configuration)'=='Release'" Label="Configuration">
+    <UseDebugLibraries>false</UseDebugLibraries>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <LinkIncremental>false</LinkIncremental>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <ImportGroup Label="ExtensionSettings">
+  </ImportGroup>
+  <ImportGroup Label="Shared">
+    <Import Project="..\..\..\react-native-windows\Microsoft.ReactNative.Cxx\Microsoft.ReactNative.Cxx.vcxitems" Label="Shared" />
```
benoitdion, line 67: is it safe to hardcode `..\..\..`?

ddalp: We probably can make it better, once we have a story for auto linking between RNW and community modules. For now, it is ok for current RNW setup.

PR #1897, opened 2020-02-04
Fix audio focus

michaeltintiuc:
Implement audio focus as per android docs:
https://developer.android.com/guide/topics/media-apps/audio-focus
https://medium.com/androiddevelopers/audio-focus-3-cdc09da9c122

`AUDIOFOCUS_LOSS` should abandon focus and not try resuming audio, this is done with `AUDIOFOCUS_LOSS_TRANSIENT`

This fixes at least:
- Audio not being paused after focus being taken by some voip applications
- Content resuming and pausing instantly sporadically (some race condition perhaps) when activity was resumed from background.

Since this is triggering events, my proposal is to trigger the same event as for `AUDIO_FOCUS_LOSS` and let the developer choose to either use stop or pause, yet we explicitly abandon focus as per the docs.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -581,6 +581,10 @@ class ReactExoplayerView extends FrameLayout implements
         switch (focusChange) {
             case AudioManager.AUDIOFOCUS_LOSS:
                 eventEmitter.audioFocusChanged(false);
+                audioManager.abandonAudioFocus(this);
+                break;
+            case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:
+                eventEmitter.audioFocusChanged(false);
                 break;
             case AudioManager.AUDIOFOCUS_GAIN:
                 eventEmitter.audioFocusChanged(true);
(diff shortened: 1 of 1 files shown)


android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -580,6 +580,10 @@ private void clearResumePosition() {
     public void onAudioFocusChange(int focusChange) {
         switch (focusChange) {
             case AudioManager.AUDIOFOCUS_LOSS:
+                eventEmitter.audioFocusChanged(false);
+                audioManager.abandonAudioFocus(this);
```
benoitdion, line 585: If we're going to abandon focus, should we pause automatically? still giving the user the option to stop?

michaeltintiuc: This sounds fine, but my idea was to allow the dev to have full control of order of operations, but could be easily done

michaeltintiuc: Come to think of it, you're right, this is the right way of _enforcing_ good behavior and will require no changes from the devs. Will add it ASAP,t hanks for the feedback!

benoitdion (approved):

benoitdion:
These changes are available in `5.1.0-alpha3`

PR #1916, opened 2020-02-19
Android Exoplayer - Resolved the play and pause issue with the full-screen functionality.

IbrahimSulai:
I have added the full-screen functionality in the Android Exo-player. That had an issue with "isPaused" property.

I have resolved the same in the full-screen functionality.

It is related to the Issue and PR's:
https://github.com/react-native-community/react-native-video/issues/1895

https://github.com/react-native-community/react-native-video/pull/1730


splend-rajesh:
Since I've upgraded to the alpha version, the props "paused" doesn't work anymore.

benoitdion (approved):

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -296,6 +297,27 @@ public void onClick(View v) {
             }
         });
 
+        //Handling the playButton click event
+        ImageButton playButton = playerControlView.findViewById(R.id.exo_play);
```
benoitdion, line 301: were these buttons not working before?

IbrahimSulai: @benoitdion - It's working before, but it's handled by the player control view by default. But now, I have handled the actions of play and pause of my own to update the "isPause" property.

benoitdion: does this means you're overriding a built-in click listener with the one below or are the player control managing that differently?

IbrahimSulai: Yes... I am overriding a built-in click listener...But with the same default logic with additional handling of "ispaused" property.

`isPaused = paused;
if (player != null) {
            if (!paused) {
                startPlayback();
            } else {
                pausePlayback();
            }
        }`


In default, it will call `startPlayback()` and `pausePlayback()` on play and pause correspondingly

PR #1931, opened 2020-02-26
[fix] dont set reactViewController view frame to whole screen

marlonandrade:
Only set the view frame when video is being presented as fullScreen.

The component, when not displayed fullScreen, shouldn't assume the parent React View is on the whole screen.

This fixes the case on any time you have a react view that displays the video, but its being rendered with controls but not as a full screen (ie. easily reproducible when you have a tabbar for instance - the reactViewController.view frame shouldn't be the whole screen bounds).

#### Update the documentation
🆗  No changes needed.

#### Update the changelog
✅ Done

Notice the bottom of the contentSize on the scroll view.

| Before | After |
| -- | -- |
| before | after |

🆗 

#### Describe the changes
✅ Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### Version 5.1.0-alpha6
+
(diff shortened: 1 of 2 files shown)


benoitdion:
Thanks for the contribution @marlonandrade!

PR #1932, opened 2020-02-27
KVO Crash Remove PlayerViewController Observers on vc dismiss

samermurad:
The KVO crash still happens under certain circumstances, for example when having a view that still displays the video component after finishing the video.

Currently the KVO of the PlayerViewController are only being removed on removeFromSuperview.
Which problematic because the `_playerViewController` instance is being set to `nil` in the  `videoPlayerViewControllerDidDismiss`
which deallocates the the instances because it is no longer retained, so if the `removeFromSuperview` method is not called directly after that, you will get the KVO crash on the two values "frame" and "readyForDisplay" because they are both registered on the `_playerViewController` and `_playerViewController. contentOverlayView`.
this happen in situations where you are displaying the video object after the video is finished as well, for example in a listView or an article, where the user will still be on the page and the Video element is still rendered.
Video is not being played, just displayed and the `_playerViewController` was deallocated, but `removeFromSuperview` was yet to be called, the view is still visible.

Longer story short, the `_playerViewController` KVO listeners must be removed before nullifying the  `_playerViewController` object

Thanks for opening a PR!

#### Update the documentation

#### Update the changelog

#### Describe the changes


jensellfors (approved):
Never had a crash for this but it all sounds correct, nice one!

ios/Video/RCTVideo.m
```diff
@@ -530,7 +531,6 @@ - (void)playerItemForSource:(NSDictionary *)source withCallback:(void(^)(AVPlaye
 - (void)playerItemForSourceUsingCache:(NSString *)uri assetOptions:(NSDictionary *)options withCallback:(void(^)(AVPlayerItem *))handler {
     NSURL *url = [NSURL URLWithString:uri];
     [_videoCache getItemForUri:uri withCallback:^(RCTVideoCacheStatus videoCacheStatus, AVAsset * _Nullable cachedAsset) {
-        switch (videoCacheStatus) {
```
jensellfors: Oh I missed this, this must be a typo no? You can't remove the switch case like that? @samermurad

jensellfors (changes requested):

ios/Video/RCTVideo.m
samermurad: @jenshandersson oh wow yeah this is definitely a typo, I'll Fix it in a sec

samermurad: @jenshandersson Fixed

samermurad:
@jenshandersson Regarding the reproduction of the crash, I think it's a special situation where you still keep the Video Thumbnail displayed (in a list for example) After dismissing the video player The Idea is to allow the `videoPlayerViewControllerDidDismiss` to be called but still keep the view thumbnail view on screen, the crash doesn't happen instantly it takes about 1.5s sec to happen (I guess it depends on the device).
But this commit fixes it anyways so not sure the long explanation is necessary, although is probably helpful for future reference

jensellfors (approved):
@samermurad thanks. Can you also add a line to the Changelog please, then I'm happy to merge

samermurad:
@jenshandersson Sorry I totally missed that message, will do so shortly

samermurad:
@jefersondaniel (I'll fix the merge conflict as well) but under what version name should I add the changelog?

PR #1941, opened 2020-03-06
Fixing ID3 Frame Error When Receiving EventMessage in TimedMetadata

raymallick:
#### Describe the changes
On certain video streams, it is possible to receive an EventMessage instead of an ID3Frame. Thus the line `Id3Frame frame = (Id3Frame) metadata.get(i);` would cause a fatal Java Error. We are now funneling the logic into two branches when receiving an ID3 Frame or an EventMessage.

#### Update the changelog


raymallick:
@cobarx I can remove the `else-if` check if we do not want to map the EventMessage and we only care about ID3Frames.

cobarx:
@raymallick I like the approach you've taken, emitting as much info as possible is ideal.

cobarx (approved):
tested and verified working

PR #1943, opened 2020-03-10
Update (NOMINMAX and Disable C4996) properties in the windows proj 

namrog84:
#### Describe the changes
This PR disables a min/max macro collision with windows api.  This collision occurs if you are using a newer version of RNW then RNV currently uses, and can also just be generally a problematic area. A request for changes on RNW side has also occurred but this should have been here since the beginning.

Also disables deprecation warnings, because many projects often build warnings as errors and since RNW is rapidly changing its API. It is deprecating things at higher than normal rate at the moment and makes it painful to use newer version of RNW.  Alternatively, RNV would need to rev it's RNW version very frequently.

This PR originates from having compiler warnings/errors from building with a newer version of RNW.  

Context: We currently use/testing ReactNativeVideo for all 3 platforms (Android/iOS/Windows)
Disclaimer: MS Xbox Employee

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### Version 5.1.0-alpha8
+
+- Fixing ID3 Frame Error When Receiving EventMessage in TimedMetadata [#2116](https://github.com/react-native-community/react-native-video/pull/2116)
+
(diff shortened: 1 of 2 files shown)


benoitdion:
@namrog84 do you mind rebasing to fix the CHANGELOG conflict?

PR #2019, opened 2020-05-26
Support preventsDisplaySleepDuringVideoPlayback

### LINKED ISSUE #1869: Configurable prop to disable "prevent display from sleep"
# Feature Request

Configurable prop to disable "prevent display from sleep"

## Why it is needed

For background video use, where the video only acts as nicer background. 

## Possible implementation

There should already the code included to disable sleep, because this is not the default behavior.

### Code sample

sleep
Controls whether the sleep timer from system settings is enabled

false(default) - Don't sleep
true - sleep as usual
Platforms: all

anderslemke:
- [x] iOS
- [x] Android
- [x] Documentation
- [x] Changelog
- [x] Test description
- [x] General description

## Description

This flag allows the device display to sleep while playing a video. Useful if video is not primary content in the app.

```jsx
<Video
  ...
  preventsDisplaySleepDuringVideoPlayback={ false }
  ...
/>
```

## Test description
To test this
1. instantiate with `preventsDisplaySleepDuringVideoPlayback={false}`
2. Run on device (on iOS it should not be attached to Xcode while running)
3. Wait and let the display sleep.

Closes #1869


jensellfors:
This looks good to me @anderslemke, do you mind adding Exoplayer support as well? For 6.0.0 we're aiming to deprecate MediaPlayer.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -605,7 +606,7 @@ private void startPlayback() {
             initializePlayer();
         }
         if (!disableFocus) {
-            setKeepScreenOn(true);
+            setKeepScreenOn(true && preventsDisplaySleepDuringVideoPlayback);
```
jensellfors: You can remove this `true` here as it doesn't do anything :)

jensellfors:
@anderslemke can you check my comment and fix conflicts.

Closes https://github.com/react-native-community/react-native-video/issues/2042

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -139,6 +139,7 @@
     private Dynamic textTrackValue;
     private ReadableArray textTracks;
     private boolean disableFocus;
+    private boolean preventsDisplaySleepDuringVideoPlayback;
```
jensellfors: You need to set this to true, as you have specified it should be true by default in README. Good to keep it consistent with iOS.

jensellfors (changes requested):
Fix default value and the conflicts, then I'm happy to merge!

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
anderslemke:
```suggestion
            setKeepScreenOn(preventsDisplaySleepDuringVideoPlayback);
```

anderslemke: Done ✅

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
anderslemke: Is there somewhere else than https://github.com/react-native-community/react-native-video/blob/1e22beeaa150e800ab7ce9c56e56a0a10409f327/ios/Video/RCTVideo.m#L110 this should be set?

jensellfors: You need to set it both for ExoPlayer and MediaPlayer, currently it's only set for MediaPlayer: https://github.com/react-native-community/react-native-video/pull/2019/files#diff-e971272a051c4ba241583904cf81ff0aR126.
So please set it to true where this commit is :)

jensellfors:
```suggestion
    private boolean preventsDisplaySleepDuringVideoPlayback = true;
```

anderslemke:
@jenshandersson I've pulled in your suggestion (thank you for that) and fixed the conflicts.

jensellfors (approved):

jensellfors:
Nice work guys! 🙌

Issue #1741, opened 2019-09-05
Where is version 5.0.1?


fylock:
# Question

jonathangreco:
# answer
Not released.

CHaNGeTe:
no it is released

PR #2071, opened 2020-07-02
[iOS] Move callback to method

sidferreira:
This PR is a really small subset of #1741 
The reason I suggest moving the callback to its own method is both to:
a) better readability
b) easier to maintain #1741 until it is ready for merge

The only change it has over the original code that was in the callback is the `if` around the player initializer.


hueniverse:
This is no longer compatible due to the iOS Swift rewrite.

PR #2075, opened 2020-07-07
Fix audioVolume value when muted is false

anthinkingcoder:
Thanks for opening a PR!

#### Update the documentation

#### Update the changelog

#### Describe the changes

when I set volume = 0.1 or other less than 1 value, it doesn't work, because muted be set `1.f` when setMutedModifier(false),
so we need to use `audioVolume` instead of `1.f`
```js
         <Video
              source={{ uri: videoPlayUrl }}
              muted={false}
              volume={0.1}
              resizeMode="contain"
            />
```
```java    
//old
audioVolume = muted ? 0.f : 1.f;
//new
audioVolume = muted ? 0.f : audioVolume;
```


hueniverse:
This change doesn't make sense if you go from muted to not muted. When you mute the volume is set to 0 and then when you set mute to false this change will leave it at 0... I'm not a Java dev unless I'm missing something.

PR #2125, opened 2020-08-24
Ability to set FontSize for textTrack in Android

msaqlain:
Pass `fontSizeTrack` prop for setting fontSize of textTracks in Android.

```
<Video
            selectedTextTrack={{ type: 'language', value: 'en' }}
            textTracks={textTracks}
            source={{
              uri: this.props.route.params.file,
              type: 'mp4',
            }}
            controls={true}
            fullscreen={true}
            fontSizeTrack={16}
          />
```


milad755:
how can change color textTracks?

Issue #2045, opened 2020-06-10
ReactVideoView does not clean up correctly


ocalvo:
# Bug

ReactVideoView manages its own instance of MediaPlayer and calls SetMediaPlayer() to initialize MediaPlayerElement.
However, it never calls SetMediaPlayer(null) and does not call Close on the MediaPlayer instance. This makes a lingering active MediaPlayer instance that sometimes makes the entire app crash as Xaml cannot handle an active MediaPlayer while the Xaml Element is no longer in the Xaml tree.

## Platform

Which player are you experiencing the problem on:
* Windows UWP

## Environment info

React native info output:

```bash
 // paste it here
```

Library version: x.x.x

## Steps To Reproduce

This has been reported in the wild and is a heavy hitter for apps that use Media.

## Expected behaviour

No crash after Media have exited the tree.

## Reproducible sample code
N/A

## Video sample
If possible, include a link to the video that has the problem that can be streamed or downloaded from.
N/A

jensellfors:
Sounds like you know how to fix it ;) Mind submitting a PR?

PR #2131, opened 2020-08-26
#2045-Fix-ReactVideoView-does-not-clean-up-correctly

SvenEssentry:
Fixed a bug where the `MediaPlayer` kept streaming audio even after dispose.
The issue can be reproduced in previous versions by starting a video stream an then move to another view.

Here is a link to the original issue: [#2045](https://github.com/react-native-community/react-native-video/issues/2045)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Version 4.4.4
 * Handle racing conditions when props are setted on exoplayer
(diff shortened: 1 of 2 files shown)


ebarahona:
Can this be merged

SvenEssentry:
We are using this change in production, everything behaves as expected and would appreciate this to be reviewed and merged too.

PR #2156, opened 2020-09-21
Fix iOS Volume not returning to normal volumes when using ducking

dorthwein:
Activates and Deactivates the audio session based on paused being set to true/false.  This is needed to restore ducking volume to normal normal level.


dandre-hound:
This PR has worked well, but it should also take into account a video's initial muted status and should set the audio session properties when toggling mute status.

Issue #1930, opened 2020-02-25
Rate is reset to 1.0 after play/pause on iOS


joeyscarim:
## Bug
If I set the rate of a video to 2.0 using the rate prop it works as expected but after hitting pause -> play on the video it reverts to a rate of 1.0.

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info
React native info output:
```bash
System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 64.47 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.13.1 - /usr/local/bin/node
    Yarn: 1.19.2 - /usr/local/bin/yarn
    npm: 6.13.6 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 24, 25, 26, 27, 28
      Build Tools: 19.1.0, 23.0.1, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.1, 27.0.2, 27.0.3, 28.0.0, 28.0.2, 28.0.3
      System Images: android-24 | Google APIs Intel x86 Atom, android-24 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.6010548
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.5 => 0.60.5 
  npmGlobalPackages:
    create-react-native-app: 1.0.0
    react-native-contacts-wrapper: 0.2.4
    react-native-git-upgrade: 0.2.7
    react-native: 0.61.2
```

Library version: `5.1.0-alpha2`

benoitdion:
thanks for the report @joeyscarim! If you have the bandwidth, we welcome bug fixes PRs.

PR #2167, opened 2020-09-28
Fix: iOS rate is reset to 1.0 after play/pause #1930

redspear:
#### Update the documentation
No changes to documentation needed.

#### Update the changelog
Will do.

If the rate prop is set it will apply as expected. However, when the video is paused & played the rate will be set to 1.0 - the custom rate is not re-applied. This bug is present on any fresh build.

#### Describe the changes
When playback resumes we check the new playbackRate against the props rate. If they're different we set the playbackRate to match the props rate.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -718,7 +718,10 @@ static int const RCTVideoUnset = -1;
     }
   } else if (object == _player) {
     if([keyPath isEqualToString:playbackRate]) {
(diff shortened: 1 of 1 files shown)


ignaciosantise:
Thanks @redspear for this fix! I'm using it in a production app and it works great. Do you know when is this going to be merged?

redspear:
I'm glad to hear it helped you. I don't know when/if it'll be merged.

Issue #2177, opened 2020-10-07
ios14 showing floating pictureInPicture by default


anshul-kai:
# Bug

## Platform
* iOS

## Environment info
```bash
System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 74.54 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.15.0 - ~/.nvm/versions/node/v11.15.0/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.7.0 - ~/.nvm/versions/node/v11.15.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 14.0, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    Android SDK:
      API Levels: 23, 25, 26, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.0, 29.0.2
      System Images: android-23 | Intel x86 Atom, android-25 | Google Play Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
    Xcode: 12.0.1/12A7300 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.11.0 => 16.11.0
    react-native: 0.61.5 => 0.61.5
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

Library version: Latest master branch

## Steps To Reproduce
Play a local video file using code similar to below. Try to put the app in background and the video file floats around.
```
             <Video
                source={{ uri: `${FileSystem.BACKGROUND_VIDEO_DIR}/${filename}` }}
                rate={rate}
                volume={0}
                muted
                paused={paused}
                onLoadStart={() => {}}
                onLoad={() => {}}
                onEnd={() => {}}
                repeat
                playInBackground={false}
                playWhenInactive
                resizeMode='cover'
                style={styles.bg}
              />
```

flexgrip:
Ha. I wish I had this bug. On my published apps, they try to play the video in PiP but then quickly disappear.

[...]

PR #2191, opened 2020-10-26
fix: Only allow Picture in picture if _pictureInPicture is set to true

christianbach:
**Provide an example of how to test the change**
Run example project on iOS 14 and picture in picture is not shown by default.

**Describe the changes**
pictureInPicture is default **false** but since iOS 14 picture in picture is shown by default. This change sets `allowsPictureInPicturePlayback` to the same value as pictureInPicture.

Issue: https://github.com/react-native-video/react-native-video/issues/2177

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -156,7 +156,9 @@ static int const RCTVideoUnset = -1;
   viewController.showsPlaybackControls = YES;
   viewController.rctDelegate = self;
   viewController.preferredOrientation = _fullscreenOrientation;
-  
+  if (@available(iOS 9.0, *)) {
+	viewController.allowsPictureInPicturePlayback = _pictureInPicture;
+  }
   viewController.view.frame = self.bounds;
   viewController.player = player;
   return viewController;


edwardlai3582:
does this work on ipad?

christianbach:
> does this work on ipad?

It should work, can't remember now if I tested it on a real iPad or not... But it should work.

https://developer.apple.com/documentation/avkit/avplayerviewcontroller/1615821-allowspictureinpictureplayback?language=objc

edwardlai3582:
> > does this work on ipad?
> 
> It should work, can't remember now if I tested it on a real iPad or not... But it should work.
> 
> https://developer.apple.com/documentation/avkit/avplayerviewcontroller/1615821-allowspictureinpictureplayback?language=objc

Thank you for the prompt reply.
I guess what i should ask is does it fix #2177 ?
I tried your branch unfortunately the pip still showed up on iPad mini 14.2 (actual device)
The quick fix I've found so far is to keep the video width narrower than the device width (https://developer.apple.com/forums/thread/123499)
feel this should be able to be fixed on the native though

christianbach:
> > > does this work on ipad?
> > 
> > 
> > It should work, can't remember now if I tested it on a real iPad or not... But it should work.
> > https://developer.apple.com/documentation/avkit/avplayerviewcontroller/1615821-allowspictureinpictureplayback?language=objc
> 
> Thank you for the prompt reply.
> I guess what i should ask is does it fix #2177 ?
> I tried your branch unfortunately the pip still showed up on iPad mini 14.2 (actual device)
> The quick fix I've found so far is to keep the video width narrower than the device width (https://developer.apple.com/forums/thread/123499)
> feel this should be able to be fixed on the native though

I guessing it doesn't work the same on iPadOS. I have tested it successfully on iPhones running iOS 14.

iamchrispham:
> > > does this work on ipad?
> > 
> > 
> > It should work, can't remember now if I tested it on a real iPad or not... But it should work.
> > https://developer.apple.com/documentation/avkit/avplayerviewcontroller/1615821-allowspictureinpictureplayback?language=objc
> 
> Thank you for the prompt reply.
> I guess what i should ask is does it fix #2177 ?
> I tried your branch unfortunately the pip still showed up on iPad mini 14.2 (actual device)
> The quick fix I've found so far is to keep the video width narrower than the device width (https://developer.apple.com/forums/thread/123499)
> feel this should be able to be fixed on the native though

Was facing a very similar issue since iOS 14.  Keeping the video's width less than device's width does the trick.

nikolaigeorgie:
Sadly it doesn't work in the background because of this line of code in `RTCVideo.m`

```
- (void)applicationDidEnterBackground:(NSNotification *)notification
{
  if (_playInBackground) {
    // Needed to play sound in background. See https://developer.apple.com/library/ios/qa/qa1668/_index.html
    [_playerLayer setPlayer:nil];
    [_playerViewController setPlayer:nil];
  }
}
```

I tested without setting the player to null in the background.. works great... and the audio works great as well..

joebernard:
This PR is over 4 months old. Can we please get it reviewed? It addresses a major issue with the player.

zchwyng:
This didn't work for me on iPhone 8 Simulator. Picture in picture were still showing up while using allowsPictureInPicturePlayback={false}

zchwyng:
Is this working for anyone? Otherwise I will close this PR.

tombaranowicz:
This is unfortunately not working, at least for me, so I created a new PR https://github.com/tombaranowicz/react-native-video/pull/1 where I simply block pipController init if pictureInPicture is set to false. That was the only way I was able to fix this issue permanently on all devices I was testing.

Also, starting with iOS 14.2 there is a new flag `canStartPictureInPictureAutomaticallyFromInline` which for some unknown reason also does not prevent player from going into PiP for 1-2 seconds, the only solution for me was to simply not init pipController at all https://developer.apple.com/documentation/avkit/avplayerviewcontroller/3689455-canstartpictureinpictureautomati

joebernard:
Not stale, we still need this one fixed.

hueniverse:
Too many people reported this fix not working for them. Looking into @tombaranowicz solution instead.

PR #2288, opened 2021-02-15
Update ReactNativeVideoCPP.vcxproj

asklar:
This PR adds a no-op build step for Windows that is required to be defined on some versions of React Native for Windows


asklar:
CC @benoitdion @brentvatne could we merge this?

asklar:
thank youuu

PR #2296, opened 2021-02-24
Reference UIManager directly, not via NativeModules

esilverberg:
Fixes initialization bug where `NativeModules.UIManager` has not yet been initialized before `render` is called and constants are accessed.

While debugging today, I was hitting a crash caused by the failure of `UIManager` to have fully intialize various propeties, including the constant values, of `RCTVideo`, which meant that this call:

```
    const RCTVideoInstance = this.getViewManagerConfig('RCTVideo');
```

would return null, and this call would crash:

```RCTVideoInstance.Constants.ScaleToFill;```

By referring to `UIManager` instead of `NativeModules.UIManager`, we seem to force initialization of the constants in time for their access within the `render` method of `Video.js`

This project also appears to be the only [one on the internet](https://www.google.com/search?q=%22NativeModules.UIManager%22&oq=%22NativeModules.UIManager%22) that accesses UIManager via NativeModules.

This is related to: https://github.com/react-native-video/react-native-video/issues/2295

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { StyleSheet, requireNativeComponent, NativeModules, View, ViewPropTypes, Image, Platform, findNodeHandle } from 'react-native';
+import { StyleSheet, requireNativeComponent, NativeModules, View, ViewPropTypes, Image, Platform, findNodeHandle, UIManager } from 'react-native';
 import resolveAssetSource from 'react-native/Libraries/Image/resolveAssetSource';
 import TextTrackType from './TextTrackType';
 import FilterType from './FilterType';
@@ -254,10 +254,10 @@ export default class Video extends Component {
     }
   }
   getViewManagerConfig = viewManagerName => {
-    if (!NativeModules.UIManager.getViewManagerConfig) {
-      return NativeModules.UIManager[viewManagerName];
+    if (!UIManager.getViewManagerConfig) {
+      return UIManager[viewManagerName];
     }
-    return NativeModules.UIManager.getViewManagerConfig(viewManagerName);
+    return UIManager.getViewManagerConfig(viewManagerName);
   };
 
   render() {

### CODE REVIEW

**Comment by lvalentine** [comment]
I hit a similar issue and this appears to have fixed it for me.

**Comment by acoates-ms** [comment]
This prevents the project from working on `react-native-windows` in release mode, since in release mode `react-native-windows` run the `UIManager` as a TurboModule rather than a NativeModule.

**Comment by elicwhite** [comment]
At a high level, I think the approach used by the package needs to change, which will avoid this problem altogether.

Today, these View Configs are read in order to get constants from native.

https://github.com/react-native-video/react-native-video/blob/6af61a35dab0bc6df15286632bcb616b1d28e925/Video.js#L280-L290

The downside of this approach is that on startup of the app React Native has to do reflection do read the values of these constants from native to pass them to JS. To avoid this performance issue, I'd avoid reading the view config for constants at all.

Instead of reading the constants to be able to just pass back to native as a prop (presumably some integer), I'd recommend just using a string. 

Consider changing that code to:

```
if (resizeMode === VideoResizeMode.stretch) {
  nativeResizeMode = 'ScaleToFill';
} else if (resizeMode === VideoResizeMode.contain) {
  nativeResizeMode = 'ScaleAspectFit';
} else if (resizeMode === VideoResizeMode.cover) {
  nativeResizeMode = 'ScaleAspectFill';
} else {
  nativeResizeMode = 'ScaleNone';
}
```

And then you can do the parsing on the native side.

This performance optimization you can make now, but it will actually be required with the new React Native architecture. Natively calculated view configs will no longer exist, so there will no longer be View Manager constants. Those will need to get defined in JS (we recommend doing it with strings like the example above).

namrog84:
not stale.
@esilverberg Do you know what needs to happen to push for this change?

esilverberg:
I am afraid not; I have no direct involvement with this project. I just happened to discover this bug over one year ago and patched it in my own branch.

hueniverse:
I'm confused by the comment from @acoates-ms. Is this fix going to break Windows?

hueniverse:
Merged via another PR.

Issue #2141, opened 2020-09-07
How to get buffer progress while paused?


nickfujita:
# Feature Request

Provide a way to access buffer progress while video is paused.

## Why it is needed

There are cases where users with slower connections will pause the video, and wait until they can see that their buffer indicator has progressed a self determined amount before starting playback.

This does not appear to be possible, as the only place where buffered state appears to be accessible is via the `onProgress` event. Based on the documented description, the `onProgress` event should be fired every `progressUpdateInterval`. However in both the documentation for `onProgress` and `progressUpdateInterval`, there does not appear to be any mention of whether this event should be fired when the player is paused.

While using the player on Android with the exoplayer version, the `onProgress` event does not fire at all while paused. Therefore we are not able to get the `playableDuration` value to show the user buffer progression while paused.

## Possible implementation

1. Update the `onProgress` event to fire even while the player is paused. This will help to provide on progress updates for the buffer state.
2. Provide a separate `onBufferProgress` event which will fire on an interval with the latest `playableDuration` to indicate buffer progress no matter if the player is playing or paused
3. Provide a method which allows the client to request the buffer state `playableDuration` at any given time. This should be accurate to the latest buffer position no matter the play/pause state of the player.

ChrisCodeCole:
+1, could be useful for me as well

cobarx:
We should update the `onBuffer` event as this is already present and people are using it. We just need to get it documented and make sure we know which platforms it works on and what the payloads are.

KingAmo:
it seems that video stop buffering when in pause state

Navipro70:
It's very useful thing. PR is opened, any updates for it?

ghost:
+1, This would be very useful for me as well.

freeboub:
This issue shall be reopened ! as it is still here :)
The onProgress callback shall be sent even if playback is in pause.
This is important for UI especially when you are playing live contents !

Notice that linked PR is not the good fix.
I have a small patch for this, I will try to on PR with it.

freeboub:
```
index d2fc693..743048b 100644
@@ -178,9 +178,7 @@ public class AndroidExoplayerView extends AndroidPlayerItf implements
             switch (msg.what) {
                 case SHOW_PROGRESS:
                     if (player != null
-                            && player.getPlaybackState() == Player.STATE_READY
-                            && player.getPlayWhenReady()
-                            ) {
[...]

PR #2302, opened 2021-03-01
Update Readme to include onBuffer event (#2141)

laurenceunmind:
This PR adds documentation for the `onBuffer` callback function, resolving issue #2141.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ var styles = StyleSheet.create({
 |--|--|
 |[onAudioBecomingNoisy](#onaudiobecomingnoisy)|Android ExoPlayer, iOS|
 |[onBandwidthUpdate](#onbandwidthupdate)|Android ExoPlayer|
+|[onBuffer](#onbuffer)|Android ExoPlayer, iOS|
 |[onEnd](#onend)|All|
 |[onExternalPlaybackChange](#onexternalplaybackchange)|iOS|
 |[onFullscreenPlayerWillPresent](#onfullscreenplayerwillpresent)|Android ExoPlayer, Android MediaPlayer, iOS|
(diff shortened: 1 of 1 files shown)


KingAmo:
`onBuffer`  callback's  payload  `isBuffering` is always false when you are seeking, is this a bug?

Champkinz:
> `onBuffer` callback's payload `isBuffering` is always false when you are seeking, is this a bug?

Im having the same issue, is there a fix ?

jakeorbtl:
> `onBuffer` callback's payload `isBuffering` is always false when you are seeking, is this a bug?

I am also experiencing this issue, but only on iOS. It works as expected for the Android ExoPlayer.

jakeorbtl:
> > `onBuffer` callback's payload `isBuffering` is always false when you are seeking, is this a bug?
> 
> Im having the same issue, is there a fix ?

I came up with a workaround that seems to work for ios. Since the `onSeek` event seems to be called after the video has finished buffering you can use that event to find when buffering has ended. When you seek to an unplayable portion of the video, the `playableDuration` value passed to the `onProgress` event callback becomes 0. We can use this to determine when buffering due to seeking has begun.

```
const onProgress = (data) => {
    setCurrentTime(data.currentTime);
    setPlayableDuration(data.playableDuration);
    setSeekableDuration(data.seekableDuration);

    if (Platform.OS === 'ios') {
        if ((data.playableDuration < data.currentTime)) {
            setIsBuffering(true);
        } else {
            setIsBuffering(false);
        }
    }
};

const onSeek = () => {
    Platform.OS === 'ios' && setIsBuffering(false);
};
```
The `playableDuration` check in `onProgress` also handles buffering when not seeking, so it is not necessary to use `onBuffer` for ios.

Also make sure to add the following:
```
const onEnd = () => {
    setIsBuffering(false);
};
```
Otherwise, your isBuffering flag will remain true when you seek to the end of the video.

hueniverse:
@laurenceunmind Mind rebasing so I can merge it?

laurenceunmind:
> @laurenceunmind Mind rebasing so I can merge it?

Sure, done 👍

Issue #1945, opened 2020-03-12
Setting paused prop to true pauses the video, but then setting it back to false does not play the video


ajpaulingalls:
# Bug

Setting paused prop to true pauses the video, but then setting it back to false does not play the video.  Instead of playing from paused location, it plays a few frames then pauses again.  

## Platform

Which player are you experiencing the problem on:
* Android ExoPlayer

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.15.3
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 64.15 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 13.8.0 - /usr/local/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.14.2 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.13.0 => 16.13.0 
    react-native: 0.61.5 => 0.61.5
```

Library version: 5.1.0-alpha6

## Steps To Reproduce

```bash
[paused, setPaused] = useState(false);

<TouchableOpacity
        onPress={() => {
          setPaused(!paused);
        }}>
  <Video
            source={{uri: videoInfo.signedVideoUrl}}
            paused={paused}
          />
</TouchableOpacity>
```

## Expected behavior

1. Tapping once should pause the video
2. Tapping a second time should start the video at paused point

## Actual behavior
1. Tapping once pauses the video
2. Tapping a second time video start then immediately stops (less than a second).

## Video sample
If possible, include a link to the video that has the problem that can be streamed or downloaded from.

ZhangTaoK:
+1 我也有相同的问题。I have the same question

nickfujita:
Confirmed. This appears to only happen for me when in fullscreen mode thought. Strangely, after setting pause then play again, playback resumes as normal for that time and any future toggles.

gauravroyzz:
whats the fix?
any updates

ivan-vilches:
[...]

PR #2311, opened 2021-03-12
Fix AudoFocus pausing video when attempting to play

armands-malejevs:
Fix AudioFocus bug that could cause the player to stop responding to play/pause in some instances. 

Fixes issue #1945  

This was caused by the player requesting audio focus on each play (un-pause) and that resulted in a small window of Audio focus loss and then gain. The focus loss results in the player being paused while the player was supposed to play at the time. The solution is to keep track of Audio focus and not request new focus if we already have it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Changelog
 
+- Fix Android AudioFocus bug that could cause player to not respond to play/pause in some instances [#2311](https://github.com/react-native-video/react-native-video/pull/2311)
+
 ### Version 5.1.0-alpha9
 
 - Add ARM64 support for windows [#2137](https://github.com/react-native-community/react-native-video/pull/2137)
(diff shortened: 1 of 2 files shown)


deepak-jha-kgp:
Hi @armadilio3  Do you think this could be a reason for #2312 as well?

armands-malejevs:
@deepak-jha-kgp It's possible that the lower end devices don't handle focus loss/gain very quickly and could start stuttering. Looking at the video the symptoms seem to be similar so its possible that this would fix the issue, but the only way to make sure would be using the player with these changes included and checking if its still reproducible.

deepak-jha-kgp:
Thank you @armadilio3. I will implement this and get back to you

t-fritsch:
Tested with RN 0.64 + react-native-video-controls 2.7.1 and this PR seems to work just fine.
Thanks a lot

ajayAloha:
Hi @armadilio3 Thanks a ton for fixing this. I have wasted a day trying to figure out the issue. :)

Issue #836, opened 2017-10-29
i am not able to stop the repeat on iOS


anantanandgupta:
this is my component constructor
```
  constructor(props) {
    super(props);
    this.state = {
      isPaused: true
    }
  }
```
this is in the render function
```
        <Video
          repeat={false}
          ref={(ref) => {
            this.player = ref
          }}
          paused={this.state.isPaused}
          resizeMode='cover'
          source={require('../../../../assets/videos/video.mp4')}
          onEnd={() => {
            this.setState({isPaused: true}, ()=> {
              this.player.seek(0);
            });
          }}
          onLoad={() => {
            this.player.seek(0)
          }}
          style={{
            alignSelf: 'stretch',
            height: videoHeight,
            borderWidth: 0,
            borderColor: '#000000',
            margin: Dimensions.padding,
            backgroundColor: Colors.white
          }}
        />
```

now this works fine in Android but not in iOS ... it keeps on repeating the video in iOS

EthanLin-TWer:
I'm running into this issue as well as the exact same behaviour(working on Android but not on iOS), and I can't explain why though but the following code works for me: 

```js
onEnd={() => {
  this.setState({ isPaused: true })
  this.player.seek(0)
}}
```

basically just put `this.player.seek(0)` after `this.setState()` and do not making it as a callback. Maybe have a quick try to see if it solves your problem.

anantanandgupta:
That was my first try. The call back option was the last step :(

maxschmeling:
I'm having this issue in iOS as well. The workaround by @linesh-simplicity is almost exactly the code I already had that wasn't working.

I was able to get it working again by doing this:
```
componentDidUpdate(prevProps, prevState) {
  if (!prevState.ended && this.state.ended)
    this._video.seek(0);
}

render() {
  return (
    <Video
      onEnd={() => this.setState({ paused: true, ended: true })}
    />
  );
}
```

PR #2329, opened 2021-03-25
#836 Fix video endless loop when repeat set to false or not specified.

Romick2005:
On video end set player pause.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -833,6 +833,7 @@ static int const RCTVideoUnset = -1;
     [item seekToTime:kCMTimeZero];
     [self applyModifiers];
   } else {
+    [self setPaused:true];
     [self removePlayerTimeObserver];
   }
 }


hueniverse:
@nickfujita Is this fix correct? Should be a quick review and if yes, a simple patch of the new code. If not, please close.

nickfujita:
yes this seems right. would allow people to remove their workarounds in #836 

please port to swift here:
https://github.com/react-native-video/react-native-video/blob/master/ios/Video/RCTVideo.swift#L1046

hueniverse:
@Romick2005 Can you rebase and change to swift?

Romick2005:
@hueniverse No I can't, but you can do it on your own `self.setPaused(true);`

hueniverse:
Closed in https://github.com/react-native-video/react-native-video/commit/553e51c63945931e2a5ce3a1665dbaa5c1f2f67e

Romick2005:
Could it be just merged? As there are no progress and package stuck in alfa for a year?

PR #2372, opened 2021-05-18
Add player.getVideoBufferedDuration()

justsml:
Addresses #2141 on Android. 

Implemented the 3rd suggested option.

The `getVideoBufferedDuration()` method returns the number of milliseconds buffered.

The current event-based code calculates `playableDuration` by converting `mVideoBufferedDuration` to seconds. Not sure if the conversion to seconds should be in the component implementation...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1189,6 +1189,19 @@ return (
 );
 ```
 
+#### getVideoBufferedDuration
+`getVideoBufferedDuration()`
+
+Return the current video buffered duration.
+
+Example:
+```js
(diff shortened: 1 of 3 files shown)


freeboub (changes requested):
Why is this change needed ?
You should just register onProgress callback ?
Anyway, I can understand such request, but it would be better to rename it to getProgressInfo and return the same object than onProgress

freeboub:
I push a better fix here: https://github.com/react-native-video/react-native-video/pull/2664

PR #2395, opened 2021-06-03
Make sure modifiers are applied before playing

anderslemke:
### Problem

On iOS, a video with `allowsExternalPlayback={false}` will hijack an ongoing audio playback if playing on AirPlay.

Steps to reproduce:
1. Start audio playing via AirPlay.
2. Initialize component with `allowsExternalPlayback={false}`
3. Now the audio will unexpectedly be interrupted, while nothing will start playing on AirPlay.

### Fix
By setting the modifiers right after a player is initialised, we ensure all configuration is respected when playback starts.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -383,6 +383,7 @@ static int const RCTVideoUnset = -1;
       }
       
       self->_player = [AVPlayer playerWithPlayerItem:self->_playerItem];
+      [self applyModifiers];
       self->_player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
       
       [self->_player addObserver:self forKeyPath:playbackRate options:0 context:nil];


nickfujita:
@anderslemke would you mind testing this PR for the iOS conversion to swift? It does not have the changes you have made here, but I'd be happy to add them in afterwards. I'm looking to have anyone who has contributed to ios test out the converted version. Please be aware that there is a required update to add `use_frameworks! :linkage => :static` to your project Podfile
https://github.com/react-native-video/react-native-video/pull/2527

Deodes:
@hueniverse hope this PR was tested before release new version, because last time when I try to use it, my app crashes. To reproduce it, I changed `source` multiple times in my `<Video/>`

hueniverse:
@Deodes thanks for the feedback. I did some basic tests (that's really all we can do right now and catch bugs are people try out the new version which will have a pre-release label on it) and didn't see an issue. Are you able to either test it locally again or fork https://github.com/react-native-video/react-native-video-test and change it to reproduce the issue? If it still breaks things I'll revert the change.

PR #2412, opened 2021-06-30
Update exoplayer to allow pre-init and content clear

nickfujita:
Exoplayer was not able to be initialized without a source, but it has now been updated to:
- Allow player to be init before source is provided, and later update once a source is provided.
- Adds handling for providing a empty source in order to stop playback and clear out any existing content

### DIFF
--- a/README.md
+++ b/README.md
@@ -770,6 +770,12 @@ Platforms: Android ExoPlayer
 #### source
 Sets the media source. You can pass an asset loaded via require or an object with a uri.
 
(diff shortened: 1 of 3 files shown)


zchwyng:
LGTM. This did solve my problems where getting an error of "Unrecognized media format" would make all videos black.

mirceaciu:
I still get a warning if I set `source={null}`. ~When build for production app does not start.~ 

Is this expected?

Issue #2157, opened 2020-09-22
ignoreSilentSwitch not working on iOS


jamesdarabi:
# Bug

Setting the `ignoreSilentSwitch` prop to `ignore` [as specified in the docs](https://github.com/react-native-community/react-native-video#ignoresilentswitch) doesn't work on iOS. @akhan118 indicates [here](https://github.com/react-native-community/react-native-video/pull/403#issuecomment-646858919) that he is having a similar problem - specifically, he thinks the issue might be related to having the `paused` prop set to `true`.

## Platform

Which player are you experiencing the problem on:
* iOS

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 3.02 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.11.0 - ~/.nvm/versions/node/v14.11.0/bin/node
    Yarn: Not Found
    npm: 6.14.8 - ~/.nvm/versions/node/v14.11.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.0, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
    Xcode: 12.0/12A7209 - /usr/bin/xcodebuild
  Languages:
    Java: javac 15 - /usr/bin/javac
    Python: 3.8.5 - /Users/jamesdarabi/.pyenv/shims/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: 0.63.2 => 0.63.2
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 0.63.2

## Steps To Reproduce

1. Set `ignoreSilentSwitch` prop to `ignore`.
2. Try to play video while iOS device is in silent mode.
3. The video plays without any audio.

## Expected behaviour

1. The audio plays despite the iOS device being in silent mode.

## Reproducible sample code

```
<Video
  controls={true}
  ignoreSilentSwitch={'ignore'} // This doesn't work.
  onEnd={() =>
[...]

PR #2421, opened 2021-07-12
Resolved ignored silent switch in iOS when player is set to paused

DentVega:
This resolves an issue where setting a video to paused would ignore the “silent switch” setting, as reported in #2157. By moving the audio configuration outside of the paused setter, we can update as needed when the silenced variable is set/changed, as one would expect.

To Test

1. Create a video component that is...
- set to paused
- set fullscreen to true
- ignoreSilentSwitch = 'ignore'
2. Set the iPhone to "silenced" mode via the switch on the side
3. Play the video.
4. You should be able to hear audio playing (previously it was not)


RaghuramLekkala:
Still not able to ignore silent switch.  It works once I lock and unlock device.

PR #2436, opened 2021-08-01
Android fix

olgunkaya:
missing interface of the new exo-player library removed. 
version updated


olgunkaya:
sorry for wrong place. closed.

PR #2512, opened 2021-10-17
fix: fix ios video not playing in background

### LINKED ISSUE #1735: [iOS] Next video does not play while in background
# Bug

Only tested in iOS.

If the app is in the foreground, setting a new source will update the player and a new video begins to play.

If the app is in the background, setting a new source does cause the `onLoadStart` and `onLoad` events to fire, but the next video does not play until the app is brought back into the foreground.

## Environment info
```
react: 16.8.6
react-native: 0.60.0
```

Library version: 5.0.0

## Steps To Reproduce

1. Setup a player with background audio, and a video source of your choice.
2. Programmatically change the source to a different video. (Using `setTimeout` is probably the best bet due to the nature of this issue).
3. Observe how the player changes the video.
4. Repeat steps 1 - 2, but after the first video begins to play put the app into the background by going to the home screen or locking the device.
5. Observe how the player does not automatically play the next video.
6. Open the app up, observe how the player immediately begins playing the next video source.

Describe what you expected to happen:

1. In either foreground or background mode, the video should change when the source prop updates.

## Reproducible sample code

```
const [uri, setUri] = useState("https://myvideo.com/one.mpeg");

setTimeout(() => setUri("https://myvideo.com/two.mpeg"), 10000);

return <Video
    ignoreSilentSwitch="ignore"
    playInBackground
    playWhenInactive
    source={{ uri }}
    onLoad={handleLoad}
    onLoadStart={handleLoadStart}
/>
```

polnikale:
fixes #1735

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -223,6 +223,10 @@ static int const RCTVideoUnset = -1;
 
 #pragma mark - App lifecycle handlers
 
+- (BOOL)canBecomeFirstResponder {
+  return YES;
+}
+
 - (void)applicationWillResignActive:(NSNotification *)notification
 {
   if (_playInBackground || _playWhenInactive || _paused) return;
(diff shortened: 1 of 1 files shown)


gaurav-cultfit (approved):

nickfujita:
@polnikale would you mind testing this PR for the iOS conversion to swift? It does not have the changes you have made here, but I'd be happy to add them in afterwards. I'm looking to have anyone who has contributed to ios test out the converted version. Please be aware that there is a required update to add `use_frameworks! :linkage => :static` to your project Podfile
https://github.com/react-native-video/react-native-video/pull/2527

hueniverse:
@rorygarand are you able to test this patch and lmk if it fixes your issue? I know it's been a long time but hoping you can chime in...

Issue #2095, opened 2020-07-23
Video repeats after reaching the end


khattab93:
# Bug

## Platform

Which player are you experiencing the problem on:
* iOS

## Environment info

React native info output:
Library version: 5.0.1

## Steps To Reproduce

1. Load a local video 
2. Play it
3. When it reaches the end, I seek it to start

## Expected behaviour

1. The plying should stop when reaching the end.

## Reproducible sample code
```
<Video
            ref={this.setVideoPlayerRef}
            style={styles.videoPreview}
            source={{ uri: Image.resolveAssetSource(require(`./test.mp4`)).uri }}
            resizeMode={"cover"}
            paused={paused}
            onLoad={this.getVideoDuration}
            onProgress={this.onProgress}
            onEnd={this.endPlaying}
          />

  private onProgress = ({ currentTime }): void => {
    if (currentTime >= this.state.trimmingEnd) {
      this.endPlaying()
    }
  }

  private endPlaying = (): void => {
    if (this.videoPlayer) this.videoPlayer.seek(0)
    this.setState({ paused: true })
  }
```

## Video sample
https://drive.google.com/file/d/17hZnL8tUUSXW8PARIZpujKUW3FovZvqu/view?usp=sharing

hengkx:
same problem

kraj011:
Same problem. Pausing works just fine before the video ends, but once the video ends and I update my paused state to true, the video still continues playing. I also have repeat={false} as a prop, which clearly is not working as intended. Might have something to do with the seek method (i.e. this.player.seek(0))? Not sure.

kr396:
Same issue

dev-fuad:
I had the same issue
this is the workaround I used
```
    const onEnd = useCallback(() => {
      player.current.seek(0);
      setTimeout(() => {
        setIsPaused(true);
      }, 100);
    }, []);
```

kr396:
> I had the same issue
> this is the workaround I used
> 
> ```
>     const onEnd = useCallback(() => {
>       player.current.seek(0);
>       setTimeout(() => {
>         setIsPaused(true);
>       }, 100);
>     }, []);
> ```

I will try this. Thnak you.

uglen:
> I had the same issue
> this is the workaround I used
> 
> ```
>     const onEnd = useCallback(() => {
>       player.current.seek(0);
>       setTimeout(() => {
>         setIsPaused(true);
>       }, 100);
>     }, []);
[...]

PR #2517, opened 2021-10-21
IOS: Fix seek bug that start playing again

msimon:
#### Update the changelog

Working example here: https://github.com/react-native-video/react-native-video/issues/2095

#### Describe the changes

Fixing this bug: https://github.com/react-native-video/react-native-video/issues/2095

The original code made an assumption that the variable `_paused` will not be changing between the time we call seek and the time seek callback execute.

If the user was calling seek and pause in `onEnd`, the seek function would set `BOOL wasPaused = _paused;` to false (as it's not yet paused), and restart the video in the callback:
```
if (!wasPaused) {
    [self setPaused:false];
}
```
If `_paused` changed during that time, then `_paused` would be true but `wasPaused` would be false, restarting the playback. Even if `props.paused` was still true on the component.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -35,26 +35,26 @@ static int const RCTVideoUnset = -1;
   NSURL *_videoURL;
   BOOL _requestingCertificate;
   BOOL _requestingCertificateErrored;
-  
+
   /* DRM */
   NSDictionary *_drm;
   AVAssetResourceLoadingRequest *_loadingRequest;
-  
+
   /* Required to publish events */
   RCTEventDispatcher *_eventDispatcher;
   BOOL _playbackRateObserverRegistered;
   BOOL _isExternalPlaybackActiveObserverRegistered;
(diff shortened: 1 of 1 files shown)


ios/Video/RCTVideo.m
```diff
@@ -967,38 +967,40 @@ - (void)setSeek:(NSDictionary *)info
...
+        [_player pause];
+      }
       [_player seekToTime:cmSeekTime toleranceBefore:tolerance toleranceAfter:tolerance completionHandler:^(BOOL finished) {
         if (!_timeObserver) {
           [self addPlayerTimeObserver];
         }
-        if (!wasPaused) {
-          [self setPaused:false];
+        if (!wasPaused && !_paused) {
+          [_player play];
```
msimon, line 992: I replaced the called from `setPaused:false` to only inverse the call l.985.

hueniverse:
@msimon Can you please rebase if this is still something you want to get reviewed?

hueniverse:

PR #2528, opened 2021-11-04
Allows custom exoplayer Version

sidferreira:
This is one of the many changes I had to do to handle delegates

### DIFF
--- a/android-exoplayer/build.gradle
+++ b/android-exoplayer/build.gradle
@@ -28,7 +28,7 @@ android {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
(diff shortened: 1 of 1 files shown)


harshaliitr:
This PR is incomplete. According to docs you should not mismatch versions of exoplayer and exoplayer:extension-okhttp.

you need to update the following lines as well - 

`implementation('com.google.android.exoplayer:extension-okhttp:2.13.2') {
        exclude group: 'com.squareup.okhttp3', module: 'okhttp'
    }`

hueniverse:
@sidferreira I don't see the feedback from @harshaliitr addressed in the PR. Is this still needed?

sidferreira:
@hueniverse we use it

hueniverse:
Thanks for getting back to me. What about the review comment from @harshaliitr? It will also require rebasing.

PR #2551, opened 2021-12-06
clarify onProgress documentation

mlc:
The docs for `progressUpdateInterval` say that the value is in milliseconds, but the docs for `onProgress` said that the value was in seconds. This fixes the latter documentation to be consistent with the former.

This is a documentation-only update with no code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1113,7 +1113,7 @@ Platforms: all
 
 
 #### onProgress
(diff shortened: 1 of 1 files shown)


sakshi2207:
Hi @mlc is this issue is open can I please take it

freeboub (approved):

PR #2555, opened 2021-12-11
Improves ExoPlayer seek performance

vemarav:
Currently, the seek takes nearly 2 seconds to finish on local files(file:///<path to file>) in android.

The default exoplayer SeekParamaters is EXACT. If we replace it with CLOSEST_SYNC the seek happens instantly with ±1 sec.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -71,6 +71,7 @@ import com.google.android.exoplayer2.upstream.DefaultAllocator;
 import com.google.android.exoplayer2.upstream.DefaultBandwidthMeter;
 import com.google.android.exoplayer2.upstream.HttpDataSource;
 import com.google.android.exoplayer2.util.Util;
+import com.google.android.exoplayer2.SeekParameters;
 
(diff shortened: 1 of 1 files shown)


freeboub:
looks Googe for me, low risk

hueniverse:
@freeboub please submit an official review through GitHub. Click on the Files Changed tab and then Review and approve.

freeboub (approved):

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -414,6 +415,7 @@ public void run() {
                                 .setTrackSelector​(trackSelector)
                                 .setBandwidthMeter(bandwidthMeter)
                                 .setLoadControl(defaultLoadControl)
+                                .setSeekParameters(SeekParameters.CLOSEST_SYNC)
```
armands-malejevs, line 418: This could be a breaking change for some, can you make this into a prop where Exact is the default so that closest sync is an optional?

(I think that this might be an issue for SSAI streams where you need to snap exactly to the beginning or end of an ad, I haven't confirmed this but the change sounds like it could have an effect on this use case)

armands-malejevs (changes requested):

hueniverse:

PR #2579, opened 2022-01-21
fix: match letterbox exoplayer version and version upgrade changes

leonchabbey:
Fork main repo because we needed this library to use the exact same version as letterbox --> `3.14.0`
Some changes were necessary following the upgrade


leonchabbey:
my bad, this PR was supposed to be on my fork but github automatically selected the destination to be this repo and not mine which i saw too late

PR #2595, opened 2022-02-14
Fix/issue 2594/content

### LINKED ISSUE #2594: Fairplay skd contentId parse issue with getLicense
# Bug
```
 <Video
        onProgress={onProgress}
        onEnd={onEnd}
        onError={(e) => console.log(e)}
        paused={state?.paused}
        ref={player}
        source={{uri: 'https://n.com/test.m3u8}}
        style={{width: "100%", height: "100%", zIndex: 1}}
        resizeMode={"cover"}
        drm={
              {
                  type: 'fairplay',
                  certificateUrl: `CERT_URL`,
                  base64Certificate: true,
                  getLicense: (spcBase64, contentId, licenseUrl) => {
                    return fetch('LICENSE_URL', {
                      method: 'POST',
                      headers: {
                        ...
                        'Content-Type': 'application/x-www-form-urlencoded',
                      },                     
                    }).then(response => response.text()).then((response) => {
                      console.log(response);
                      return response;
                    }).catch((error) => {
                      console.error('Error', error);
                    });
                  }
            }
        }
      />
```
#### m3u9
```
...
EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-KEY:METHOD=SAMPLE-AES,URI="skd://asdsad23ss/rw==", ...
...
```

## Platform
* iOS

## Environment info

Library version: 5.2.0

## Steps To Reproduce
- When i set `getLicense` in drm property, below results.

  - Expected results: asdsad23ss/rw==
  - Actual results: asdsad23ss

purpletortoise-choi:

### DIFF
--- a/DRM.md
+++ b/DRM.md
@@ -16,10 +16,17 @@ DRM object allows this members:
 | [`certificateUrl`](#certificateurl) | string | undefined | iOS | Specifies the url to obtain your ios certificate for fairplay, Url to the .cer file |
 | [`base64Certificate`](#base64certificate) | bool | false | iOS | Specifies whether or not the certificate returned by the `certificateUrl` is on base64 |
 | [`getLicense`](#getlicense)| function | undefined | iOS | Rather than setting the `licenseServer` url to get the license, you can manually get the license on the JS part, and send the result to the native part to configure FairplayDRM for the stream |
+| [`fullUriString`](#fullUriString)| bool | false | iOS | if set `getLicense` propperty, choose `host` or `full uri` for `contentId` | 
 
 ### `base64Certificate`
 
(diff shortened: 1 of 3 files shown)


hueniverse:
@purpletortoise-choi Is there a way of solving this without adding another config option?

adityapatadia:
Why was this closed? Is this resolved?

freeboub:
It was stale, btw it was reimplemented on V6. Should be ok now

adityapatadia:
As in, should we use beta version?

freeboub:
> As in, should we use beta version?

Yes it is available in last 6.0.0 beta ! No maintenance planned on 5.2

PR #2609, opened 2022-03-04
Always check for hideShutterView before showing shutterView

tironiigor:
# Always check for hideShutterView before showing shutterView

## Changes

For every call, we made to set `shutterView` visibility to VISIBLE we should check for the `hideShutterView` variable to avoid flashing a black screen. This makes the Android consistent with iOS

## Testing

1. Pass `hideShutterView={true}` to the component
2. Before this fix it would flash a black screen before hiding it


andac-ozcan (approved):

freeboub (changes requested):
Ok for me, can you just fix the changelog please ?

tironiigor:
@freeboub done!

tironiigor:
@freeboub should it be alpha1 or alpha2?

hueniverse:
We're still working on alpha1.

tironiigor:
Done @hueniverse @freeboub

Issue #2475, opened 2021-08-31
AVFoundationErrorDomain Code -11839: Cannot Decode


tennispro7797:
# Bug
When loading a screen full of videos with a pagination size of 15, there are multiple videos that do not render and return the following error:
```
Object {
  "error": Object {
    "code": -11839,
    "domain": "AVFoundationErrorDomain",
    "localizedDescription": "Cannot Decode",
    "localizedFailureReason": "The decoder required for this media is busy.",
    "localizedRecoverySuggestion": "Stop any other actions that decode media and try again.",
  },
  "target": 2607,
}
```
After looking through https://github.com/react-native-video/react-native-video/issues/1785, I thought perhaps having 15 AVPlayer instances was too many so I tried reducing the pagination size to 5 and the same number of videos failed to render.

## Platform
* iOS 14.7.1
* iPhone 12

## Environment info

```
  Expo CLI 4.9.1 environment info:
    System:
      OS: macOS 11.4
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 14.17.3 - /usr/local/bin/node
      npm: 6.14.13 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    IDEs:
      Android Studio: 4.2 AI-202.7660.26.42.7351085
      Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^42.0.0 => 42.0.3 
      react: 16.13.1 => 16.13.1 
      react-dom: 16.13.1 => 16.13.1 
      react-native: https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz => 0.63.2 
      react-native-web: ~0.13.12 => 0.13.18 
      react-navigation: ^4.0.0 => 4.4.4 
    npmGlobalPackages:
      expo-cli: 4.9.1
    Expo Workflow: managed
```

## Reproducible sample code

```
  const renderMediaGrid = (item: any, index: number, isEven: boolean) => {
    return (
      <View>
            <Video
              style={{
                width: window.window.width / 2 - 25,
                height: window.window.height * 0.3,
                marginHorizontal: 10,
                marginBottom: 25,
                borderRadius: 2,
                backgroundColor: Colors.lightGrey,
                borderWidth: 5,
              }}
              source={{
[...]

PR #2640, opened 2022-04-15
Fix issue using video player on FlatList - AVFoundationErrorDomain Code -11839

khanhld1910:
## What

Issue: https://github.com/react-native-video/react-native-video/issues/2475

Using `react-native-video` with FlatList has the issue being reported above

## Why

I think the cause is the video player is not being deallocated when the component (here is the FlatList item) being unmounted.

## How

I guess we could deallocate the Video player using `componentWillUnmount` and support calling `destroy` method by native module.

I tested with FlatList this way and it worked but I don't know object C and I think there could be another way to fix this issue so this PR is for notifying and ask for help on this issue.

Please help me review it and comment your better idea. Thanks a lot!

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -79,6 +79,10 @@ export default class Video extends Component {
     return await NativeModules.VideoManager.save(options, findNodeHandle(this._root));
   }
 
+  componentWillUnmount() {
+    NativeModules.VideoManager.destroy(findNodeHandle(this._root));
+  }
+
   restoreUserInterfaceForPictureInPictureStopCompleted = (restored) => {
     this.setNativeProps({ restoreUserInterfaceForPIPStopCompletionHandler: restored });
   };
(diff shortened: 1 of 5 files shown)


hueniverse:
Thank you for the contribution. I'm in the same boat so would appreciate if anyone else could review this PR and provide feedback.

@khanhld1910 if you can create a sample app using https://github.com/hueniverse/react-native-video-test to show how the fix works, it would help validate it and get it merged quicker.

khanhld1910:
@hueniverse Thank you! I've just tested my PR and it was having other issues still.
So I have to spend more time on this and I will create a sample app when it's better.
This issue was known for a while and hopefully there will be other people join in to help.

hueniverse:
@khanhld1910 Ping me when it's ready for review!

hueniverse:

PR #2645, opened 2022-04-20
Disable pipController init if pictureInPicture is false

tombaranowicz:
Disable pipController init if pictureInPicture is set to false prevents going to PIP mode on iOS when the app goes to the background. 

Tested on both iOS 14 and iOS 15 devices.

CC @hueniverse


hueniverse:
Much appreciated!

PR #2651, opened 2022-04-21
Fixes Android build fail.

AdityaXJha:
Added jitpack support for Android-ScalableVideoView as jcenter is deprecated.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,5 +22,5 @@ android {
 dependencies {
     //noinspection GradleDynamicVersion
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    implementation 'com.yqritc:android-scalablevideoview:1.0.4'
+    implementation 'com.github.adityaxjha:Android-ScalableVideoView:10e7b4da8b'
 }


freeboub:
This PR is linked MediaPlayer, but it should be dropped...
Not sure this is needed

freeboub:
This PR is linked MediaPlayer, but it should be dropped...
Not sure this is needed

hueniverse:
MediaPlayer support is dropping in v6.

@freeboub Can you PR removing code that is only used for MediaPlayer support.

evanjmg:
Any update on this? We need to not use jcenter

freeboub:
I open following pr to remove mediaplayer support: https://github.com/react-native-video/react-native-video/pull/2681

arsimr16:
I was able to use `react-native-video 5.2.0` without using jcenter by switching to the Android ExoPlayer.  `android-scalablevideoview` is only used by the Android MediaPlayer which is being removed in version 6.x of react-native-video anyway.

1) Add the following lines to settings.gradle to use the ExoPlayer instead of the MediaPlayer:
```
include ':react-native-video'
project(':react-native-video').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-video/android-exoplayer')
```

2) Once I switched to the ExoPlayer, I got errors from `react-native-video/android-exoplayer/build.gradle` about `com.google.android.exoplayer:exoplayer:2.13.2` and `com.google.android.exoplayer:extension-okhttp:2.13.2` not being available.  I made a patch to change both versions to `2.13.3`, and my app built successfully after that.

PR #2664, opened 2022-04-27
fix(android, exoplayer): ensure playback progress is reported

freeboub:
fix(android, exoplayer): ensure playback progress is reported even if player is paused

Adresses #2141 on Android.
and shall fix: https://github.com/react-native-video/react-native-video/pull/2372

A live channel sample is available in basic sample.
On live stream, position indicator keeps updating after playback is paused.

#### Focus area
You can ensure that position is not reported too often

#### Describe the changes
- Don't test player state before sending time updates
- safety check to avoid sending the same value multiple times (avoid UI updates)


hueniverse:
Thanks! As soon as we get it reviewed it will be merged!

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -165,21 +165,30 @@ class ReactExoplayerView extends FrameLayout implements
...
+    private long lastPos = -1;
+    private long lastBufferDuration = -1;
+    private long lastDuration = -1;
+
     private final Handler progressHandler = new Handler() {
         @Override
         public void handleMessage(Message msg) {
             switch (msg.what) {
                 case SHOW_PROGRESS:
-                    if (player != null
```
armands-malejevs: Why did you remove this null check? Wouldn't it be safer to leave it there? (I'm OK with removing the state checks, but null check might be beneficial here).

armands-malejevs (changes requested):

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: I fully agree ! change done

armands-malejevs (approved):

freeboub:
Just fix conflict, ready to merge

Issue #2665, opened 2022-04-28
MaxBitRate is ignored in Android


evkwan:
# Bug

Problem: 
After setting MaxBitRate property, Android Exoplayer is seen to ignore this value and continues downloading/playing the video track with highest bitrate in the HLS playlist.

## Platform

* Android ExoPlayer

## Environment info

```json
"react-native": "0.66.4",
"react-native-video": "5.2.0",
```

## Steps To Reproduce

1. Play a video with HLS list (.m3u8 format), having a few resolutions and set MaxBitRate to a value such that it should only qualify 3 tracks.

Example master manifest file:
```
#EXT-X-STREAM-INF:BANDWIDTH=2516370,AVERAGE-BANDWIDTH=2516370,CODECS="mp4a.40.2,avc1.640020",RESOLUTION=720x1280,CLOSED-CAPTIONS=NONE

#EXT-X-STREAM-INF:BANDWIDTH=1640580,AVERAGE-BANDWIDTH=1640580,CODECS="mp4a.40.2,avc1.640020",RESOLUTION=540x960,CLOSED-CAPTIONS=NONE


#EXT-X-STREAM-INF:BANDWIDTH=4755240,AVERAGE-BANDWIDTH=4755240,CODECS="mp4a.40.2,avc1.64002a",RESOLUTION=1080x1920,CLOSED-CAPTIONS=NONE


#EXT-X-STREAM-INF:BANDWIDTH=994068,AVERAGE-BANDWIDTH=994068,CODECS="mp4a.40.2,avc1.64001f",RESOLUTION=360x640,CLOSED-CAPTIONS=NONE


#EXT-X-STREAM-INF:BANDWIDTH=697078,AVERAGE-BANDWIDTH=697078,CODECS="mp4a.40.2,avc1.64001e",RESOLUTION=270x480,CLOSED-CAPTIONS=NONE

```
With the list, above, set `maxBitRate={1640580}`

2. Observe on Android Exoplayer, it still downloads the highest resolution chunks to play after the initial chunk. You can cross check the downloaded chunk url belongs to the manifest of the highest resolution.

## Expected behaviour

After setting MaxBitRate, Android Exoplayer should be downloading/playing only video tracks that honour the max bitrate set. 

## Reproducible sample code
```
<Video
[...]

PR #2670, opened 2022-05-03
#2665 - fix maxbitrate property being ignored on Android 

evkwan:
#### Update the documentation
- No new property added

#### Update the changelog
- updated change log

- Test steps can be found on issue page [#2665](https://github.com/react-native-video/react-native-video/issues/2665) 

- single fix

#### Describe the changes
- Inside ReactExoplayerView.java `videoLoaded()`, added guards to check type before allowing track selection overwrite. If type is null, don't do a track selection overwrite as that would cause a default `type=auto` overwrite to happen where it allows all available tracks to be used for ABR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 - Fixed Android seeking bug [#2712](https://github.com/react-native-video/react-native-video/pull/2712)
 - Fixed `onReadyForDisplay` not being called [#2721](https://github.com/react-native-video/react-native-video/pull/2721)
 - Fix type of `_eventDispatcher` on iOS target to match `bridge.eventDispatcher()` [#2720](https://github.com/react-native-video/react-native-video/pull/2720)
+- Fix maxBitRate property being ignored on Android [#2670](https://github.com/react-native-video/react-native-video/pull/2670)
 
 ### Version 5.2.0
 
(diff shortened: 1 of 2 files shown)


hueniverse:
@evkwan Thank you! Change looks good but how does it relate to maxBitRate?

evkwan:
@hueniverse when MaxBitRate is set, at first ExoPlayer's AdaptiveTrackSelection will honour the value set and only select tracks that matches the bitrate threshold. However, once the function `videoLoaded()` runs, that line of setSelectedVideoTrack will run with `videoTrackType == null` (here I am assuming if user puts a MaxBitRate value they won't want to put a specific track selection, as track selection locks the player to play only 1 track and defeats the purpose of the threshold value set!).

When `setSelectedVideoTrack()` runs with type null, it cause all tracks to be selected instead. You can see some screenshots from the issue thread [here](https://github.com/react-native-video/react-native-video/issues/2665).

hueniverse:
Thanks @evkwan.

Hopefully we'll get two reviews quickly and get this merged.

freeboub (approved):

hueniverse:
Can we get one more person to verify the fix and approve it?

hueniverse:
@evkwan Can you do a quick rebase? The android files moved around. Thank you!

hueniverse (approved):

evkwan:
Hi @hueniverse , sorry for the delay in response. In the process I sort of messed up the rebase a bit 😓  but I think it should be fixed now. I have added one update to update the changelog pointing to the PR number instead of issue number.  Please let me know if this is OK for merge. Thanks and sorry for the inconvenience.

PR #2678, opened 2022-05-12
Treat black screen as transparent before onLoad when using exoplayer

junbeomheo:
hello.
I can not speak English very well.
I used the Android media player in this library, but there was a bug where autoplay (repeat) didn't work, so I replaced it with EXO-Player and changed the phenomenon that resets to a black screen before loading the video to be transparent. This will make it easier for developers to customize the background at the JavaScript level, and I think it can be used in common in the UI. thank you.

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -71,7 +71,7 @@ public final class ExoPlayerView extends FrameLayout {
 
         shutterView = new View(getContext());
         shutterView.setLayoutParams(layoutParams);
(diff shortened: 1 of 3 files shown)


tironiigor:
Nice idea!

tironiigor:
@junbeomheo isn't it the same as using [hideShutterView](https://github.com/react-native-video/react-native-video#hideshutterview)?

junbeomheo:
@tironiigor  just discovered this option, but the hideShutterView:boolean option doesn't handle black screens. It doesn't work. Why do you have this option? Is it a performance issue? I wonder why it's not transparent by default. If transparency is the default, I think UI handling at the react.js level would be easier.

tironiigor:
@junbeomheo I thought hiding the shutter view would allow you to have components behind it

Issue #2656, opened 2022-04-22
Remove Android MediaPlayer support


hueniverse:
With limited resources we need to focus on ExoPlayer. It is the player used by most users of this module.

swikars1:
I agree, lets remove it

PR #2681, opened 2022-05-15
Remove Android MediaPlayer support #2656

freeboub:
#### Update the documentation
Remove MediaPlayer References (execpt in old release description)

Tested in basic sample (path integration updated)

Just ensure changing the exoplayer-android to exoplayer is enough

#### Describe the changes
remove deprecated Media Player implementation as requested here:
https://github.com/react-native-video/react-native-video/issues/2656


evanjmg:
Plz merge this

michaelbayday:
Could you folks please merge this?

hueniverse:
Everyone - Please stop asking for things to be merged. Bugging the maintainers is not going to magically create free resources and time. We are doing the best we can with the resources we have. Nothing is preventing you from forking and creating temporary patched versions until the main repo is back to normal. This will take a few weeks to a few months. Thank you for your understanding.

hueniverse:
@armadilio3 Is this PR going to create merge issues for your Android PR coming?

freeboub:
@hueniverse i wait your GO to fix conflicts ;) thanks !

armands-malejevs:
@hueniverse I think that there is a good chance that it will. Looking at the changes it appears that some Exoplayer files have been moved, really depends on how well git will resolve moving files and editing them.

I would really want to see this merged in, but I'd hold off until we merge #2689 and deal with the conflicts after.

hueniverse:
@freeboub If you can help review #2689 that would move things along!

freeboub:
I did a git mv on folder so it should work but it seems it doesn't work :'(
Can can re create the patch if needed, it was 1 hour job :)
I agree let's merge @armadilio3 PR first !

freeboub:
Thinking back on this PR, @hueniverse don't you think it would be better to swap players ? Ie, put exoplayer as default and keep old mediaplayer for 'dummy' video use cases ?

evanjmg:
Doesn’t using the old player require jcenter? Best to remove altogether

hueniverse:
@freeboub less code is better. We don't have enough resources so we should support the minimum possible.

README.md
```diff
@@ -343,26 +335,26 @@ var styles = StyleSheet.create({
 |[onBuffer](#onbuffer)|Android ExoPlayer, iOS|
 |[onEnd](#onend)|All|
 |[onExternalPlaybackChange](#onexternalplaybackchange)|iOS|
-|[onFullscreenPlayerWillPresent](#onfullscreenplayerwillpresent)|Android ExoPlayer, Android MediaPlayer, iOS|
-|[onFullscreenPlayerDidPresent](#onfullscreenplayerdidpresent)|Android ExoPlayer, Android MediaPlayer, iOS|
-|[onFullscreenPlayerWillDismiss](#onfullscreenplayerwilldismiss)|Android ExoPlayer, Android MediaPlayer, iOS|
-|[onFullscreenPlayerDidDismiss](#onfullscreenplayerdiddismiss)|Android ExoPlayer, Android MediaPlayer, iOS|
+|[onFullscreenPlayerWillPresent](#onfullscreenplayerwillpresent)|Android ExoPlayer, iOS|
```
armands-malejevs, line 338: Can we just use `Android` instead of `Android ExoPlayer` since we only have ExoPlayer

android/build.gradle
```diff
@@ -8,19 +8,43 @@ android {
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
     buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
 
+    compileOptions {
+        targetCompatibility JavaVersion.VERSION_1_8
```
armands-malejevs, line 12: Is this required? We might run into issues on older Android devices that only support Java 7 features, since the issue will no longer be caught by the compiler. This change itself shouldn't cause issues, but if a developer uses Java 8 features then it might not work on Android 7 and older devices.

armands-malejevs (approved):
Looks good. See some of my minor comments if you want to address them or not.

android/build.gradle
freeboub: I am not sure, but I think yes ... This is not my code, it already there in master branch :/ I just did a " git mv "

freeboub: It was introduced in version : 5.1.0-alpha7

freeboub:
I am not able to merge this PR with master, it was easier to create a new one here:
https://github.com/react-native-video/react-native-video/pull/2724

Regarding to comments:
- I fix Readme.md to change Android Exoplayer to Android (and remove Android MediaPlayer of course)
- For JAVA 1.8, I don't think it will cause interoperability issues:  https://developer.android.com/studio/write/java8-support

Issue #2684, opened 2022-05-16
`isPlayerPlaying` API for Android Native ExoPlayer


custadian:
### Current behavior
I don't see such API for Android.

### Expected behavior
By using this API, I can check if the Player is currently Playing or not.

### Platform
Specifically looking for 
* Android ExoPlayer

Is there any API where I can check if the Player is playing or not ?

freeboub:
on master there is now : onPlaybackStateChanged : {isPlaying}
Let close this issue as done !

Issue #2683, opened 2022-05-15
Split AudioTracks, VideoTracks & TextTracks events notification from onLoad to have dedicated events 


freeboub:
# Feature Request

Split AudioTracks, VideoTracks & TextTracks events notification from onLoad to have dedicated events.
- remove Audio / Text / Video Tracks fields from onLoad event
- Create a new event (or 3 different events) to handle these message 

## Why it is needed

Dash implementation allows to have multiple Period.
see: https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/Timeline.Period.html
This period management can be used to do server side ads insertion.
But ads inserted may not have the same track configuration that original content.
To allow dynamic update of tracks from an external UI, It is needed to dynamically update tracks configuration 

## Possible implementation

- remove Audio / Text / Video Tracks fields from onLoad event
- Create a new event (or 3 different events) to handle these message

hueniverse:
Why not keep the current event as-is and add the new events in addition?

freeboub:
Yes for sure !

freeboub:
First implementation for android available here: https://github.com/react-native-video/react-native-video/pull/2806

ChaTho7:
> First implementation for android available here: #2806


I asked that https://github.com/react-native-video/react-native-video/issues/2617#issuecomment-1207268335 because i thought that we can customize the subtitle view (backgroundColor, textColor, fontSize, fontFamily, paddings, etc...) **completely** by adding a prop that request a view component. I'm curios about this feature if it's possible to do, **unless decrease the performance significantly**.

freeboub:
There is a misunderstanding. The feature you request starts with this PR https://github.com/react-native-video/react-native-video/pull/2759 i merged this afternoon :) but I agree, it needs to be enhanced to support more styling !

This ticket is about update of track list during playback. This is another topic.

freeboub:
This has been merged, will be available in 6.0.0-alpha.5

PR #2689, opened 2022-05-23
Crunchyroll Android Improvements

armands-malejevs:
This is a collection of improvements we've done for the Crunchyroll Android app:
* Support disabling buffering
* Fix AudioFocus bug that could cause the player to stop responding to play/pause in some instances.
* Fix player crashing when it is being cleared.
* Add support for customising back buffer duration and handle network errors gracefully to prevent releasing the player when network is lost.
* Allow player to be init before source is provided, and later update once a source is provided.
* Adds handling for providing a empty source in order to stop playback and clear out any existing content
* Add support for onBufferProgress prop on Android to get buffer data even when the player is paused.
* Add support for detecting if format is supported and exclude unsupported resolutions from auto quality selection and video track info in RN.
* Improve error handling
* Add support for L1 to L3 Widevine fallback if playback fails initially.
* Reduce buffer size based on available heap
* Force garbage collection when there is no available memory
* Improve memory usage
* Support disabling screen recording
* Improved error capturing
* Fix DRM init crashes
* Improve progress reporting
* Fix progress loss when network connection is regained

#### Testing
Either use the test apps in this project or link your own project to this PR and test everything related to playback.


android-exoplayer/src/main/java/com/brentvatne/exoplayer/VideoEventEmitter.java
```diff
@@ -206,6 +219,19 @@ void buffering(boolean isBuffering) {
         receiveEvent(EVENT_BUFFER, map);
     }
 
+    void playbackStateChanged(boolean isPlaying) {
+        WritableMap map = Arguments.createMap();
+        map.putBoolean(EVENT_PROP_IS_PLAYING, isPlaying);
```
freeboub, line 223: Not clear to me in terms of API ...
should we add an isPlaying event or should we use playback rate value update.
IsPlaying is more simple PlaybackRateChange is more flexible ...

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -388,116 +427,292 @@ private void reLayout(View view) {
...
+        DrmSessionManager drmSessionManager = null;
+        if (self.drmUUID != null) {
+            try {
+                drmSessionManager = self.buildDrmSessionManager(self.drmUUID, self.drmLicenseUrl,
+                        self.drmLicenseHeader);
+            } catch (UnsupportedDrmException e) {
+                int errorStringId = Util.SDK_INT < 18 ? R.string.error_drm_not_supported
+                        : (e.reason == UnsupportedDrmException.REASON_UNSUPPORTED_SCHEME
+                        ? R.string.error_drm_unsupported_scheme : R.string.error_drm_unknown);
+                eventEmitter.error(getResources().getString(errorStringId), e, "3003");
```
freeboub, line 589: This is a verry good thing to unify errors, but:
- I think you should remove the getResources().getString() from the api ...
   There is no reason to have translated error string coming from a library because UI will not be able to handle it depending of the platform language. 
   The translated value will never be alligned with customer requierments.

Please also add documentation of error codes, then it should be unified with ios / windows errors and this onError API will be good

freeboub: here is onError definition I use in my fork:

#### onError
Callback function that is called when the player fails due to an error.

Payload:

| Property                 | Type    | Description                                                 |
| ------------------------ | ------- | ----------------------------------------------------------- |
| code | string | String describing the error code |
| exception | string | String detailing the error (for debugging) |


| errorString                 | Type    |
| ------------------------ | ------- | ----------------------------------------------------------- |
| Unknown | unexpected error |
| DRM | DRM initialization issue |
| Decoder | Video decoder issue |
| Format | Format not supported |
| InvalidState | An invalid action has been requested |
| Canceled | Request has been canceled |
| Network | Network access issue |
| InvalidParamUrl | Source is not valid |
| InvalidParam | An invalid paramter has been provided |
| Timeout | Timeout during initialization |
| MaxConcurentStream | Maximum simultaneous stream count has been reach |
| Restricted | Content cannot be played due to restriction (geolocation for exemple) |
| DeviceNotInHousehold | Cannot play due to Account configuration issue |
| ContentNotFound | asset not found |
| Specific to WebView |
| Html5 | HTML5 player Error |
| VideoNotPlayableInEmbdedPlayer | content cannot be played in webview |
| WebViewNotAvailable | Webview not available on device |

Example:

```
{
  error:
  {
    errorString: 'Network',
    errorException:  'Yeah I had an error, that is bad',
  }
}
```

freeboub: I think this is enough to handle error on UI application

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -294,6 +302,26 @@ public void setDisableFocus(final ReactExoplayerView videoView, final boolean di
         videoView.setDisableFocus(disableFocus);
     }
 
+    @ReactProp(name = PROP_BACK_BUFFER_DURATION_MS, defaultInt = 0)
+    public void setBackBufferDurationMs(final ReactExoplayerView videoView, final int backBufferDurationMs) {
+        videoView.setBackBufferDurationMs(backBufferDurationMs);
+    }
+
+    @ReactProp(name = PROP_CONTENT_START_TIME, defaultInt = 0)
```
freeboub, line 310: Great, this is a very good feature! :) 
But I think we need to do a bigger change to have a safe API...

As you see there is an uggly postDelayed in initializePlayer.
I think we should addapt the APi to avoid this postdelay.

Its goal is to be able to receive all initialization props before really starting playback.

To avoid this there is a big api change to:
I really think we shall merge all props to startup a content in source object.

I identify: DRM config, sideloaded subtitles track, buffering parameters, useTextureView,  and this content start time !

Notice that I also implement a contentStartTime in my fork, This is a good feature !

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -388,116 +427,292 @@ private void reLayout(View view) {
...
+        this.bufferCheckTimer = new Timer();
+        TimerTask bufferCheckTimerTask = new TimerTask() {
+            @Override
+            public void run() {
+                if (mainHandler != null) {
+                    mainHandler.post(new Runnable() {
+                        public void run() {
+                            if (player != null) {
+                                double bufferedDuration = (double) (player.getBufferedPercentage() * player.getDuration() / 100);
+                                eventEmitter.bufferProgress(0d, bufferedDuration);
```
freeboub: I don't understand why you need this new event ?
The onProgress event should be enough.
It may just need documentation clarification

examples/basic/yarn.lock
```diff
@@ -2571,6 +2571,11 @@ emitter-listener@^1.0.1, emitter-listener@^1.1.1:
   dependencies:
     shimmer "^1.2.0"
 
+eme-encryption-scheme-polyfill@^2.0.1:
```
freeboub: I don't see any change in package.json, do you have an idea this this package has been added ?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -607,7 +823,6 @@ private void startPlayback() {
                 case Player.STATE_IDLE:
                 case Player.STATE_ENDED:
                     initializePlayer();
-                    break;
```
freeboub: are you sure you need to remove this break ?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -388,116 +427,292 @@ private void reLayout(View view) {
...
-                        keyRequestPropertiesArray[i + 1]);
+        try {
+            HttpMediaDrmCallback drmCallback = new HttpMediaDrmCallback(licenseUrl,
+                    buildHttpDataSourceFactory(false));
+            if (keyRequestPropertiesArray != null) {
+                for (int i = 0; i < keyRequestPropertiesArray.length - 1; i += 2) {
+                    drmCallback.setKeyRequestProperty(keyRequestPropertiesArray[i], keyRequestPropertiesArray[i + 1]);
+                }
+            }
+            FrameworkMediaDrm mediaDrm = FrameworkMediaDrm.newInstance(uuid);
```
freeboub, line 658: I understand the idea of this change, but I am not really confident with it...
I think the exoplayer DRM level chosen by exoplayer should be the good one.
Forcing to L3 should not fix any issue.

For exemple if your DMR request fails due to a network issue, it will fallback to L3 where it must be L1.

Is it a fix for a device specific issue ?
maybe it can be enabled on a <Video prop ? (not sure this is the beest idea)

freeboub (changes requested):

freeboub:
Hi,
Sorry for the lot of comment, but it is a good PR with new great features !
Thank you !
(and let's discuss on comments, not everything is mandatory)

android-exoplayer/src/main/java/com/brentvatne/exoplayer/VideoEventEmitter.java
freeboub: please also notice this ticket : https://github.com/react-native-video/react-native-video/issues/2684

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
cobarx: We've had issues where some older devices simply can't support L1, the Nexus 7 among others. A reasonable approach would be to have a prop that controls whether the L3 fallback is allowed or not.

IIRC, Armands already added retry logic for failed network requests.

armands-malejevs: This is a fix specifically for devices that identify as L1 supported, but for whatever reason their L1 certificate has been revoked (like Nexus 7) so Exoplayer fails to play on those devices because they appear "L1" but the certificate request if rejected so playback fails.

I can add a prop like, `minDRMLevel={"L1 | L3"}` which would be used to determine if fallback is necessary.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/VideoEventEmitter.java
armands-malejevs: In terms of API I think this makes sense because it is similar to how ExoPlayer handles/names this change, in this case we're only limited to is playing or not, but this can be expanded in the future to support all [player states](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/Player.State.html)

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: I think that this would be a great improvement separately, to refactor/improve the error reporting code, but for this PR I'd leave it as is, since it only adds new errors and I'd prefer not to make this PR have more refactors than it already has 😅

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
armands-malejevs: I 100% agree that we should do the refactor to get rid of the posDelayed in init, but I think this change would be outside of scope for this PR.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: This was meant to be expanded with the buffer start being reported exactly (right now it's always 0, but it should be read in to represent the actual start position of the buffer). This meant to be slightly different than `onProgress` which just shows the buffered duration, this currently doesn't track the back buffer length, but it needs to be expanded in the future to show the start and end of buffer so that you know exactly where is the buffered area when using back buffer, since the playable/seekable duration in onProgress event has the total time which doesn't really tell where the buffer starts or ends when using both forward and back buffer.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Looks like some merge conflict issue. Great catch! I added it back.

examples/basic/yarn.lock
armands-malejevs: Looks like some merge conflict issue, I removed it 👍

armands-malejevs:
@freeboub Thanks for the great feedback! 

I think that for most of the comments we should address them separately since this is already a large PR and it doesn't appear that the current state is inherently bad, but could be improved upon in general not only relating to changes in this PR.

I would recommend we create issues with all of the improvements not addressed in this PR and start getting feedback from the community on which of them is the highest priority and start implementing them one by one.

@cobarx @hueniverse What do you think about this approach?

nickfujita (approved):
Tested in the repo test harness and loaded in our application. All features appear to be working with the upstream changes merged as well.

hueniverse:
I agree we should only fix actual issues with the PR and defer all improvements to follow up work. Please open an issue for each comment and resolve it so we can keep track of what needs to happen.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/VideoEventEmitter.java
freeboub: No problem, ok for me, it's also part of mediasession PR

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
armands-malejevs: I created an issue for this https://github.com/react-native-video/react-native-video/issues/2693 feel free to comment any implementation details you feel could be useful to include in the issue.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Created an issue for this https://github.com/react-native-video/react-native-video/issues/2694

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: I created and issue for this https://github.com/react-native-video/react-native-video/issues/2695

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: I created an issue for this https://github.com/react-native-video/react-native-video/issues/2696

armands-malejevs:
@hueniverse I created issues for all addressable issues 👍 If anyone finds that I missed any, please let me know.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: OK, I understand you need additionnal informations on available buffer, but it should be added to onProgress event.
the new bufferProgress has 2 paramters: 0 Ok for futur use, so not needed now.
bufferedDuration which is exactly equals to bufferedDuration in onProgress.
The only difference is the type you use double instead of long.

Another difference is the frequency. If another issue for you is linked to the the missing report when playback is pause, I invite you to check following PR: https://github.com/react-native-video/react-native-video/pull/2664

For me this new event doesn't bring new infomation for application...

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -847,14 +1094,85 @@ private WritableArray getVideoTrackInfo() {
...
         }
+
         return videoTracks;
     }
 
+    private WritableArray getVideoTrackInfoFromManifest(Timeline timeline) {
+        return this.getVideoTrackInfoFromManifest(timeline, 0);
+    }
+
+    private WritableArray getVideoTrackInfoFromManifest(Timeline timelineRef, int retryCount) {
```
freeboub: I don't understand why you need a retry strategy to get Video Tracks ?
Is it related to loadManifest which can be long ?

freeboub: timeline is not used in the function, I think you remove it.
I agree that timeline handling shall be enhanced.
I can also point following ticket which will be needed: https://github.com/react-native-video/react-native-video/issues/2683

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -847,14 +1094,85 @@ private WritableArray getVideoTrackInfo() {
...
+            Timeline timeline = timelineRef;
+            long startTimeUs = startTime * 1000; // ms -> us
+
+            public WritableArray call() throws Exception {
+                WritableArray videoTracks = Arguments.createArray();
+                try  {
+                    DashManifest manifest = DashUtil.loadManifest(this.ds, this.uri);
+                    int periodCount = manifest.getPeriodCount();
+                    for (int i = 0; i < periodCount; i++) {
+                        Period period = manifest.getPeriod(i);
```
freeboub, line 1098: Is this a workaround for manifest with empty period ?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -847,14 +1094,85 @@ private WritableArray getVideoTrackInfo() {
...
+                            }
+                            boolean hasFoundContentPeriod = false;
+                            for (int representationIndex = 0; representationIndex < adaptation.representations.size(); representationIndex++) {
+                                Representation representation = adaptation.representations.get(representationIndex);
+                                Format format = representation.format;
+                                if (representation.presentationTimeOffsetUs <= startTimeUs) {
+                                    break;
+                                }
+                                hasFoundContentPeriod = true;
+                                WritableMap videoTrack = Arguments.createMap();
```
freeboub, line 1112: can you avoid creating WritableMap is !isFormatSupported please ?
Additioannal question, can this can be factorized with getVideoTrackInfo ?

freeboub (changes requested):

README.md
```diff
@@ -8,6 +8,1639 @@ Version 4.x requires react-native >= 0.57.0
...
+Platforms: all
+
+#### onPlaybackStateChanged
+Callback function that is called when the playback state changes.
+
+Payload:
+
+Property | Description
+--- | ---
+isPlaying | boolean | Boolean indicating if the media is playing or not
```
freeboub, line 1186: There is a very big diff in README.md, it make the change very difficult to review.
Any idea why ? (maybe a formating issue ?)

freeboub (changes requested):

README.md
hueniverse: That's my fault. I moved the content of the README to API. I will fix that once the PR is merged.

cobarx:
Agree with @armadilio3's suggestion that we record the improvements and take them up one by one. Our team has limited capacity to devote to changes, so picking them up incrementally would be much better.

gabriel-rivero (approved):
double checked the changes in test-harness, features are working with no issues

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: It can take long and I think that the manifest request can fail for whatever reason so we added a retry to attempt another load if it was an intermittent network issue etc.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Not sure what you mean by this? In each period we look for the video tracks and do checks like are they after the content start or supported by the device.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: What do you mean by avoiding WritableMap here? We are returning the video tracks back to JS in a WritableArray which I think needs WritableMap as elements if they are an object in JS.
Also what do you mean by `factorized with getVideoTrackInfo ?`

armands-malejevs:
@freeboub I cleaned up some stuff 👍 I had some questions on some of your comments, let me know how you want to handle those

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
cobarx: Let's add a comment noting this.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
cobarx: Once you understand the question, let's add a comment noting this.

README.md
freeboub: @hueniverse did you check this point ?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Other points are critical for me, but I really think this event shall be removed as it is duplicated from onProgress
It brings complexity are will make hard to maintain...
Can you please remove it and use onProgress in your app ? 
If there is issue or missing information in onProgress, let's fix it !

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Ok, DashUtil.loadManifest will download a new manifest file.
But why not using onTracksChange event ? 
This new behavior will be dash specific and it triggers not mandatory manifest download ? 
A custom manifest parser is not better ?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Just move allocation of videotracks inside   if (isFormatSupported(format)) { 
To avoid unnecessary operation.  A simple point in fact :)

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: I see that we added this so that we get buffer progress events even when paused. On progress doesn't seem like the best place to put this interval based reporting. I'm not sure if we want to keep receiving progress events when buffering. I feel that having the buffer quickly accessible on progress if needed is a nice to have, but having an independent buffer reporting that isn't tied to video progress I think is best. Does it make sense why this was separated out?

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Wouldn't `onTracksChange` give us the available resolutions only when content starts playing since the tracks only then become available? This was meant to show the user content resolutions even while ads are playing.

Yes, this is DASH specific, but we can expand this in the future. I updated the code to only download dash manifest if contentStartTime is set so it will no longer do unnecessary requests.

We could implement a custom manifest parser but ExoPlayer already has one so we decided to go with that since it had everything we needed to implement this feature.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Lol, didn't notice that, this was really dumb 🤦 😆 Fixed it 👍

armands-malejevs:
@freeboub Can you take another look? I fixed/responded to some of the feedback

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: ok, you can overload the manifest parser. anyway tjis is ok, thanks

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
lvffilho: @armadilio3 I have the same issue that you. I need to know if video is real buffering. There are some cases that we loose connection and it stop onProgress and start buffering, but i dont know if player is real buffering, because only the onProgress event show me that. Than it keep trying reconnect always and dont throw an error.

README.md
hueniverse: Just ignore it. I will fix it after the merge.

freeboub:
PR Tested on sample provided in exemple/basic:
I have following crash when playing local file (require('./broadchurch.mp4'))

[Fri Jun 03 2022 18:39:31.221]  LOG      {"error":{"errorStackTrace":"com.google.android.exoplayer2.ExoPlaybackException: Unexpected runtime error\n\tat com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:587)\n\tat android.os.Handler.dispatchMessage(Handler.java:102)\n\tat android.os.Looper.loop(Looper.java:223)\n\tat android.os.HandlerThread.run(HandlerThread.java:67)\nCaused by: java.lang.IllegalStateException\n\tat com.google.android.exoplayer2.util.Assertions.checkState(Assertions.java:86)\n\tat com.google.android.exoplayer2.source.ProgressiveMediaPeriod.selectTracks(ProgressiveMediaPeriod.java:281)\n\tat com.google.android.exoplayer2.source.MaskingMediaPeriod.selectTracks(MaskingMediaPeriod.java:186)\n\tat com.google.android.exoplayer2.MediaPeriodHolder.applyTrackSelection(MediaPeriodHolder.java:292)\n\tat com.google.android.exoplayer2.ExoPlayerImplInternal.reselectTracksInternal(ExoPlayerImplInternal.java:1643)\n\tat com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:503)\n\t... 3 more\n","errorCode":"2001","errorException":"com.google.android.exoplayer2.ExoPlaybackException: Unexpected runtime error","errorString":"ExoPlaybackException type : 2"}}


Please notice also that after the error occurs, onBufferProgress is still sent in loop


Edit: it is reproduced with all mp4 content
Edit2: In this PR I also add m3u and live streams samples it would be intersting to test with these new streams.

Thanks

armands-malejevs:
@freeboub I fixed single track playback. Can you test again?

freeboub:
> @freeboub I fixed single track playback. Can you test again?

I confirm this is fixed, thanks

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: @armadilio3 This is the last point to handle I think.
I still don't understand why you need to duplicate this event as information is already available in onProgress event.
once again there is a bug in onProgress which is blocking even when player is paused, it has to be solved at least for live playback.
Duplicating event can overload the bridge, so having too much event is a bad idea.

The advatage of onProgress is that timeout can already be configured.
I also think having 1 event with multiple parameters is lighter than have 2 events in terms of CPU consumption

freeboub (changes requested):

armands-malejevs:
@freeboub I'm still not sold on having the an independent value buffer tied to on progress events which seems non intuitive if the two are unrelated to one another, but to get this PR moving forward I removed it. We can open a separate PR to properly implement the feature and discuss separating or leaving in the progress event.

freeboub:
Thank you @armadilio3 !
In the same time, can you please review this PR and check if you are now well receiving the event even if player is paused or have connectivity issue : https://github.com/react-native-video/react-native-video/pull/2664

freeboub (approved):

EyMaddis:
> Force garbage collection when there is no available memory

I think this might be the culprit why our FireTV Sticks starting having frame drops, this works on 5.x versions of react-native-video

PR #2708, opened 2022-06-03
Video caching on Android

jaspermeijaard:
## Description

Video caching might be great for users who play the same video multiple times or would like to play that same video offline. Currently `react-native-video` doesn't provide a way to cache videos on Android. There has been a few changes to ExoPlayer which makes video caching simpler than before. This PR intends to add caching in a simple and effective way. 

These changes wrap the default ExoPlayer `DataSource` in to a `CacheDataSource`. When there's no cache configured, the default data source is used and not wrapped. If cache is enabled, the video will be retrieved from the default DataSoure on a cache MISS. On a cache HIT, the video is retrieved from the file system.

## PR todos

- [x] Update the documentation
- [x] Update the changelog
- [x] Provide an example of how to test the change
- [x] Focus the PR on only one area
- [x] Describe the changes

## How to test

1. Deploy debug app to real device or emulator
2. Connect to instance and open shell by running command `adb shell`
3. Open app environment by running command in opened shell `run-as <package-name>`
4. Check cache size `du -h cache/videos` and/or see that's empty
5. Initialize video player with cache configuration
```
<Video
    {...}
    cache={{
        dir: "videos",
        maxSizeBytes: 100 * 1024 * 1024,
    }}
/>
```
6. Play a video with that same player
7. Check again cache size `du -h cache/videos` and validate that it contains more data
8. Play more videos until 100 MB data is cached
9. Check again cache size `du -h cache/videos` and validate that it does not contain more than 100MB data


android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -333,6 +337,18 @@ public void setBufferConfig(final ReactExoplayerView videoView, @Nullable Readab
         }
     }
 
+    @ReactProp(name = PROP_CACHE)
+    public void setCache(final ReactExoplayerView videoView, @Nullable ReadableMap cacheConfig) {
```
freeboub: I think there are missing else in that case.
If user/UI enable caching, it will not be able to disable it.
Please allow to disable caching.

Video.js
```diff
@@ -481,6 +481,10 @@ Video.propTypes = {
   progressUpdateInterval: PropTypes.number,
   useTextureView: PropTypes.bool,
   hideShutterView: PropTypes.bool,
+  cache: PropTypes.shape({
+    dir: PropTypes.string,
```
freeboub, line 496: it would be usefull to allow some 'default' directory (just resolve /cache dir or /data/data/pkg/).
finding these path may be tricky in react.
Just a proposal, it is not mandatory to merge this Pr

API.md
```diff
@@ -424,6 +425,24 @@ bufferConfig={{
 
 Platforms: Android ExoPlayer
 
+#### cache
```
freeboub, line 406: Juste few questions:
- will it work with all types of content (mp4 / dash / hls, ... ) ? (and/or what did you test ?)
- if we play local file, will it try to use cache ?
- what about live video url ? can it be cached ? do it really make sens ?

Thanks

freeboub (changes requested):

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
jaspermeijaard: Nice one. That's something what should be in there yes. Added it in this commit https://github.com/react-native-video/react-native-video/commit/fe426b2840e715eaff9560a82dbf3e1680e00977.

Video.js
jaspermeijaard: Yes that's a nice one. I set up some defaults in this commit: https://github.com/react-native-video/react-native-video/commit/fe426b2840e715eaff9560a82dbf3e1680e00977:
- When no cache dir is given, use subdirectory "videos" inside app cache directory provided by context. 
- When no cache max size is given, use a default of 100 MB.

API.md
jaspermeijaard: I have to dig in to this a bit more. Currently I only tested playback and caching of MP4's served from a S3 bucket. I'll come back on this later.

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
jaspermeijaard: Okay today I changed a few other things after more testing. ExoPlayer will throw errors if the a new SimpleCache instance is initialized with a directory which was used before and isn't released. That's fixed and should play nice together with the lifecycle of setting react component props.

Video.js
jaspermeijaard: I made some changes so that the cache directory can be changed on the fly in JS. And also to allow fully disabling of the cache. The code relies on @ReactProp being invoked but if you remove the prop in JS, that method is never invoked. In order to allow disabling (when being enabled), `cache={null}` should be provided. If there's a better way please let me know.

hueniverse:
@jaspermeijaard Thank you for all the work on this. What's the current status? Are all issues mentioned by @freeboub resolved? Also can you rebase?

freeboub:
@jaspermeijaard can you update the PR with master please ? notice that we now use last exoplayer release !
thanks

jaspermeijaard:
@hueniverse @freeboub The branch is updated with master and compiles and runs. In general most feedback is processed and I think the react/native API is in a good state now.

On version 5 caching fully worked for playback of MP4's. After that I tested dash and hls, the files were actually stored in cache but they were not consumed from cache again by the player which is an issue. A quick test run on version 6 shows that all formats are stored in cache but none of them are consumed again from cache :( 

That's a major issue which needs to be solved. I have no clue yet why Exoplayer writes the cache but doesn't read from it again. I think the same instance is used and enough data is written to be reused. It might [relate to this issue](https://github.com/google/ExoPlayer/issues/9936). Happy to hear your thoughts.

PR #2511, opened 2021-10-15
Fix inaccurate seeking.

hossein-zare:

PR #2712, opened 2022-06-07
updated ReactVideoView.java to fix seeking issue on android

monahans:
This PR fixes a bug when trying to seek in the Video component on Android. Before, the video would skip to the beginning of the video if the user tried to seek, but now it works as expected.

To test, simply use the video component on android, and you are now able to seek.


freeboub:
@monahans can you please use the constant ? "3" seems strange here, nobody will understand.
As sample there is already:  
https://github.com/react-native-video/react-native-video/pull/2555/files
https://github.com/react-native-video/react-native-video/pull/2511/files

freeboub (changes requested):
please use enum / constant: SeekParameters.CLOSEST_SYNC I think

monahans:
Updated to use MediaPlayer.SEEK_CLOSEST

freeboub (approved):

khushal87:
When will the PR be in the latest beta release?

PR #2718, opened 2022-06-13
Replace Image.propTypes with ImagePropTypes.

### LINKED ISSUE #2714: ERROR  TypeError: undefined is not an object (evaluating '_reactNative.Image.propTypes.resizeMode')
# Bug

## Platform

* iOS

## Environment info

React native info output:

```bash
 ERROR  TypeError: undefined is not an object (evaluating '_reactNative.Image.propTypes.resizeMode')
```

Library version: `react-native@0.69.0-rc.6`, `react@18.1.0` and `react-native-video@^5.2.0`

## Steps To Reproduce

Trying out all the latest stuff and I get the error shown above. I'm using nx & react-native without expo. The following code seems to be the culprit:
https://github.com/react-native-video/react-native-video/blob/master/Video.js#L429

devon-whil:
Fixes: https://github.com/react-native-video/react-native-video/issues/2714

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { StyleSheet, requireNativeComponent, NativeModules, UIManager, View, Image, Platform, findNodeHandle } from 'react-native';
-import { ViewPropTypes } from 'deprecated-react-native-prop-types';
+import { ViewPropTypes, ImagePropTypes } from 'deprecated-react-native-prop-types';
 import resolveAssetSource from 'react-native/Libraries/Image/resolveAssetSource';
 import TextTrackType from './TextTrackType';
 import FilterType from './FilterType';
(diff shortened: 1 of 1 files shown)


devon-whil:
I am also getting the following error with `react-native@0.69.0-rc.6`:
```
▸ Compiling RCTVideoManager.swift

❌  /Users/devon/projects/frontend/node_modules/react-native-video/ios/Video/RCTVideoManager.swift:8:49: cannot convert value of type 'RCTEventDispatcherProtocol?' to expected argument type 'RCTEventDispatcher?'
```
This error goes away when I change the type as it suggested. Seems this change would break support for older versions of react-native. @freeboub Should that change be part of this PR?

freeboub (approved):
tested on sample, looks good

freeboub:
> I am also getting the following error with `react-native@0.69.0-rc.6`:
> 
> ```
> ▸ Compiling RCTVideoManager.swift
> 
> ❌  /Users/devon/projects/frontend/node_modules/react-native-video/ios/Video/RCTVideoManager.swift:8:49: cannot convert value of type 'RCTEventDispatcherProtocol?' to expected argument type 'RCTEventDispatcher?'
> ```
> 
> This error goes away when I change the type as it suggested. Seems this change would break support for older versions of react-native. @freeboub Should that change be part of this PR?

Thank @devon-whil, I think it is better to keep separate change in small Prs, it is really easier to review !

PR #2730, opened 2022-06-20
fix: keep playback state while seeking when video is ended on android

matinzd:
# What is the problem?
Video will stay paused when `repeat={false}` and video is ended. When user tries to seek, the playback state will remain on pause.
After seeking back, the video should start playing without changing the `paused` prop, since this is the expected behavior. iOS is behaving the same way.

Screen record of the issue on Android:
https://user-images.githubusercontent.com/24797481/174677945-18a47430-9c5c-4216-a217-1d267d562b51.mov

# What's changed?

- Added a new method `notifyPlaybackState`
- Call `notifyPlaybackState` when seeking

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1640,7 +1640,7 @@ class ReactExoplayerView extends FrameLayout implements
 
     public void setPausedModifier(boolean paused) {
         isPaused = paused;
-        notifyPlaybackState()
+        notifyPlaybackState();
     }
 
     public void setMutedModifier(boolean muted) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1627,18 +1627,22 @@ public void setSelectedTextTrack(String type, Dynamic value) {
...
                 startPlayback();
             } else {
                 pausePlayback();
             }
         }
     }
 
+    public void setPausedModifier(boolean paused) {
+        isPaused = paused;
+        notifyPlaybackState()
```
armands-malejevs: Needs `;`

matinzd: Fixed.

armands-malejevs (approved):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
armands-malejevs: Thanks!

matinzd:
I am closing this as it seems that there isn't any issue with the master branch version and this bug is happening on the older version with android-exoplayer folder.

Issue #2510, opened 2021-10-15
[Android] Expose acquireDrmInfo from AndroidDRM framework


SabatinoMasala:
# Feature Request

If we have access to the result of [acquireDrmInfo](https://support.google.com/widevine/answer/6072714?hl=en) we can determine the Android device security levels in Javascript land.

## Why it is needed

If you have access to the device security levels, you could request different streams based on this info. Eg. an HD stream for L1 and an SD stream for L3.

### Example usage

```javascript
import { acquireDrmInfo, DEVICE_IS_PROVISIONED, DEVICE_IS_NOT_PROVISIONED, DEVICE_IS_PROVISIONED_SD_ONLY } from 'react-native-video'

const drmInfo = await acquireDrmInfo()
const WVDrmInfoRequestStatusKey = drmInfo.WVDrmInfoRequestStatusKey

if (WVDrmInfoRequestStatusKey === DEVICE_IS_NOT_PROVISIONED) {
  console.error('Hardware / integration error')
}

if (WVDrmInfoRequestStatusKey === DEVICE_IS_PROVISIONED) {
  // Fetch L1 (hd) stream
} else {
  // Fetch L3 (sd) stream
}

```

freeboub:
I develop a similar feature. I will try to create a pull request with it

freeboub:
Implemented on 6.0.0-alpha.2

PR #2740, opened 2022-06-23
Feature/add api to retrieve decoder capabilities

freeboub:
New feature to fix following issue: https://github.com/react-native-video/react-native-video/issues/2510

The goal of this PR is to allow application to query device capabilities.
Following api was added:

    getWidevineLevel => integer between 0 and 3 (not sure 2 make sense :/)
    isCodecSupported(codec, width, height) => boolean
    isHEVCSupported() => boolean

Api.md has been updated
basic sample has been updated to test this feature


android/src/main/java/com/brentvatne/react/VideoDecoderPropertiesModule.java
```diff
@@ -0,0 +1,105 @@
...
+            MediaFormat format = MediaFormat.createVideoFormat(mimeType, width, height);
+            String codecName = mRegularCodecs.findDecoderForFormat(format);
+            if (codecName == null) {
+                p.resolve(false);
+            } else {
+                p.resolve(true);
+            }
+        } else {
+            p.resolve(false);
+        }
```
rafaelbrun: Can we make this an early return?
```java
if (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.LOLLIPOP) {
   p.resolve(false);
   return;
}
...
```

rafaelbrun (approved):

rafaelbrun:
Tested on my project and it worked, I was able to get all decoder properties, great work!!

android/src/main/java/com/brentvatne/react/VideoDecoderPropertiesModule.java
freeboub: Yes it makes sense, I will do it

Issue #2690, opened 2022-05-23
onEnd called twice when paused prop changes at the end of the media


ldabiralai:
# Bug

When using ExoPlayer, the onEnd prop is called for a second time when the paused prop changes in the `onEnd` first callback execution.

When the paused prop is updated using Android MediaPlayer the `onEnd` callback is not called for a second time.

## Platform
* Android ExoPlayer

## Environment info

React native info output:

```bash
System:
    OS: macOS 12.3.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.85 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.18.2 - ~/.volta/tools/image/node/14.18.2/bin/node
    Yarn: 1.22.18 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
    npm: 6.14.17 - ~/.volta/tools/image/npm/6.14.17/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: Not Found
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 23, 29, 30, 31, 32
      Build Tools: 28.0.2, 28.0.3, 30.0.2, 31.0.0, 32.0.0
      System Images: android-28 | Google APIs Intel x86 Atom_64, android-30 | Google APIs Intel x86 Atom, android-30 | Google APIs Intel x86 Atom_64, android-30 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8139111
    Xcode: 13.4/13F17a - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.12 - /usr/local/opt/openjdk@11/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.2 => 0.68.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 5.2.0

## Steps To Reproduce

[...]

PR #2741, opened 2022-06-23
fix(android): ensure we don't disable playback when player is ENDED

freeboub:
fix issue #2690

ensure we don't pause playback when player is ended.
In this use case, Exoplayer send a EVENT_PLAYBACK_STATE_CHANGED event, then onEnd JS event is triggered a second time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Version 6.0.0-alpha.2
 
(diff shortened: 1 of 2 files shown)


rafaelbrun:
Tested and it's fixed

rafaelbrun (approved):

paulohc (approved):
Tested here and looks fine

PR #1590, opened 2019-05-17
[Android] Font Size and padding Botton Subtitles

allinelara:
Fix issue [#1335](https://github.com/react-native-community/react-native-video/issues/1335)

**Font size**

Before the font size of the subtitles was set by default for Android and the default size was really large and looks weird in some devices. It was no possible to set the font size via JavaScript or editing the VTT file. Now with this changes it is possible to set a font size via JavaScrip using the props: fontSizeTrack.  

**Padding botton** 

It seems that in some devices the subtitles were hidden, so with the props paddingBottonTrack is possible to set via JavaScript a padding botton to position better the subtitles in the screen.

PR #2759, opened 2022-07-08
feat: add support of subtitles style configuration

freeboub:
Fix issue https://github.com/react-native-community/react-native-video/issues/1335
+ replace replace following PR: https://github.com/react-native-video/react-native-video/pull/1590

Add subtileStyle prop with following sub properties: 

Font size

Before the font size of the subtitles was set by default for Android and the default size was really large and looks weird in some devices. It was no possible to set the font size via JavaScript or editing the VTT file. Now with this changes it is possible to set a font size via JavaScript using the props: fontSizeTrack.

Padding

It seems that in some devices the subtitles were hidden, so with the props paddingBotton is possible to set via JavaScript a padding botton to position better the subtitles in the screen.
props: paddingBottom, paddingTop, paddingLeft and paddingRight
This padding is applyed only to the subtitleView

### DIFF
--- a/API.md
+++ b/API.md
@@ -304,6 +304,7 @@ var styles = StyleSheet.create({
 |[selectedTextTrack](#selectedtexttrack)|Android, iOS|
 |[selectedVideoTrack](#selectedvideotrack)|Android|
 |[source](#source)|All|
+|[subtitleStyle](#subtitleStyle)|Android|
 |[textTracks](#texttracks)|Android, iOS|
 |[trackId](#trackId)|Android|
 |[useTextureView](#usetextureview)|Android|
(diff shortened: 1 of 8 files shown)


rafaelbrun:
Working great!
Awesome job, this is helpful

rafaelbrun (approved):

dr-thca:
Has anyone looked at the possibility of doing padding on ios as well? Do you think its possible from a technical standpoint?

freeboub:
> Has anyone looked at the possibility of doing padding on ios as well? Do you think its possible from a technical standpoint?

No, I don't think anyone checked... I don't know if it is possible...

PR #2806, opened 2022-08-06
Feature(android): new events on tracks changed

freeboub:
Add new events when tracks changed.
Audio, video, text tracks are supported.
This featue is usefull when trcks change during playback, especially for SSAI

Documentation and sample updated.

In the same time I fixed default value from contentStartTime & add support of square video on android.


freeboub:
@bulkinav I saw you like but a review/approuval would be better :pray:

PR #2527, opened 2021-11-04
iOS Swift Conversion

nickfujita:
This PR is a conversion of the iOS implementation from Objective-c to Swift.

# Goals of conversion
- Make the iOS codebase more accessible to developers coming from JS side to debug and contribute to the iOS side
- Cleanup codebase by utilizing more modern code patterns in Swift
- Prep codebase for unit tests by starting to modularize logic into smaller functions
- Improve code navigation by organize code into smaller files, grouping similar functions

# During the conversion process some updates to the code structure were also made
- Modularize codebase from single file to smaller focused files
- Untangled large nested IF statements
- Added more null checks, since Swift is more strict with null pointers
- Added property to allow for decoding of local video sources with self contained key for offline playback
- Updates example apps to test react-native 0.63.4 and uses auto native dependency imports for android and ios

# Test Instructions

If you have an existing project, please give this branch a try and reply with any issues, and will be happy to have a look.

## iOS project update
There is a breaking change in terms of setup, where you will need to add support for static dependency linking in the Podfile in your ios project. For details, please see the example ios project: https://github.com/crunchyroll/react-native-video/blob/ios-swift-conversion/examples/basic/ios/Podfile#L5

## Testing in updated basic example apps

```
yarn xbasic install
```

### iOS Example
```
yarn xbasic ios
```

### Android Example
```
yarn xbasic android
```

### Windows Example
```
yarn xbasic windows
```

PR #2845, opened 2022-09-05
chore: migrate project examples to RNTA

matinzd:
# Problem
As we are migrating to the new architecture, upgrading examples app is the important part of the changes. Without having a proper test app we are unable to test new features and migrate to the new architecture.

# Solution
Move all of the existing outdated example apps to React Native Test App. 
React Native Test App (RNTA) provides test apps for all platforms as a package. It handles the native bits for you so you can focus on what's important: your product.

https://github.com/microsoft/react-native-test-app


matinzd:
/cc @freeboub @cobarx @hueniverse @nickfujita 
Not sure who to tag.

package.json
```diff
@@ -12,9 +12,10 @@
     "devDependencies": {
         "@react-native-community/eslint-config": "^0.0.5",
         "eslint": "^6.5.1",
-        "react": "16.9.0",
-        "react-native": "0.61.5",
-        "react-native-windows": "^0.61.0-0"
+        "react": "18.0.0",
+        "react-native": "0.69.5",
+        "react-native-test-app": "^1.6.13",
```
freeboub, line 17: I don't think you need this react-native-test-app as devDependency here ?

freeboub:
On android TV emulator, If I select  EmbedFullScreen, I have following error:

 ERROR  Invariant Violation: "EmbedFullScreen" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called., js engine: hermes

Basic sample works fine!

freeboub:
Additionnally, can you also add a line the changeLog ?

freeboub:
Last point the sample app package name has been renamed, not a big issue for me, but it is com.microsoft.reacttestapp, It would be better to keep either the existing one or something like com.react.video.sample
But thank you for react 0.69 support !

package.json
matinzd: Removed under [a0cecca](https://github.com/react-native-video/react-native-video/pull/2845/commits/a0ceccaad8ff1a06c9b67757d3c312bddc1c38e0).

matinzd: > 63538df

Package names are now fixed [63538df](https://github.com/react-native-video/react-native-video/pull/2845/commits/63538df3f296bcefbe8c29247c9311459a04c89c).

matinzd:
> Last point the sample app package name has been renamed, not a big issue for me, but it is com.microsoft.reacttestapp, It would be better to keep either the existing one or something like com.react.video.sample But thank you for react 0.69 support !

Package names are now fixed [63538df](https://github.com/react-native-video/react-native-video/pull/2845/commits/63538df3f296bcefbe8c29247c9311459a04c89c).

matinzd:
The only problem I found and I believe that it might have happened during swift conversion (#2527) is that saving the video is not working in video caching example. 

```
 WARN  Possible Unhandled Promise Rejection (id: 0):
Error: Could not export video
promiseMethodWrapper@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.react.video.sample:22948:45
```

@freeboub @nickfujita  

Not sure if we had video caching before swift conversion.

package.json
freeboub: I think you did a mistake here, you removed react-native-test-app from exemple/basic/package.json but it is need here
And let it in root package.json where it is not needed.
(Then currently the sample doesn't build ... )

matinzd: Oops. I fixed it now.

freeboub:
Hello @matinzd,
It tooks me long time, but I back with this PR :)
This RNTA integration is usefull for simple tests and it make sense, but not for a 'full testing integration.
Here is an exemple, there is no easy way to enable / disable newArchitecture. Another exemple, would be the RTSP support, enabling it would need gradle change to include a new exoplayer plugin, but I don't think this is good to enable it by default, so I would like to configure it for gradle file, but it will be tricky do it with the sample.

So, can you rework the patch by doing this:
- restore basic sample and rename it to **full**
- move this full integrated sample (and the subfolder) to a new folder **basic**.

Is it Ok for you ?

matinzd:
> Hello @matinzd, It tooks me long time, but I back with this PR :) This RNTA integration is usefull for simple tests and it make sense, but not for a 'full testing integration. Here is an exemple, there is no easy way to enable / disable newArchitecture. Another exemple, would be the RTSP support, enabling it would need gradle change to include a new exoplayer plugin, but I don't think this is good to enable it by default, so I would like to configure it for gradle file, but it will be tricky do it with the sample.
> 
> So, can you rework the patch by doing this:
> 
> * restore basic sample and rename it to **full**
> * move this full integrated sample (and the subfolder) to a new folder **basic**.
> 
> Is it Ok for you ?

That makes sense!  I will update basic sample project without RNTA to the latest version and I will keep these changes with RNTA inside basic folder.

PR #2855, opened 2022-09-14
fix: fix the system status bar and touch twice to show control after going back from fullscreen

wood1986:
@freeboub My apologies, I mis-understood our previous conversation and i did not test it carefully

Current behaviour on master. The system status bar does not show anything and you need to touch twice before you can see the player control after going back from full screen
Kapture 2022-09-13 at 23 09 56

Expected behaviour
Kapture 2022-09-13 at 23 05 56


wood1986:
Not sure if it is possible to publish alpha.3

freeboub (approved):

freeboub:
I think it is possible :)

PR #2857, opened 2022-09-14
Fix/ensure player stop on invalid url

### LINKED ISSUE #2260: onError is not working Android
# Bug

I am developing a feature in my project for download a video when it fail. For that logic, I use onError callback in react-native-video for do something when it fail. In IOS it working fine, but in Android onError is not called but the` <Video />` don't play anything. Basically, onError event on Android is not working.

## Platform

Android 

(I am not sure about this)

* Android ExoPlayer
* Android MediaPlayer

## Environment info

System:
    OS: macOS 11.0.1
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 190.20 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.16.2 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.9.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.3, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.0
      System Images: android-29 | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6392135
    Xcode: 12.3/12C33 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_271 - /Library/Java/JavaVirtualMachines/jdk1.8.0_271.jdk/Contents/Home/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.3 => 0.63.3 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

Library version: 5.1.0-alpha8

## Steps To Reproduc

1. Give to <Video /> an incorrect source on Android
2. Set a log at onError prop
3. Run it and verify if onError is executed

## Expected behaviour

1. onError must be called with a parameter with the error
2. `<Video />` will no reproduce anything

## Reproducible sample code

`
          <Video
            source={{uri: "video.mp4"}}
            onError={(error) => console.log(error)}
          />
`

freeboub:
fix: https://github.com/react-native-video/react-native-video/issues/2260

sample updated to test the change

This PR allows to stop playback when invalid url is provided

#### Describe the changes
- Allow to pass invalid url to player implementation
- send early error on JS part
- ensure exoplayer is stopped on native side

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -285,7 +285,7 @@ export default class Video extends Component {
     const isNetwork = !!(uri && uri.match(/^https?:/i));
     const isAsset = !!(uri && uri.match(/^(assets-library|ph|ipod-library|file|content|ms-appx|ms-appdata):/i));
 
-    if (uri && !isNetwork && !isAsset) {
+    if ((uri || uri === '') && !isNetwork && !isAsset) {
       if (this.props.onError) {
         this.props.onError({error: {errorString: 'invalid url, player will stop', errorCode: 'INVALID_URL'}});
       }
--- a/Video.js
+++ b/Video.js
@@ -276,16 +276,20 @@ export default class Video extends Component {
     let uri = source.uri || '';
     if (uri && uri.match(/^\//)) {
       uri = `file://${uri}`;
-    } else if (uri === '') {
-      return null;
     }
 
     if (!uri) {
-      console.warn('Trying to load empty source.');
+      console.log('Trying to load empty source.');
     }
 
-    const isNetwork = !!(uri && uri.match(/^https?:/));
-    const isAsset = !!(uri && uri.match(/^(assets-library|ph|ipod-library|file|content|ms-appx|ms-appdata):/));
+    const isNetwork = !!(uri && uri.match(/^https?:/i));
+    const isAsset = !!(uri && uri.match(/^(assets-library|ph|ipod-library|file|content|ms-appx|ms-appdata):/i));
+
+    if (uri && !isNetwork && !isAsset) {
+      if (this.props.onError) {
+        this.props.onError({error: {errorString: 'invalid url, player will stop', errorCode: 'INVALID_URL'}});
+      }
+    }
 
     let nativeResizeMode;
     const RCTVideoInstance = this.getViewManagerConfig('RCTVideo');
--- a/examples/basic/src/VideoPlayer.android.tsx
+++ b/examples/basic/src/VideoPlayer.android.tsx
@@ -52,7 +52,7 @@ class VideoPlayer extends Component {
     require('./broadchurch.mp4'),
     {
       description: '(dash) sintel subtitles',
-      uri: 'Https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
+      uri: 'https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
     },
     {
       description: '(mp4) big buck bunny',
--- a/examples/basic/src/VideoPlayer.android.tsx
+++ b/examples/basic/src/VideoPlayer.android.tsx
@@ -51,7 +51,7 @@ class VideoPlayer extends Component {
     require('./broadchurch.mp4'),
     {
       description: '(dash) sintel subtitles',
-      uri: 'https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
+      uri: 'Https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
     },
     {
       description: '(mp4) big buck bunny',
@@ -65,6 +65,12 @@ class VideoPlayer extends Component {
       description: '(mp4|subtitles) demo with sintel Subtitles',
       uri:
         'http://www.youtube.com/api/manifest/dash/id/bf5bb2419360daf1/source/youtube?as=fmp4_audio_clear,fmp4_sd_hd_clear&sparams=ip,ipbits,expire,source,id,as&ip=0.0.0.0&ipbits=0&expire=19000000000&signature=51AF5F39AB0CEC3E5497CD9C900EBFEAECCCB5C7.8506521BFC350652163895D4C26DEE124209AA9E&key=ik0',
+      type: 'mpd',
+    },
+    {
+      description: 'invalid URL',
+      uri:
+        'mmt://www.youtube.com',
       type: 'mpd',
     },
     { description: '(no url) Stopped playback', uri: undefined },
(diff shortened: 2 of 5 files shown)


examples/basic/src/VideoPlayer.android.tsx
```diff
@@ -51,7 +51,7 @@ class VideoPlayer extends Component {
     require('./broadchurch.mp4'),
     {
       description: '(dash) sintel subtitles',
-      uri: 'https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
+      uri: 'Https://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
```
Gta-Cool: Incorrect schemes shouldn't be promoted (mixed upper and lower cases).

Although schemes are case-insensitive, the canonical form is lowercase and the implementation should accept uppercase letters as equivalent to lowercase in scheme names (e.g., allow "HTTP" as well as "http") for the sake of robustness.

See:
- https://www.rfc-editor.org/rfc/rfc3986#section-3.1
- https://www.rfc-editor.org/rfc/rfc1738#section-2.1

evoactivity: This change ensures `startsWithValidScheme` converts the url to lowercase. To make sure it satisfies
> the implementation should accept uppercase letters as equivalent to lowercase in scheme names (e.g., allow "HTTP" as well as "http") for the sake of robustness.

I don't see how this is promoting using incorrect schemes at all?

evoactivity (approved):
LGTM 👍

examples/basic/src/VideoPlayer.android.tsx
freeboub: The goal is not to promote bad url, but to be able to test it easily

freeboub:
> LGTM +1

@evoactivity Just a question, I am not able to test ios build for now, did you test the behavior with ios ?
Prior to this change, setting an invalid url was keeping current playback running. With the change it should stop current playback. I am afraid of ios stability with this change ....
Thank you for the review :+1:

evoactivity:
I only saw java files so only tried android. I have just tested iOS. 
When I set the source to `null` or `''` I see the console.log `Trying to load empty source.`. The video disappears but the sound keeps going.

I think we can add the same checks from android here https://github.com/react-native-video/react-native-video/blob/66b2ba3fcfd4efb7863530da76950ba397ebde85/ios/Video/RCTVideo.swift#L220

evoactivity:
I have the beginning of a fix that will stop playback on iOS.

Add this before `removePlayerLayer()` in  `func setSrc(_ source:NSDictionary!)` in `RCTSwift.swift`
```swift
if (_source?.uri == nil || _source?.uri == "") {
  DispatchQueue.global(qos: .default).async {
    self._player?.replaceCurrentItem(with: nil)
  }
  return;
}
```

Video.js
```diff
@@ -276,16 +276,20 @@ export default class Video extends Component {
...
-      console.warn('Trying to load empty source.');
+      console.log('Trying to load empty source.');
     }
 
-    const isNetwork = !!(uri && uri.match(/^https?:/));
-    const isAsset = !!(uri && uri.match(/^(assets-library|ph|ipod-library|file|content|ms-appx|ms-appdata):/));
+    const isNetwork = !!(uri && uri.match(/^https?:/i));
+    const isAsset = !!(uri && uri.match(/^(assets-library|ph|ipod-library|file|content|ms-appx|ms-appdata):/i));
+
+    if (uri && !isNetwork && !isAsset) {
```
evoactivity: If `uri` is `''` this will never run

change to

```typescript
if ((uri || uri == '') && !isNetwork && !isAsset) {
```

evoactivity (changes requested):

examples/basic/src/VideoPlayer.android.tsx
Gta-Cool: I'm saying that because this change is located on the example directory, which, I assume, may be used by some people as examples. For me, it should only contain practical code and if possible, try to enforce good practices. 

@evoactivity: I understand why @freeboub did it that way. In my opinion, it shouldn't be done here, it's not the responsibility of the example to test the code.

evoactivity: I would agree with that in general.

With the work @matinzd is doing on #2845 to move to [react-native-test-app](https://github.com/microsoft/react-native-test-app) we should consider separating the examples from the testing apps.

I wouldn't wait on having separate test apps to land this though.

Maybe instead of changing the sintel example we could add another src to the list and give it a more appropriate description?

```js
{
  description: 'allows non-standard casing in scheme'
  uri: 'HttPs://bitmovin-a.akamaihd.net/content/sintel/sintel.mpd',
  type: 'mpd',
},
```
Then once #2845 lands we can clean up the examples for v6.

Video.js
freeboub: This is done, thank you

freeboub:
> I have the beginning of a fix that will stop playback on iOS.
> 
> Add this before `removePlayerLayer()` in `func setSrc(_ source:NSDictionary!)` in `RCTSwift.swift`
> 
> ```swift
> if (_source?.uri == nil || _source?.uri == "") {
>   DispatchQueue.global(qos: .default).async {
>     self._player?.replaceCurrentItem(with: nil)
>   }
>   return;
> }
> ```

This is added ! thank you

examples/basic/src/VideoPlayer.android.tsx
freeboub: OK, I just remove it

Gta-Cool: I'm ok with @evoactivity's proposal to add a new entry with an appropriate description

Gta-Cool (approved):
LGTM 👍🏻

matinzd:
Looks good to me as well. I tested it last night.

matinzd (approved):

evoactivity (approved):

PR #2861, opened 2022-09-20
add subtitles style support to 5.2.x

sam1463:
Thanks for opening a PR!

#### Update the documentation

#### Update the changelog

#### Describe the changes
Taking the changes from [this merged PR](https://github.com/react-native-video/react-native-video/pull/2759) and applying them to 5.2.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,7 +855,7 @@ Platforms: Android MediaPlayer
 
 Property | Description | Platforms
 --- | --- | ---
-fontSizeTrack | Adjust the font size of the subtitles. Default: font size of the device | Android
+fontSize | Adjust the font size of the subtitles. Default: font size of the device | Android
 paddingTop | Adjust the top padding of the subtitles. Default: 0| Android
 paddingBottom | Adjust the bottom padding of the subtitles. Default: 0| Android
 paddingLeft | Adjust the left padding of the subtitles. Default: 0| Android
--- a/README.md
+++ b/README.md
@@ -325,6 +325,7 @@ var styles = StyleSheet.create({
 * [selectedVideoTrack](#selectedvideotrack)
 * [source](#source)
 * [stereoPan](#stereopan)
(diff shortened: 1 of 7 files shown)


evoactivity:
Do we want to backport features? I understand fixing 5.2.X so it runs on later versions of react-native but taking features from 6 and backporting to 5.2 seems like something that would be managed as a personal fork.

What do you  think @freeboub?

sam1463:
Certainly makes sense not to backport features to old, unsupported versions, and I'm fine with just relying on my own fork until I can update to v6, if that's the best route for the library.
However, since 5.2.1 is still the most recently published on npm, if there's no intent of feature parity with v6, it's probably best to clarify in the documents for this library that some features (such as subtitle styling support) aren't supported in the version available on npm.
It's counter-intuitive to install via yarn, follow the docs, and then have it not work because the version you got isn't actually the latest version of the library and doesn't include feature parity.

evoactivity:
That's a good point about the docs.

The readme on NPM is the 5.2.X version which makes no mention of `subtitleStyle` feature but the readme on github doesn't make it clear on where to go for 5.2.X docs so it would be easy to end up on the v6 docs.

I've added a PR that is more clear about the current state of the repo
https://github.com/react-native-video/react-native-video/pull/2875

freeboub:
I agree this is not intended to backport new featues to 5.X branch, but while there is no stable V6 release it can make sense...
@sam1463 If you don't already have your own fork I am ok to merge it and make a new release, please just add a note in changelog.

sam1463:
@evoactivity Thanks for the readme updates!

@freeboub what do you mean by the second part of your comment? This is a PR from my own fork of this library. Do you want me to add a note in the changelog on my fork/this PR?

freeboub:
@sam1463 if you have your own fork let's cancel this PR as this is already available on master.
I was requesting a note in changeLog to be able to merge this PR

sam1463:
👍 closing this then

PR #2870, opened 2022-09-25
Fix iOS RCTSwiftLog naming collision with rive-react-native

### LINKED ISSUE #2868: conflicting with Rive
as like i mentioned here in my post on rive github

these 2 libs are conflicting each other on XCode build time.

[https://github.com/rive-app/rive-react-native/issues/145](https://github.com/rive-app/rive-react-native/issues/145)

evoactivity:
This PR will fix #2868, there may be other libraries where this collision should now be avoided, eg react-native-ios-context-menu ran into a similar issue and solved it in the same way https://github.com/dominicstop/react-native-ios-context-menu/commit/088254f53737822ded82b20ad499ad9c1307b70f

I've tested on an iPhone 14 emulator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Changelog
+### Version 6.0.0-alpha4
+- Fix iOS RCTSwiftLog naming collision [#2868](https://github.com/react-native-video/react-native-video/issues/2868)
+
 ### Version 6.0.0-alpha3
-- fix ios build [#2854](https://gthub.com/react-native-video/react-native-video/pull/2854)
+- Fix ios build [#2854](https://github.com/react-native-video/react-native-video/pull/2854)
 
 ### Version 6.0.0-alpha.2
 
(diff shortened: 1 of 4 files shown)


freeboub:
@evoactivity Not tested but it looks a safe change for me. OK to merge ?

evoactivity:
I've had no problems whilst using my fork of this. The user who reported the issue was able to build after making these changes so I think it should be a safe merge.

freeboub:
> I've had no problems whilst using my fork of this. The user who reported the issue was able to build after making these changes so I think it should be a safe merge.

OK thanks! (managing changeLog is a nightmare ! )

zhekix:
what about https://github.com/react-native-video/react-native-video/blob/master/ios/Video/RCTVideo-Bridging-Header.h:2 ?

#import "RCTSwiftLog.h" -> #import "RCTVideoSwiftLog.h"

evoactivity:
I wonder why it didn't error when building. I'll submit a new PR fixing that import. Good spot @zhekix :)

evoactivity:
@zhekix this has been addressed with #2879 
Thanks for letting us know.

PR #2880, opened 2022-10-03
Feature/saaa 9116 atv audio focus

joudard-wyplay:
Thanks for opening a PR!

#### Update the documentation

#### Update the changelog

#### Describe the changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Changelog
 
+### Version 5.x 2022/10/04.
+
+- Fix Android TV AudioFocus specific case
+  Android-TV has a specific use-case not related to the “Play-Pause” button (intention) 
+  forwarded to all apps even the ones in background.
+  So if App is active we must capture back the AudioFocus to any “concurrent” sound App.
+
+  Test with music App (e.g. TuneIn) VS video player,
+  on play-pause the music App sound should not come to foreground
+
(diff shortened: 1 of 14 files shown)


evoactivity:
Hi @joudard-wyplay 👋

Thanks for submitting your first PR.

This will need some work before it can be landed including a rebase. Most work right now is happening on master for version 6, this looks like it was forked from an older copy of version 5.

**Can you please explain what this fixes or adds** and we can decide if it should be merged to v6 or v5 or both.

There seems to be unrelated work in the PR too, some dependencies and an analytics prop that would need to be removed.

joudard-wyplay:
Ho sorry, I think I just messed up the repo  / master branch
I was just making an update on the old fork we work with
only the two last commit could be of general interest

freeboub:
I think youbora support can make sens ! :) Even if It would need some rework

evoactivity:
@freeboub maybe a COOKBOOK.md could be added with a tutorial on how to integrate with an analytics platform? I feel like analytics should be left to the consumer. If we add youbora why not any of the other platforms? Will be more to maintain.

freeboub:
@evoactivity I agree about maintenance, I already integrate something like 5 differents analytics library and each one have completly different API ... so the cookbook will be a nightmare to maintain :) let's wait and see

PR #2882, opened 2022-10-04
added homepage to package.json

Adnan-Bacic:
#### Describe the changes
Added "homepage" to package.json. This will let people see/click the link when hovering over this library in their own package.json file. Then it opens the link. Right now there is no link shown, and you have to manually google the library to go to the github repo. This just makes it a little easier.


Adnan-Bacic:
not sure i added it to the correct changelog section, but i think it goes there.

freeboub:
> not sure i added it to the correct changelog section, but i think it goes there.

It looks good, thanks

Issue #2887, opened 2022-10-07
Feature: playing audio over earpiece


liviu-padurariu:
# Feature Request

playing audio over earpiece

## Why it is needed

to be able to play video and audio while keeping a quiet atmosphere.

## Possible implementation

The `Video` component will receive a new prop `audioOutput: PropTypes.oneOf(['speaker', 'earpiece'])` and based on that one can update the audio output on the native side.

### Code sample

`<Video ... audioOutput={'earpiece'} />`

freeboub:
This feature has been merged in 6.0.0-alpha.8 (with regression :/ )

opensourcekam:
What is the state of this change? I noticed the prop is here in [props.md](https://github.com/react-native-video/react-native-video/blob/master/docs/pages/component/props.md#audioOutput) but this still is not configurable in the Video component. Could you share what is happening in the regression?

PR #2888, opened 2022-10-07
Feature: playing audio over earpiece

liviu-padurariu:
Feature: playing audio over earpiece [#2887](https://github.com/react-native-video/react-native-video/issues/2887)

-  updated docs and changelog

`<Video ... audioOutput={'earpiece'} />`

This PR allow to change the audio output to the earpiece.

#### Describe the changes

Change the native audio output based on the `audioOutput` prop.


freeboub:
Thank you for this PR !
Code looks good, but I have few questions:
- The earpeace is only for JACK output ? 
- Will it work with bluetooth speakers ?
- Do you use an external package to detect if an earpiece is connected ?
I think these answers can be doc addon.

A plus, would be integration in sample, but it is not mandatory

evoactivity:
@freeboub I believe this is to play from the speaker used normally for phone calls.

liviu-padurariu:
@evoactivity  Yes, you are right.

@freeboub `earpiece` is the earpiece speaker of the phone. sorry for the confusion. I can rename it, if you have any suggestions.

freeboub:
ok, great I understood, wording is OK to me :) 
Just one missing answer: Do you use an external package to detect if an earpiece is connected ?
maybe another package include such feature ?
https://stackoverflow.com/questions/52127683/check-if-headset-is-connected-on-different-api-levels

liviu-padurariu:
no, I didn't use any external packages.

freeboub (approved):

freeboub:
I approuve the PR, but I cannot test as the USB connector of my phone is broken ...
I would like a confirmation, you don't reapply to audio output at playback start, is it working after playback change ?
sample use case:
- start playback of content 1
- change audio output to earpiece
- stop playback but putting an empty url
- start another playback
- > here is the audio output correctly reapplied ?
Thank you

freeboub:
Has been merged in another PR

PR #2897, opened 2022-10-22
Fix predictive back gesture handler on android video player

jasminmif:
This adds support for android devices which do no have "physical back button" so the only way to go back is using the new Android predictive back gesture. 
Before when we would use the back gesture we would end up in an infinite loop triggering & hiding the media controls and not being able to exit out of the video because the gesture is handled as a `onTouch` event.


freeboub:
Closing this old, It is not targeting to master branch.
BTW, I hope last V6 version should solve the issue you are dealing here.

PR #2901, opened 2022-10-28
Pause Native Video on React Video Component unmount

DmitriyGalyanov:
[Looky-1313](https://oneclicklife.youtrack.cloud/issue/Looky-1313/120-299390-iOS-Zvuk-ot-video-posta-ne-ostanavlivaetsya-posle-zakrytiya-posta)
Возможно, [Looky-1314](https://oneclicklife.youtrack.cloud/issue/Looky-1314/120-299390-V-lente-dubliruetsya-zvuk-ot-video-postov-na-novom-profile)

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -267,6 +267,17 @@ export default class Video extends Component {
     return UIManager.getViewManagerConfig(viewManagerName);
   };
 
+  setNativePause(paused = this.props.paused) {
+    this.setNativeProps({ paused })
+  }
+  componentDidMount() {
+    this.setNativePause()
+  }
+  componentWillUnmount() {
+    // ensures video will stop playing on React-Component unmount
+    this.setNativePause(true)
+  }
+
   render() {
     const resizeMode = this.props.resizeMode;
     const source = resolveAssetSource(this.props.source) || {};


freeboub:
Hi, I disagree with this, unmouting the <video should fully stop the player and release all resources! If it is not the case that is a big issue !

What about if you umount 10 different video ?

DmitriyGalyanov:
> Hi, I disagree with this, unmouting the <video should fully stop the player and release all resources! If it is not the case that is a big issue !
> 
> 
> 
> What about if you umount 10 different video ?
Hi!

I'm sorry, didn't mean to open a PR in main repo, meant to open in a colleague's fork

I'm kinda worried about resource release, too.
But I don't have enough time to dig native :(

In our case video doesn't stop audio after React Component unmount (eg: when leaving screen with a playing video)

freeboub:
I don't reproduce this issue on master, this feature is testable on basic sample (look for noView field in the sample).
Please open a ticket if you find reproduction step

alexfoxy:
I have fixed with this component:
```
import React from 'react'
import RNVideo from 'react-native-video'

let ref: RNVideo | null

const Video = (props: RNVideo['props']) => {
  return <RNVideo
    ref={r => {
      if (!r) {
        // @ts-ignore setNativeProps does not exist on 'Video' .. but it does!
        if (ref?.setNativeProps) ref.setNativeProps({ paused: true })
      } else {
        ref = r
      }
    }}
    {...props}
  />
}

export default Video
```

PR #2919, opened 2022-11-10
fix: delete unwrap

StIch0:
- добавлена проверка на наличие reactViewController, что бы избежать ошибки при которой приложение крашится


freeboub:
This change looks good for me, but can you please add a line in changeLog describing why you need this fix?
To fix a crash I guess :)

freeboub:
Similar fix has already be merge, then let's close this PR.
Let me know if it needs to be reopened

Issue #488, opened 2017-02-10
Integrate VAST Ads for Pre-Roll Mid-Roll Post-Roll


FullstackJack:
Both Android and iOS possess the capability to integrate with third-party Video Ad Serving Tempaltes (VAST) services to pre-roll, mid-roll, and post-roll ads interspersed with video playback. If  React Native is to be considered a serious replacement for native app development, we need a mechanism to generate ad revenue from video views.

As mentioned, there are pre-existing native solutions that exist for both iOS and Android platforms. Interactive Media Ads (IMA) SDK is one such solution which integrates with both. The IMA SDK is maintained by Google. 

IMA SDK integrates with DoubleClick for Publishers (DFP), Google AdSense network, or any VAST-compliant ad server. [Google Developers](https://developers.google.com/interactive-media-ads/) has documentation for integration on each platform.

[IMA Platform Support Matrix](https://developers.google.com/interactive-media-ads/docs/sdks/ios/compatibility)
[Android Get Started](https://developers.google.com/interactive-media-ads/docs/sdks/android/quickstart)
[iOS Get Started](https://developers.google.com/interactive-media-ads/docs/sdks/ios/quickstart)

OrenMe:
Hi @FullstackJack, do you have an example of IMA SDK for Android or iOS that actually works?
From a couple of experiments with Android it seems that if you wrap IMA SDK in a RN component then there are some weird quirks with the clocks and the outcome is that although ad request does load and also the ad creative (you can see it in a network sniffer) the IMA SDK always emits an error and ad playback fails.

FullstackJack:
I wish I did, but I hadn't made it to that point before switching to a new project. I was working on a prototype and the company Ibwas building it for decided to spend a huge amount of money on full native versions instead. Maybe I'll come back to it someday.

OrenMe:
Thanks @FullstackJack.
I asked around with the IMA SDK team but unfourtounatly they are not supporting react native officially yet.

brinkkemper:
I've spend the past couple of days trying to get this to work. If it works I could contribute to this project. However, I'm horribly stuck on this strange issue: 
[Google IMA webview is not showing when used inside react native on android
](https://stackoverflow.com/questions/48567794/google-ima-webview-is-not-showing-when-used-inside-react-native-on-android)
Perhaps someone here has the solution?

benoit-ponsero:
Hi, 
any news on this feature ? it could be interesting to have

[...]

PR #2923, opened 2022-11-16
Feature: Add VAST support for AVOD

avencat:
Hi and thank you very much for this library!
I need to add AVOD ads to my app and following [this issue (#488)](https://github.com/react-native-video/react-native-video/issues/488) I managed to add Google's IMA sdk to this library.

I would like to thank @RobbyWH, @olgunkaya, @sdkcy and @ahmetberkinc for their contributions that made this PR possible. Also thank you @FullstackJack for opening this PR in the first place and all contributors to the issue.

@hueniverse asked for a PR so here it is! 🙂 (Code has been re-written in swift and the java part has been updated to the latest exoplayer version available on this repository)

This is how to test the change:
```typescript
import Video from 'react-native-video'

const VideoComponent = () => {
  const onReceiveAdEvent = (event) => {
    console.log(event)
  }

  return (
    <Video
      onReceiveAdEvent={onReceiveAdEvent}
      source={{ uri: 'https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_fmp4/master.m3u8' }}
      style={{ flex: 1 }}
    />
  )
}
```


android/build.gradle
```diff
@@ -45,5 +45,7 @@ dependencies {
     implementation('com.google.android.exoplayer:extension-okhttp:2.18.1') {
         exclude group: 'com.squareup.okhttp3', module: 'okhttp'
     }
+    implementation 'com.google.android.exoplayer:extension-ima:2.11.4'
```
freeboub: Should be 2.18.1. can there be some side effects to add this dependency?

API.md
```diff
@@ -1212,6 +1213,41 @@ Example:
...
+Payload:
+
+| Property | Type   | Description           |
+|----------|--------|-----------------------|
+|  event   | string | The ad event received |
+
+Example:
+```
+{
+  "event": "LOADED"
```
freeboub, line 1292: can you retail all types of ads events available please ?

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -83,7 +86,10 @@ public ExoPlayerView(Context context, AttributeSet attrs, int defStyleAttr) {
         layout.addView(shutterView, 1, layoutParams);
         layout.addView(subtitleLayout, 2, layoutParams);
 
+        adOverlayFrameLayout = new FrameLayout(context);
+
         addViewInLayout(layout, 0, aspectRatioParams);
+        addViewInLayout(adOverlayFrameLayout, 1, layoutParams);
```
freeboub: Is there a reason why it is not added in layout ?
`        layout.addView(adOverlayFrameLayout, 3, layoutParams);
`
With this you should be able to benefit of aspect ratio management ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -221,6 +231,9 @@ public void handleMessage(Message msg) {
             switch (msg.what) {
                 case SHOW_PROGRESS:
                     if (player != null) {
+                        if (isPlayingAd()) {
```
freeboub: I think you should check that controls are enable or at least playerControlView is not null

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -261,8 +274,13 @@ public ReactExoplayerView(ThemedReactContext context, ReactExoplayerConfig confi
         audioManager = (AudioManager) context.getSystemService(Context.AUDIO_SERVICE);
         themedReactContext.addLifecycleEventListener(this);
         audioBecomingNoisyReceiver = new AudioBecomingNoisyReceiver(themedReactContext);
+        // Create an AdsLoader.
+        adsLoader = new ImaAdsLoader.Builder(themedReactContext).setAdEventListener(this).build();
     }
 
+    private boolean isPlayingAd() {
+        return player != null && player.isPlayingAd() && player.getPlayWhenReady();
```
freeboub: why  `&& player.getPlayWhenReady()` ? we cannot pause during ads ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -277,6 +295,8 @@ private void createViews() {
             CookieHandler.setDefault(DEFAULT_COOKIE_MANAGER);
         }
 
+        MultiDex.install(getContext());
```
freeboub: It looks strange to me.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -726,7 +769,11 @@ private MediaSource buildMediaSource(Uri uri, String overrideExtension, DrmSessi
         int type = Util.inferContentType(!TextUtils.isEmpty(overrideExtension) ? "." + overrideExtension
                 : uri.getLastPathSegment());
         config.setDisableDisconnectError(this.disableDisconnectError);
-        MediaItem mediaItem = new MediaItem.Builder().setUri(uri).build();
+
+        MediaItem mediaItem = new MediaItem.Builder()
+            .setUri(uri)
+            .setAdsConfiguration(new MediaItem.AdsConfiguration.Builder(adTagUrl).build())
```
freeboub: should you test adTagUrl value before calling setAdsConfiguration ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -810,12 +857,14 @@ private MediaSource buildTextSource(String title, Uri uri, String mimeType, Stri
...
     private void releasePlayer() {
         if (player != null) {
+            adsLoader.setPlayer(null);
             updateResumePosition();
             player.release();
             player.removeListener(this);
             trackSelector = null;
             player = null;
         }
+        adsLoader.release();
```
freeboub, line 874: would be better to adsLoader to null to allow garbage collector to clear memory

freeboub (changes requested):

android/build.gradle
avencat: Yes you're absolutely right, I forgot to push this file, sorry, I'll push it now

API.md
avencat: Will do!

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
avencat: I'll check with your suggestion! 🙂

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: I think we should be able to do this, I'll remove this condition!

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: Yes I should and will!

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: Hum… The fact that we need MultiDex?

avencat: Maybe we don't need it, I just followed Google's documentation, I'll make all your suggestions then remove it to see if it still works

freeboub: Yes I understand but it should be called once in an application. Here you will call once per Video view (maybe multiple times on remount). I think this is more a responsibility of application.
Additionally you should include multidex dependency in gradle file if it is really mandatory.
But i think a note in documentation should be enough!

API.md
avencat: Done in the last commit

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: Ok nice, I'll add this to the documentation and remove multidex from grade file and from here! 😉

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
avencat: It works like a charm, thanks! 😄

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: Apparently it works just fine without MultiDex so we can remove it purely and simply. 🙂

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: I'm not sure about this one, what do you want me to do?
```suggestion
        adsLoader = null;
```

or

```suggestion
        adsLoader.setPlayer(null);
```
like in the condition above
or

```suggestion
        adsLoader.release();
        adsLoader = null;
```
?

freeboub: I think the 3rd one would be better !

avencat: Alright, I made this change! 😉

avencat: As this is my first PR I'm not sure how to deal with the conversations, should I resolve them myself or you will once you think the corrections are ok?

freeboub: You put them as closed, I will do another review round & some tests in sample app before my final approval :)

ios/Video/RCTVideo.swift
```diff
@@ -279,19 +295,24 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
                 self._player?.actionAtItemEnd = .none
-                
+
                 if #available(iOS 10.0, *) {
                     self.setAutomaticallyWaitsToMinimizeStalling(self._automaticallyWaitsToMinimizeStalling)
                 }
-                
+
+                // Set up your content playhead and contentComplete callback.
+                self.contentPlayhead = IMAAVPlayerContentPlayhead(avPlayer: self._player!)
```
freeboub: you should do it only if ads url is available ?

ios/Video/RCTVideo.swift
```diff
@@ -1080,4 +1118,159 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     //         }
     //         */
     //    }
+
```
freeboub: Do you think this changes can be moved to a new file ?
Just try to keep code cleaner
Thanks

avencat: Yes I'll try to move it to a new file! 😉

avencat: I created a new file RCTIMAAdsManager under Features folder and put everything I could in it! 😉

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -189,6 +190,23 @@ public void setSrc(final ReactExoplayerView videoView, @Nullable ReadableMap src
...
+    @ReactProp(name = PROP_AD_TAG_URL)
+    public void setAdTagUrl(final ReactExoplayerView videoView, final String uriString) {
+        if (TextUtils.isEmpty(uriString)) {
+            return;
+        }
+
+        if (startsWithValidScheme(uriString)) {
+            Uri adTagUrl = Uri.parse(uriString);
+
+            if (adTagUrl != null) {
```
freeboub: Here you should set the ad tag uri even if it is null !
With this code you will always keep previous one.
I also to think you should do it even if you have an invalid uriSheme. In case of local video, uriSheme will be invalid

API.md
```diff
@@ -1212,6 +1213,86 @@ Example:
...
+| `AD_PERIOD_STARTED`        | iOS           | Fired every time the stream switches from content to advertising or slate. This will be fired even when an ad is played a second time or when seeking into an ad (only used for dynamic ad insertion).      |
+| `AD_PROGRESS`              | Android       | Fires when the ad's current time value changes. Calling getAdData() on this event will return an AdProgressData object.                                                                                     |
+| `ALL_ADS_COMPLETED`        | Android, iOS  | Fires when the ads manager is done playing all the valid ads in the ads response, or when the response doesn't return any valid ads.                                                                        |
+| `CLICKED`                  | iOS           | Fires when the ad is clicked.                                                                                                                                                                               |
```
freeboub: Is it possible to merge both events 😒 I understand it come from IMA, but it would be easier for app implementation!

API.md
```diff
@@ -1212,6 +1213,86 @@ Example:
...
+| `LOG`                      | Android, iOS  | Fires when a non-fatal error is encountered. The user need not take any action since the SDK will continue with the same or next ad playback depending on the error situation.                              |
+| `MIDPOINT`                 | Android, iOS  | Fires when the ad playhead crosses midpoint.                                                                                                                                                                |
+| `PAUSED`                   | Android       | Fires when the ad is paused.                                                                                                                                                                                |
+| `PAUSE`                    | iOS           | Fires when the ad is paused.                                                                                                                                                                                |
```
freeboub: Same here 😣

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
avencat: Ok I'm updating this now! 🙂

API.md
avencat: Yes I'll merge them! 🙂

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
avencat: Done! 😉

freeboub (approved):

freeboub:
I merged this PR as it is a very big change ! code looks fine, let see on next release !
Thank you

avencat:
Nice, thank you very much!

reinismu:
Can we make this optional? I bet 99% of these libs users don't want to build google ads in their application. Don't want it to bulk up my application with it

hassanrazaby433:
@reinismu @freeboub  I cannot find the alpha.5 version mentioned in the change log which supports adTagUrl prop.
Can you please help?

sametcl2:
you can use master branch

mitchdowney:
To @reinismu's point, our app's build uploads are getting rejected from Google Play after updating to the latest react-native-video nightly. Can this be optional? Or is there a workaround so we don't have to register our app as having ads with Google?

There is a fix to the onProgress interval we'd like to use in alpha.5. If there are not better options, I can just create a patch of alpha.4 with the onProgress interval fix for now.

freeboub:
@mitchdowney can you please detail rejection feedback ? If it causes mandatory ads registration steps, I agree we need to find a solution to disable it ! 
Regarding to the hot fix you need I can only advise you to create your own fork if needed...

mitchdowney:
@freeboub I was able to patch in the onProgress interval fix to alpha.4 and it seems to work fine for our purposes.

Here is the error that our App Center distribution pipeline received after updating to alpha.5:

> Edit could not be committed due to an error. This release includes the com.google.android.gms.permission.AD_ID permission but your declaration on Play Console says your app doesn't use advertising ID. You must update your advertising ID declaration.

We specified just a few weeks ago that Podverse doesn't use an advertising ID, but it appears that we would need to change that designation to use alpha.5.

After I reverted to alpha.4, the error went away and App Center was able to send the build to Google Play successfully.

reinismu:
@mitchdowney I just reverted google AD PR https://github.com/reinismu/react-native-video/pull/1/files

freeboub:
@mitchdowney Here is a pull request where you can enable or disable ads insertion on demand on android: https://github.com/react-native-video/react-native-video/pull/3027

> We specified just a few weeks ago that Podverse doesn't use an advertising ID, but it appears that we would need to change that designation to use alpha.5.
You mean that it doesn't impact ios certification ?


@reinismu I don't think reverting this feature is a good thing. CSAI is a good feature !

mitchdowney:
@freeboub fantastic! thank you!

hassanrazaby433:
I found out some issues with react-native-video with an adTagUrl on IOS.
1. In PRE-ROLL Ads, the video starts with an AD on android, while on IOS the video is played for 2 seconds and then the AD is streamed. Can we add a prefetching of ads to resolve that?
2. On IOS when a PRE ROLL AD is loaded, it pauses the player and we have to manually play it again in order continue.
3. On IOS , When we go to the player and the Ad request is sent by the player to fetch the ad, in this process if we go back (unmount the player)  , the Ad audio keeps on playing in the background.

sametcl2:
Hi, @avencat I have separate VAST ids for preroll and midroll. How can I use multiple VAST ids?

avencat:
Hi @sametcl2 I really don't know, but can't you merge all VAST ids in one? As one VAST could handle both…

sametcl2:
> Hi @sametcl2 I really don't know, but can't you merge all VAST ids in one? As one VAST could handle both…

I solved this issue with using VMAP

PR #2926, opened 2022-11-22
fix: repeat video in iOS with AVPlayerLooper in 5.2.x

swemail:
Hello and thank you for a great library! This is a shot at the fix proposed in #2817 

To test this is just to run the basic example in this repo. Maybe remove the alert at the end to get the sense of the smoother repeat 😄 

The solution proposed is to create a `AVQueuePlayer` with a `AVPlayerLooper` if supported (iOS >= 10) and fall back to the previous used `AVPlayer` if it's not. The `AVPlayerLooper` need to be references to not be disposed. The variable holding that reference is typed to `NSObject` to make it compatible with previous iOS versions not supporting `AVPlayerLooper`. The order of things matter and I had to move passing the `playerItem` to the `playerObserver` until after initiation of the player


MitoMonkey:
Tested it and using it now in our app. Works like a charm. Thanks a lot @swemail

MitoMonkey:
A bit more testing revealed a problem with this PR: If multiple (audio) files are played at the same time, I get a weird behaviour. Sometimes only one of them is audible at all, sometimes both. In one of them I am using the onLoad and onProgress props to display a progress bar. But I receive currentTime values from onProgress before onLoad is executed and a lot of currentTime = 0 logs before the value starts to change. Sometimes onLoad is not executed at all, even though the audio is audible... Can't make much sense of that behaviour, but it is not like that without the changes in this PR.

swemail:
> A bit more testing revealed a problem with this PR: If multiple (audio) files are played at the same time, I get a weird behaviour. Sometimes only one of them is audible at all, sometimes both. In one of them I am using the onLoad and onProgress props to display a progress bar. But I receive currentTime values from onProgress before onLoad is executed and a lot of currentTime = 0 logs before the value starts to change. Sometimes onLoad is not executed at all, even though the audio is audible... Can't make much sense of that behaviour, but it is not like that without the changes in this PR.

Oh, thank you for finding this and explaining what happens. Will look at it.

swemail:
@MitoMonkey now I've tried to address your issues. I could not reproduce several files not playing but I could see that the observers was not set up correctly so at least `onVideoLoad` was borked. My hope is that some issues are resolved by that. Could you please try again? If it still fails I could use a minmal test app that reveals the issue if that is possible.

MitoMonkey:
I got an error that there is a } too much in the code (see comment in the files section). After deleting that the build is successful, but our app completely freezes (and then crashes) as soon as any component using the RNvideo library is called. The only hints I get from Xcode are `[plugin] AddInstanceForFactory: No factory registered for id <CFUUID...` and `Message from debugger: Terminated due to signal 9`. Any ideas what is happening?

Does it work for you as it is now?

swemail:
Sorry about the typo,  my environment for this is sub par 🙈 . Removed the typo but I have no crash. I would need your react versions and look at your Podfile to get the the example running with an older version

swemail:
Installed  `examples/basic` after som Podfile adjustments with react-native 0.64.4 and it worked without crash.

MitoMonkey:
Oh my... these version differences and dependencies are really such a clutter. We are still using react-native 0.64.1 (for exactly that version compatibility reason). 

But the good news is, I got it to work with your version now. Interruptions between loop repeats are gone and playing multiple audio files simultaneously works as well. Thanks a lot again for all your effort @swemail !!!

swemail:
Great to hear @MitoMonkey and thank you for sticking to it and helping out!

BarkhaArora:
Hi! i have an audios of length more than 1 min 25 seconds. It does not repeat even after setting repeat true on ios. Can you help? Audio with smaller length are repeating.

BarkhaArora:
> Oh my... these version differences and dependencies are really such a clutter. We are still using react-native 0.64.1 (for exactly that version compatibility reason).
> 
> But the good news is, I got it to work with your version now. Interruptions between loop repeats are gone and playing multiple audio files simultaneously works as well. Thanks a lot again for all your effort @swemail !!!

Hi! i have similar situation i need to run multiple audios simultaneously on repeat until repeat is set false. can you please help or share your code? Also my audios are not repeating if there length is greater that 1 min

Issue #2723, opened 2022-06-15
[Project. Status] Version 6.0.0-alpha


hueniverse:
After a long period of inactivity, developments have restarted in this repository. V6.x branch already contains number of alpha releases.

These releases includes many bug fixes and improvements from many contributors. I want to express my deep gratitude for all the help in getting this important milestone. Many contributors had to rebase their contributions multiple times as some where a few years old.

Please test the new release with your app and report back any issues. Please clearly label your issues as v6 so that we can prioritize them. 

Main changes in V6 branch:
- Remove support for DOM
- Remove support for Android media player (only support Exo player)
- Rewrite AVPlayer (ios) binding in swift
- Add client side ads insertion on ios and android (google ima)
- Update exoplayer to media3 
 
This branch should now be be stable, we reach the beta state.
Please report any regressions comparing to 5.2.1, they will be handled as a priorityand  tagged  https://github.com/react-native-video/react-native-video/labels/critical%20for%20release
If you want to contribute, following ticket are a good first approach: https://github.com/react-native-video/react-native-video/labels/good%20first%20issue

Roadmap to be discussed here: https://github.com/react-native-video/react-native-video/discussions/3351
Project status discussions here: https://github.com/react-native-video/react-native-video/discussions/3350

(Edited by @freeboub)

JstnEdr:
Looks like fullscreen video on Android is still not supported. Are there any plans to address this in v6?

hueniverse:
@JstnEdr if there is a PR ready or someone willing to take the lead, sure. There are no free resources around atm...

baveku:
I hope we can custom control UI, on iOS it's beautiful but on Android, we don't have a fullscreen button.
With 0.69.0 we need to integrate using New Architecture, I think we need to migrate this library.

JongGyuChoi:
A build error occurred in RCTVideoManager
<img width="689" alt="Screen Shot 2022-06-22 at 10 57 01 AM" src="https://user-images.githubusercontent.com/33406511/174926942-0c8d98fd-3e37-4713-89e8-4c58781fdb2d.png">
<img width="560" alt="Screen Shot 2022-06-22 at 10 57 07 AM" src="https://user-images.githubusercontent.com/33406511/174926956-2d45ff81-63e6-48d2-be48-f1617e904ff5.png">

.

baveku:
[...]

PR #2943, opened 2022-12-07
Update RCTVideo.swift

marcel-happyfloat:
Prevents a crash in my case: https://github.com/react-native-video/react-native-video/issues/2723#issuecomment-1336130120


freeboub (approved):

marcel-happyfloat:
@freeboub Thanks for maintaining this library. Unfortunately I don't know much about iOS development, so I'm not really sure why it can be undefined and what consequences it could have. It just solved a problem in my case and I didn't see any other problems with it. Thanks

PR #2955, opened 2022-12-14
fix(android): check null activity

wjaykim:
If activity is null at initialization, the app crashes. This change checks for null activity and throw react native error, which prevents app crash.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -572,6 +572,12 @@ class ReactExoplayerView extends FrameLayout implements
                                     eventEmitter.error("Failed to initialize DRM Session Manager Framework!", new Exception("DRM Session Manager Framework failure!"), "3003");
                                     return;
                                 }
+                                    
+                                if (activity == null) {
+                                    Log.e("ExoPlayer Exception", "Failed to initialize Player!");
+                                    eventEmitter.error("Failed to initialize Player!", new Exception("Current Activity is null!"), "1001");
+                                    return;
+                                }
 
                                 // Initialize handler to run on the main thread
                                 activity.runOnUiThread(new Runnable() {


alekhinen:
hey @wjaykim what stack trace were you getting for the native crash? the app i'm working on is seeing quite a bit of null pointer exceptions on android (we're running `6.0.0-alpha.2`). this is the trace we're getting from the crashes:
```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.Activity.runOnUiThread(java.lang.Runnable)' on a null object reference
  at com.brentvatne.exoplayer.ReactExoplayerView$6$1.run (ReactExoplayerView.java:548)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:929)
```

freeboub:
It seems to be the same issue regarding to the callstack. I merged the PR as it not a risky change, but I don't think it really solve the issue. 

I had similar issue on my working branch.
The use case is the following: 
- boot the app when screen is off
- app starts and try to play the video
--> crash due to null Activity

It seems to be an issue in react native but i never find time to reproduce it in a unit sample.

Is this use case similar to yours?

wjaykim:
@alekhinen @freeboub I couldn't reproduce crash in my own device, but crashlytics logged some user's crash. The error trace was same with @alekhinen posted.

alekhinen:
@freeboub I haven't reproduced the crash on my end but I'll follow those steps and see if that reliably causes an app crash. Your steps seem to line up with what Google Play is reporting for my app. About 98% of all instances of this crash occur when the app is backgrounded. 

Update: My team's been able to somewhat reliably reproduce the crash. For some context, our app has a video that plays on the first screen. Opening up the app from cold start and immediately backgrounding the app causes a crash.

PR #2917, opened 2022-11-08
fix: fix the kvo compliance crash when rapidly switching source

wood1986:
fix #2908

PR #2961, opened 2022-12-17
fix: fix wrong merge conflict due to the kvo compliance crash when rapidly switching source

wood1986:
This is the right one.


wood1986:
@freeboub as you complete #2917 while I am resolving the conflict. It missed some code. I added it back.

freeboub:
@wood1986 sorry, I was too fast ! Thank you for the update!

PR #2969, opened 2022-12-21
fix: fix ios compile issue

wood1986:
fix RCTVideo.swift:300:26 Value of optional type 'AVPlayer?' must be unwrapped to refer to member 'replaceCurrentItem' of wrapped base type 'AVPlayer'


wood1986:
@freeboub. I resolved the conflict incorrectly. This will fix the compilation issue on iOS. Sorry for the inconvenience caused

freeboub:
@wood1986 only people how don't do anything, don't do mistakes ;) thank you for your reactivity !

freeboub:
And I prefer a fix than a revert!

PR #2985, opened 2023-01-04
[Android] Set controls and landscape when going fullscreen

reinismu:
This allows fullscreen to work even if you don't have native controls. (Who even thought to merge it in without this working?)

While at it, it puts the screen in the landscape for the best viewing experience and puts it back as it was when exiting. 

Maybe there were reasons why it was done as it was before, so before cleaning up I want to hear your opinion on this.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -9,6 +9,7 @@ import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.ActivityManager;
 import android.content.Context;
+import android.content.pm.ActivityInfo;
 import android.media.AudioManager;
 import android.net.Uri;
 import android.os.Handler;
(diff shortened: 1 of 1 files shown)


freeboub:
Hello, and thank you for submitting a PR!
But it cannot be merged as is :) 
I wonder why you don't handle this feature in your application? Once you move to Fullscreen, show controls and force orientation (need another rn package)
I mean, these behavior are part of your business logic.
1/ you cannot force displaying controls once Fullscreen, if other users implement there own controls, they will have double controls displayed... (I even personally think controls should be out of this package, but too much people use it to drop it)
2/ orientation: i don't think it is a good idea to control orientation in this package as we can have other applications which controls orientation with another package (to use there own business logic), then we conflict between both orientation control expections.

I hope it answers to you questions!

Anyway, you can keep using your own fork if you think it is simplier to implement like that !

freeboub:
OK, I aml just reviewing ios implementation.
In fact your PR shall implement new properties like on ios, see following props in API.md

| [fullscreen](#fullscreen)                                                           | iOS                       |
| [fullscreenAutorotate](#fullscreenautorotate)                                       | iOS                       |
| [fullscreenOrientation](#fullscreenorientation)                                     | iOS                       |

(and controls shall still be handled by your application I think)

Issue #3009, opened 2023-01-23
DRM NOT WORKING - _resouceLoaderDelegate destroyed accidentally


PierluigiCod:
### Current behavior
When i try to play a DRM content, RCTVideo._resouceLoaderDelegate is accidentally destroyed when func RCTVideo.removePlayerLayer() is called.

### Reproduction steps
Instantiate Video player with drm object.

### Expected behavior
RCTVideo._resouceLoaderDelegate is destroyed only in deinit.

### Platform
* iOS

freeboub:
will be included in 6.0.0-alpha.5

PR #3010, opened 2023-01-23
fix issue #3009

PierluigiCod:
Destroy RCTVideo._resouceLoaderDelegate only when an instance of AVPlayerItem or an instance of AVPlayer is destroyed.


freeboub:
Tested sucessfully on my side !

PR #3015, opened 2023-01-28
merge ios and android samples

freeboub:
update sample to ensure it works on ios


freeboub:
Mix between 2 changes. This PR will be canceled and replaced by other one

KrzysztofMoch (approved):
LGTM 👍 but you can wrap app with safe area view

PR #3017, opened 2023-01-28
fix(ios): ensure controls are not displayed when disabled by user

freeboub:
Fix an issue found when testing with basic sample.
if you disable controls, set fullscreen={true} and fullscreen={false}
when leaving fullscreen controls are displayed even if controls are false


vargajacint:
LGTM!

PR #3028, opened 2023-02-08
Ad Keeps On playing in background Even when the player is unmounted FIX

hassanrazaby433:
Added   _imaAdsManager.getAdsManager()?.destroy() in removeFromSuperview method to clear Ad state when the player is unmounted. 
Prevents Ad audio to play in background even when the player is unmounted.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -883,7 +883,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         _player?.pause()
         _player = nil
         _playerObserver.clearPlayer()
-
+        _imaAdsManager.getAdsManager()?.destroy()
         self.removePlayerLayer()
 
         if let _playerViewController = _playerViewController {
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -883,6 +883,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         _player?.pause()
         _player = nil
         _playerObserver.clearPlayer()
+        _imaAdsManager.getAdsManager()?.pause()
         _imaAdsManager.getAdsManager()?.destroy()
         _imaAdsManager.getAdsLoader()?.contentComplete()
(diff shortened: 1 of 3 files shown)


ios/Video/RCTVideo.swift
```diff
@@ -883,7 +883,10 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         _player?.pause()
         _player = nil
         _playerObserver.clearPlayer()
-
+        _imaAdsManager.getAdsManager()?.destroy()
```
freeboub, line 887: Are you really sure this is working ?
As _imaAdsManager is instanciated in init function, It will never be reallocated

won't   calling: _imaAdsManager.getAdsManager()?.pause() be better in that case ?

hassanrazaby433: No this is not working, i thought it might. I am not really an IOS developer, just tried it. 
I will try _imaAdsManager.getAdsManager()?.pause() and test and see the results. This issue is causing a delay in build and we really need this feature ASAP.

freeboub: @hassanrazaby433 any news ?

kchuah:
I have the same issue with version `6.0.0-beta.5` and was able to resolve it thanks to the pointers above by adding this code instead. I wasn't able to get it working if having ad manager destroy within the `deinit` function.

Disclaimer - I am not an iOS dev 😄 

```
    override func removeFromSuperview() {
        _player?.pause()
        _player = nil
        _resouceLoaderDelegate = nil
        _playerObserver.clearPlayer()
         // this line onwards
        _imaAdsManager.getAdsManager()?.pause()
        _imaAdsManager.getAdsManager()?.destroy()
```

[react-native-video+6.0.0-beta.5.patch](https://github.com/react-native-video/react-native-video/files/14661408/react-native-video%2B6.0.0-beta.5.patch)

EDIT: This line had to be added as well: 

```
        _imaAdsManager.getAdsManager()?.pause()
        _imaAdsManager.getAdsManager()?.destroy()
        _imaAdsManager = nil // new line
```

KrzysztofMoch:
I believe that #3722 should fix this - if not please ping me

PR #3030, opened 2023-02-09
Feature: Video range support

RadinGospodinov:
# Playing only a portion of the video (start & end time)
Provide an optional `startTime` and/or `endTime` for the video. Value is in milliseconds. Useful when you want to play only a portion of a large video.
Example
```
source={{ startTime: 36012, endTime: 48500 }}
```

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -342,6 +342,8 @@ export default class Video extends Component {
         mainVer: source.mainVer || 0,
         patchVer: source.patchVer || 0,
         requestHeaders: source.headers ? this.stringsOnlyObject(source.headers) : {},
+        startTime: source.startTime,
+        endTime: source.endTime
       },
       onVideoLoadStart: this._onLoadStart,
       onVideoPlaybackStateChanged: this._onPlaybackStateChanged,
--- a/Video.js
+++ b/Video.js
@@ -342,7 +342,7 @@ export default class Video extends Component {
         mainVer: source.mainVer || 0,
         patchVer: source.patchVer || 0,
         requestHeaders: source.headers ? this.stringsOnlyObject(source.headers) : {},
-        startTime: source.startTime,
+        startTime: source.startTime || 0,
         endTime: source.endTime
       },
       onVideoLoadStart: this._onLoadStart,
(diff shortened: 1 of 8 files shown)


Video.js
```diff
@@ -342,6 +342,8 @@ export default class Video extends Component {
         mainVer: source.mainVer || 0,
         patchVer: source.patchVer || 0,
         requestHeaders: source.headers ? this.stringsOnlyObject(source.headers) : {},
+        startTime: source.startTime,
```
freeboub: maybe add || 0, then you ensure implementations get a correct value ?

freeboub:
It is a very good PR, it you good point that you put new fields in src prop !
Can you please update API.md to add the new API ?
And also add a line in changeLog.
I will merge it rapidly ! 
Thank you

yavor87:
Hey @freeboub , 
Sorry, we issued this PR a bit sooner than we intended by mistake. I have updated the api & changelog.

Video.js
freeboub: (Sorry, this is a long time comment I forgot to submit ...)

yavor87: Thank you @freeboub .

tslater:
@RadinGospodinov Does this normalize the other times of the video so that time 36012 on the video file is now time 0 (from your example? or is time 36012 still 36012?

Example

`source={{ startTime: 36012, endTime: 48500 }}`

RadinGospodinov:
Correct, it creates ClippingMediaSource and 36012 will become 0. The duration of the clip will be 48500 - 36012, so if you need to seek will need to pass a in the range between 0 and 48500 - 36012

tslater:
@RadinGospodinov  Thanks for the quick reply. I noticed that normally the `seek()` function takes a parameter in seconds, but the `startTime` and `endTime` are in milliseconds. 

Does using `startTime` and `endTime` in the source also change the seek time to seconds? (Seems like it might be the case from your example here?) 

If that's the case, would it also affect the `onProgress` event values as well?

RadinGospodinov:
There are no changes to seek and in progress, so they should continue working the same way as without start and end time.

Issue #2744, opened 2022-06-25
UI frame drop while loading (initial buffer) the video


matinzd:
# Bug
The main UI thread is being blocked by loading the video in the initial buffer, causing the UI to drop to 5-20 fps.

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info

React native info output:

```bash
System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 315.81 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
    Watchman: 2022.05.16.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /Users/matin/.rvm/gems/ruby-3.0.2/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 28, 29, 30, 31
      Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 31.0.0, 31.0.0
      System Images: android-27 | Google APIs Intel x86 Atom, android-29 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom_64, android-30 | Google APIs Intel x86 Atom, android-30 | Google Play ARM 64 v8a, android-30 | Google Play Intel x86 Atom, android-30 | Google Play Intel x86 Atom_64, android-31 | Google APIs Intel x86 Atom_64, android-S | Google Play Intel x86 Atom_64
      Android NDK: 23.0.7344513-beta4
  IDEs:
    Android Studio: 4.2 AI-202.7660.26.42.7351085
    Xcode: 13.3/13E113 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.10 - /usr/local/opt/openjdk@11/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.2 => 0.68.2 
    react-native-macos: Not Found
  npmGlobalPackages:
[...]

PR #3042, opened 2023-02-20
Fix UI lag on initial buffer for iOS

kyle-ssg:
This fixes a reoccurring issue from https://github.com/react-native-video/react-native-video/issues/2744


kyle-ssg:
Reading the history, I guess I'll close this since it can result in an edge case of fast changing src.

I would however say that reverting the original fix seems crazy since it results in an unacceptably slow UX. I'll keep the fix in my fork.

PR #3045, opened 2023-02-23
Fix PiP opening by default in 5.2.X on iOS 14+

c0rnonthec0bb:
Fixes https://github.com/react-native-video/react-native-video/issues/2177

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -883,7 +883,9 @@ static int const RCTVideoUnset = -1;
   if (_pictureInPicture == pictureInPicture) {
     return;
   }
-  
+
+  [self setupPipController];
+
   _pictureInPicture = pictureInPicture;
   if (_pipController && _pictureInPicture && ![_pipController isPictureInPictureActive]) {
     dispatch_async(dispatch_get_main_queue(), ^{
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -907,10 +907,12 @@ static int const RCTVideoUnset = -1;
 }
 
 - (void)setupPipController {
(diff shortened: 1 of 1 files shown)


cjpete (approved):
LGTM. Prevents the initial flash of PIP video.

ios/Video/RCTVideo.m
```diff
@@ -883,7 +883,9 @@ - (void)setPictureInPicture:(BOOL)pictureInPicture
   if (_pictureInPicture == pictureInPicture) {
     return;
   }
-  
+
+  [self setupPipController];
+
   _pictureInPicture = pictureInPicture;
```
watadarkstar, line 889: I believe this needs to be in the reverse order:

```suggestion
  _pictureInPicture = pictureInPicture;
  [self setupPipController];
```

ios/Video/RCTVideo.m
```diff
@@ -883,7 +883,9 @@ - (void)setPictureInPicture:(BOOL)pictureInPicture
   if (_pictureInPicture == pictureInPicture) {
     return;
   }
-  
+
+  [self setupPipController];
```
watadarkstar, line 887: I believe you should call this after you set `_pictureInPicture` to `pictureInPicture`. It should look like this:

```
[self setupPipController];
_pictureInPicture = pictureInPicture;
```

Otherwise the condition in the if statement in the function `setupController` will never be true. This one:

```
if (!_pipController && _playerLayer && _pictureInPicture && [AVPictureInPictureController isPictureInPictureSupported]) {
```

Because `_pictureInPicture` will be false unless you set it first.

KrzysztofMoch:
I will close this one, since we don't plan to make any more releases for v5 and this is not an issue for v6

PR #3062, opened 2023-03-12
fix: disable Google IMA by default on ios.

### LINKED ISSUE #3035: [V6-alpha] Why has GoogleAds been added to the dependencies?
Hello,

I've just upgraded the package to the latest `6.0.0-alpha.5` and noticed that a new GoogleAds dependency was added both to the Android and Ios native dependencies.

I mean, I understand that adding support for CSAI could be useful for some projects, but making it a mandatory dependency for the whole library seems to take us away from the main purpose of this library.

Maybe I haven't notice something so feel free to correct me 😅

freeboub:
#### Update the documentation
Done

#### Update the changelog
TODO

exemple updated

#### Describe the changes
Disable Google IMA by default and add a flag to enable it when needed
Fix: https://github.com/react-native-video/react-native-video/issues/3035

### DIFF
--- a/API.md
+++ b/API.md
@@ -74,6 +74,17 @@ Video with caching ([more info](docs/caching.md)):
 end
 ```
 
+#### Enable custom feature in podfile file
+
+##### Google IMA
+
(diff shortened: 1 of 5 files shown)


isilher:
This looks good to me. Alpha 5 did not work in my project because of the Google IMA inclusion. Targeting this commit made it work 👍

PR #3080, opened 2023-03-30
Add support for Android Media Session

phillip-kic:
# Media Session

This adds 2 props, `enableMediaSession` and `mediaSessionMetadata`. With these the video can create an Android Media Session and provide it with metadata.
Simply adding a media session like this allows the video to be controlled and respond to a number of external controls. Some examples include:
- Hardware buttons (eg. Play/Pause on a Google TV Remote)
- Google Assistant (eg. saying "Play"/"Pause"/"Stop")
- Other devices

## Testing
Because this opens the door to many controls there are number of ways it can be tested. I've been testing 2 different ways:

### Trigger pause/play event through adb
You can mimic the event of pressing a pause/play button through adb.
`adb shell input keyevent 126`
Key 126 is PLAY, 127 is PAUSE, 85 is PLAY/PAUSE. [See more event codes in the Android docs here](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_MEDIA_PLAY)

### Google Assistant
You can run on a device that has Google Assistant, then test instructions such as "Play"/"Pause"/"What's Playing?"
Pretty sure you can also use Google Assistant in emulators

### DIFF
--- a/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
+++ b/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
@@ -71,6 +71,7 @@ public class ReactExoplayerViewManager extends ViewGroupManager<ReactExoplayerVi
     private static final String PROP_HIDE_SHUTTER_VIEW = "hideShutterView";
     private static final String PROP_CONTROLS = "controls";
     private static final String PROP_ENABLE_MEDIA_SESSION = "enableMediaSession";
+    private static final String PROP_MEDIA_SESSION_METADATA = "mediaSessionMetadata";
 
     private ReactExoplayerConfig config;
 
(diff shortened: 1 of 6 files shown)


freeboub:
This Pr looks good thank you. but unfortunately, I didn't plan to make evolution on 5.2 branch.
Is it possible to move this PR to master branch ?

phillip-kic:
@freeboub I avoided using the master branch as we haven't moved over to v6 while it's in alpha. If that gets a stable release within the next few months then I'm happy to adapt this over there with some other features too

android-exoplayer/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -314,6 +316,26 @@ public void setControls(final ReactExoplayerView videoView, final boolean contro
         videoView.setControls(controls);
     }
 
+    @ReactProp(name = PROP_ENABLE_MEDIA_SESSION, defaultBoolean = false)
+    public void setEnabledMediaSession(final ReactExoplayerView videoView, final boolean enableMediaSession) {
```
s-rschie, line 320: I'm not sure, but wanted to double check just in case. Is there a spelling error here?

```suggestion
    public void setEnableMediaSession(final ReactExoplayerView videoView, final boolean enableMediaSession) {
```

s-rschie (commented):
Looks like there may be a spelling error?

KrzysztofMoch:
This is now implemented via #3723 - if I missed something please let me know

PR #3096, opened 2023-04-13
fix: Player in Player (PiP) restoreUserInterface from crashing the application

watadarkstar:
1. Create a React Native app
2. Install react-native-video using this branch
3. Create a `<Video>`  inside App.jsx with the following code:

```jsx
const VIDEO_SOURCE = {
  uri: "https://cdn81168665.blazingcdn.net/timeline/hartley-e001-s001a-01-2b6d4c/stream/index.m3u8",
}
const bufferConfig = {
  minBufferMs: 100,
  maxBufferMs: 50000,
  bufferForPlaybackMs: 100,
  bufferForPlaybackAfterRebufferMs: 100,
}
const [isPip, setPip] = useState(false)
 <Video
        paused={false}
        ref={videoRef}
        style={styles.video}
        source={VIDEO_SOURCE}
        resizeMode="contain"
        onProgress={onVideoProgress}
        controls={false}
        muted={false}
        pictureInPicture={isPip}
        onPictureInPictureStatusChanged={(
          data: OnPictureInPictureStatusData,
        ) => {
        }}
        rate={1.0}
        progressUpdateInterval={250}
        ignoreSilentSwitch="ignore"
        automaticallyWaitsToMinimizeStalling={false}
        bufferConfig={bufferConfig}
      />
   <Button onPress={() => setPip(!isPip)} />
```
4. Launch the app and enable PiP by clicking the button. Then click the button on the top right inside PiP to exit PiP. Keep doing this without any crashes (before there was a crash due to the `NSAssert`)

#### Describe the changes

Previously when exiting PiP it would crash the application due to the `NSAssert` I removed, now it won't crash the app due to this race condition.

https://stackoverflow.com/questions/1375786/whats-the-point-of-nsassert-actually

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -2020,7 +2020,6 @@ didCancelLoadingRequest:(AVAssetResourceLoadingRequest *)loadingRequest {
 }
 
 - (void)pictureInPictureController:(AVPictureInPictureController *)pictureInPictureController restoreUserInterfaceForPictureInPictureStopWithCompletionHandler:(void (^)(BOOL))completionHandler {
-  NSAssert(_restoreUserInterfaceForPIPStopCompletionHandler == NULL, @"restoreUserInterfaceForPIPStopCompletionHandler was not called after picture in picture was exited.");
   if (self.onRestoreUserInterfaceForPictureInPictureStop) {
     self.onRestoreUserInterfaceForPictureInPictureStop(@{});
   }


ios/Video/RCTVideo.m
```diff
@@ -2020,7 +2020,6 @@ - (void)pictureInPictureController:(AVPictureInPictureController *)pictureInPict
 }
 
 - (void)pictureInPictureController:(AVPictureInPictureController *)pictureInPictureController restoreUserInterfaceForPictureInPictureStopWithCompletionHandler:(void (^)(BOOL))completionHandler {
-  NSAssert(_restoreUserInterfaceForPIPStopCompletionHandler == NULL, @"restoreUserInterfaceForPIPStopCompletionHandler was not called after picture in picture was exited.");
```
watadarkstar, line 2023: @aalzanki You wrote this code 5 years ago, do you remember why you had this assertion? Its crashing our application if we leave it in like this otherwise the app works fine.

lightrow: we fixed it the same way in our variant, but i suppose we should call `playerRef.current?.restoreUserInterfaceForPictureInPictureStopCompleted(true)` after toggling off PiP, according to Apple's documentation 
https://developer.apple.com/documentation/avkit/avpictureinpicturecontrollerdelegate/1614703-pictureinpicturecontroller

> 
> completionHandler
> You must call the completion handler with a value of true to allow the system to finish restoring your player user interface.

KrzysztofMoch:
Closed due to inactivity and out dated (we don't plan to make any release for v5)

Issue #1963, opened 2020-04-01
 memory leak caused by requestAudioFocus() on ReactExoplayerView


zhaozhe-ysy:
# memory leak caused by requestAudioFocus() on ReactExoplayerView

## Platform

Which player are you experiencing the problem on:
* Android ExoPlayer

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 18.35 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.0.1 - /usr/local/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.12.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 17, 19, 21, 22, 23, 25, 26, 27, 28, 29
      Build Tools: 20.0.0, 21.1.2, 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.2, 26.0.3, 27.0.2, 27.0.3, 28.0.0, 28.0.3, 29.0.0, 29.0.1
      System Images: android-24 | Google Play Intel x86 Atom
      Android NDK: 17.1.4828580
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.4/11E146 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.60.0 => 0.60.0

```

Library version: x.x.x

## Steps To Reproduce

1. render react-native-view
2. destroy react-native-view
3. repeat 1 and 2
4. ReactExoplayerView didn't release

## Expected behaviour

1. ReactExoplayerView can release

## Cause
when requestAudioFocus() was executed, the ReactExoplayerView instance will register in AudioManager and never release. If we didn't close app and repeat render view, it may cause memory leak.
## Solution
We found a solution to fix this issue, invoke audioManager.abandonAudioFocus() in the releasePlayer() as below.
```
    private void releasePlayer() {
        if (player != null) {
            updateResumePosition();
[...]

PR #3123, opened 2023-05-09
Fix AudioManager memory leak by moving OnAudioFocusChangeListener

BasixKOR:
This PR patches RNV to no longer cause memory leaks by AudioManager holding a reference of ReactExoplayerView.

First, an explanation of how this fixes the leak: The AudioManager tracks its listeners using [`toString()` as the key of the listener.](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/media/java/android/media/AudioManager.java;drc=a5cfe59a5db654f0f9230bd63aa68f31d7cfc121;l=3949) However the listener in RNV's case, is a View, which has the toString method overridden. This causes the key to mismatch, failing to remove the listener properly on `abandonAudioFocusRequest`.

This fixes the leak by giving the listener its own class for the listener, preventing issues that arise from inheritance. After this patch memory leak should be fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### Version 6.0.0-alpha.7
+- Android: Fix memory leaks from AudioManager [#3123](https://github.com/react-native-video/react-native-video/pull/3123)
+
 ### Version 6.0.0-alpha.6
 - Feature: Video range support [#3030](https://github.com/react-native-video/react-native-video/pull/3030)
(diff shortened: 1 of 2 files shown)


BasixKOR:
The same fixes can be applied to 5.X line as well; fixing #1963.

freeboub:
Thank you for this change! Is it possible to extract the OnAudioFocusChangedListener to a dedicated file please ?

BasixKOR:
That would be great, I'll work on it this week!

BasixKOR:
It seems like it would require some level of refactoring to decouple the listener completely, as it highly depends on the internals of the view. Although the refactoring of the project would be great I think it would be better reviewed separately from this PR. What do you think?

PR #3153, opened 2023-06-17
fixes balck splash, thumbnail when used with textureview

### LINKED ISSUE #3148: provide prop to set shutterView background color
hideShutterView prop not working as expected flashes black screen then video played, setting shutterView background color to transparent working as expected.

https://github.com/google/ExoPlayer/issues/2843#issuecomment-363148181
`simpleExoPlayerView.setShutterBackgroundColor(Color.TRANSPARENT);`

vemarav:
fixes https://github.com/react-native-video/react-native-video/issues/3148

Added props `frameQuality` and `onFrameChange` to get current frame thumbnail.

All three props work only when useTextureView={true} i.e., default 

Platforms: Android

#### shutterColor
Apply color to shutter view, if you see black flashes before video start then set shutterColor='transparent'

won't work if useTextureView={false}

* **black (default)**

Platforms: Android

#### frameQuality
Use this prop to set quality of Base64 encoded string image obtained by onFrameChange callback. 

won't work if useTextureView={false}

* **undefined (default)** - range [0-100]

Platforms: Android

#### onFrameChange
Callback function that is fired when video frame changes
It works only when frameQuality prop is set, return Base64 encoded string image.

won't work if useTextureView={false}

Example

```
  <Video
    frameQuality={30}
    onFrameChange={(imageString) => {
      setImageSource(`data:image/jpeg;base64,${imageString}`)
    }
  />
```

### DIFF
--- a/API.md
+++ b/API.md
@@ -310,6 +310,9 @@ var styles = StyleSheet.create({
 | [fullscreenOrientation](#fullscreenorientation)                                     | iOS                       |
 | [headers](#headers)                                                                 | Android                   |
 | [hideShutterView](#hideshutterview)                                                 | Android                   |
+| [shutterColor](#shutterColor)                                                       | Android                   |
+| [frameQuality](#frameQuality)                                                       | Android                   |
+| [onFrameChange](#onFrameChange)                                                     | Android                   |
 | [ignoreSilentSwitch](#ignoresilentswitch)                                           | iOS                       |
 | [maxBitRate](#maxbitrate)                                                           | Android, iOS              |
 | [minLoadRetryCount](#minLoadRetryCount)                                             | Android                   |
@@ -581,6 +584,43 @@ Controls whether the ExoPlayer shutter view (black screen while loading) is enab
 
 * **false (default)** - Show shutter view 
 * **true** - Hide shutter view
+
+Platforms: Android
+
+#### shutterColor
+Apply color to shutter view, if you see black flashes before video start then set shutterColor='transparent'
+
+won't work if useTextureView={false}
+
+* **black (default)**
+
+Platforms: Android
+
+#### frameQuality
+Use this prop to set quality of Base64 encoded string image obtained by onFrameChange callback. 
+
+won't work if useTextureView={false}
+
+* **undefined (default)** - range [0-100]
+
+Platforms: Android
+
+#### onFrameChange
+Callback function that is when video frame changes
+It works only when frameQuality prop is set, return Base64 encoded string image.
+
+won't work if useTextureView={false}
+
+Example
+
+```
+  <Video
+    frameQuality={30}
+    onFrameChange={(imageString) => {
+      setImageSource(`data:image/jpeg;base64,${imageString}`)
+    }
+  />
+```
 
 Platforms: Android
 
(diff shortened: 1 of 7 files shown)


API.md
```diff
@@ -310,6 +310,9 @@ var styles = StyleSheet.create({
 | [fullscreenOrientation](#fullscreenorientation)                                     | iOS                       |
 | [headers](#headers)                                                                 | Android                   |
 | [hideShutterView](#hideshutterview)                                                 | Android                   |
+| [shutterColor](#shutterColor)                                                       | Android                   |
```
freeboub, line 313: can you keep alphabetical order in doc please ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -244,12 +248,29 @@ public void handleMessage(Message msg) {
...
+                        setFrame();
                     }
                     break;
             }
         }
     };
 
+    private void setFrame() {
+        if(frameQuality == null) return;
+        ((Runnable) () -> {
```
freeboub, line 260: I don't understand why you use a runnable here.
I think you want to do something asynchroniously. If that is the case, you should also use a Handler ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -626,8 +647,10 @@ private void initializePlayerCore(ReactExoplayerView self) {
         );
         DefaultRenderersFactory renderersFactory =
                 new DefaultRenderersFactory(getContext())
+                        .setEnableDecoderFallback(true)
```
freeboub, line 650: I think this can lead to performancies issue, can you disable it if user uses SurfaceView ?
But maybe also if frameQuality is null ? No sure this is feasible because you will need to change the value if user set dynamically frameQuality.

freeboub:
Thank you for this PR, small general comment, it would easier to manage if you open multiple small PRs.
(1 for shutterView color and 1 for reporting)

API.md
```diff
@@ -584,6 +587,43 @@ Controls whether the ExoPlayer shutter view (black screen while loading) is enab
 
 Platforms: Android
 
+#### shutterColor
+Apply color to shutter view, if you see black flashes before video start then set shutterColor='transparent'
+
+won't work if useTextureView={false}
```
freeboub, line 593: Is it true ?

API.md
```diff
@@ -584,6 +587,43 @@ Controls whether the ExoPlayer shutter view (black screen while loading) is enab
...
+
+* **black (default)**
+
+Platforms: Android
+
+#### frameQuality
+Use this prop to set quality of Base64 encoded string image obtained by onFrameChange callback.
```
freeboub, line 600: can you info that onFrameChange will be reported with same frequency than onProgress ?
No sure this is good, time position should be fast, frame callback can be slower

API.md
vemarav: default is true

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
vemarav: no effect if surfaceview is used

API.md
vemarav: Tested on lower end devices the frame drop will increase as frameQuality increase. But keeping frameQuality=30 will feasible and one can get all the frames.

This functionality is needed in Flatlist as user can start from where he left.
frameCallback will be slower but the last frame will be the same where user paused the video that's why I have used runnable thread.

vemarav:
Will create new PR with small changes

Issue #3085, opened 2023-04-04
onFullscreen call backs are never fired


mmckinley8:
# Bug

## Platform

Which player are you experiencing the problem on:
* iOS
* Android

## Environment info

React native info output:

```bash
System:
    OS: macOS 13.2.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 76.05 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.13.1 - /var/folders/hz/dn9pznz96_1frz690md0l0100000gn/T/yarn--1680642584961-0.038080415682131274/node
    Yarn: 1.22.19 - /var/folders/hz/dn9pznz96_1frz690md0l0100000gn/T/yarn--1680642584961-0.038080415682131274/yarn
    npm: 8.1.2 - /usr/local/bin/npm
    Watchman: 2022.01.17.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    Android SDK:
      API Levels: 28, 29, 30, 31, 32
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.0, 30.0.2, 30.0.3, 31.0.0, 34.0.0
      System Images: android-30 | Android TV Intel x86 Atom, android-30 | China version of Wear OS 3 - Preview Intel x86 Atom, android-30 | Wear OS 3 Intel x86 Atom, android-30 | Google TV Intel x86 Atom, android-30 | Google APIs ARM 64 v8a, android-30 | Google APIs Intel x86 Atom, android-30 | Google APIs Intel x86 Atom_64, android-30 | Google Play ARM 64 v8a, android-30 | Google Play Intel x86 Atom, android-30 | Google Play Intel x86 Atom_64, android-31 | Google APIs Intel x86_64 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9619390
    Xcode: 14.3/14E222b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.15 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0
[...]

Issue #3040, opened 2023-02-20
[Bug] Fullscreen broken on iOS due to #3017


jspizziri:
# Bug

The [check for `_controls == true` introduced here](https://github.com/react-native-video/react-native-video/commit/2032a36969a022c57232baa8f3b9746eff103c28#diff-1536a22e8e888a1d8bccd1916bfec1af06373dd52ddb442c09544718e06a7193R617) results in fullscreen presentation not working on iOS unless the `controls={true}` prop is passed to the video player. Perhaps this was intentional, however, if that is the case when setting that prop [the RNSScreen parent view controller](https://github.com/react-native-video/react-native-video/issues/2808) bug manifests itself.

I'm digging into it now but insights would be appreciated (first look under the hood at this library).

## Platform

iOS

## Environment info

Library version: `6.0.0-alpha.5`

## Steps To Reproduce

1. Embed a Video view on iOS
2. set `controls={true}`
3. Attempt to trigger fullscreen mode
4. Also, see that with `controls={false}` fullscreen presentation fails quietly due to [no `_playerViewController` being instantiated.](https://github.com/react-native-video/react-native-video/blob/master/ios/Video/RCTVideo.swift#L639)

## Expected behaviour

1. Video player presents fullscreen correctly on iOS.

jspizziri:
To add to the above, it appears that when `controls = false` there is no crash resulting from the `RNSScreen` exception. So it appears that the combination of requiring `controls = true` AND the crash when presenting due to `controls = true` is what renders fullscreen entirely broken on iOS. If you revert to `v6.0.0-alpha.4` and apply [the patch issued here](https://github.com/react-native-video/react-native-video/pull/2971/files) AND have `controls = false` THEN fullscreen presentation works on iOS.

As a result and [along with this issue](https://github.com/react-native-video/react-native-video/issues/3035), I've ended up reverting to `alpha.4` and applying a patch for now.

Here's the `alpha.4` patch in case anyone is interested: [react-native-video+6.0.0-alpha.4.patch](https://github.com/react-native-video/react-native-video/files/10785110/react-native-video%2B6.0.0-alpha.4.patch)

kyle-ssg:
After playing around with this, it seems that the following fixes this issue in ``RCTVideo.swift``:

```
      if let playerViewController = _playerViewController {
                    if(_controls) { //Necessary for preventing crash
                        self._playerViewController?.removeFromParent()
                    }

[...]

PR #3167, opened 2023-07-06
iOS fixes

### LINKED ISSUE #3166: [iOS] Memory leaks
# Bug
Mount then unmount Video component result in swift init called but deinit is never called 

## Platform
* iOS

## Environment info

React native info output:
```bash
System:
  OS: macOS 13.2
  CPU: (8) x64 Apple M1
  Memory: 25.48 MB / 16.00 GB
  Shell:
    version: 5.8.1
    path: /bin/zsh
Binaries:
  Node:
    version: 18.9.1
    path: /var/folders/m6/g_g8x9mx6bsdd0z4bz2h649r0000gn/T/yarn--1688635068292-0.5680216196061589/node
  Yarn:
    version: 1.22.19
    path: /var/folders/m6/g_g8x9mx6bsdd0z4bz2h649r0000gn/T/yarn--1688635068292-0.5680216196061589/yarn
  npm:
    version: 8.19.1
    path: /usr/local/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.12.1
    path: /Users/***/.rvm/gems/ruby-2.7.6/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.4
      - iOS 16.4
      - macOS 13.3
      - tvOS 16.4
      - watchOS 9.4
  Android SDK:
    API Levels:
      - "23"
      - "25"
      - "28"
      - "29"
      - "30"
      - "31"
      - "32"
      - "33"
    Build Tools:
      - 28.0.3
      - 29.0.2
      - 29.0.3
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
    System Images:
      - android-31 | Google Play ARM 64 v8a
    Android NDK: 22.1.7171670
IDEs:
  Android Studio: 2021.3 AI-213.7172.25.2113.9123335
  Xcode:
    version: 14.3/14E222b
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 11.0.10
    path: /usr/bin/javac
  Ruby:
    version: 2.7.6
    path: /Users/***/.rvm/rubies/ruby-2.7.6/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.1
    wanted: 0.72.1
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false

```

Library version: 6.0.0-alpha.6

## Steps To Reproduce

1. Navigate to / Open a screen with a Video
2. Start playing the video
3. Go back / Unmount Video component
4. Video is still running

## Expected behaviour

Video should stop and RCTVideo should be deinit

cguino:
#3166 Memory leaks
#3085 onFullscreen call backs are never fired
#3040 [Bug] Fullscreen broken on iOS due to #3017
#2744 Unblock the main thread while replacing the source ([cf comment](https://github.com/react-native-video/react-native-video/issues/2744#issuecomment-1219444287))

### DIFF
--- a/ios/Video/Features/RCTIMAAdsManager.swift
+++ b/ios/Video/Features/RCTIMAAdsManager.swift
@@ -4,7 +4,7 @@ import GoogleInteractiveMediaAds
 
 class RCTIMAAdsManager: NSObject, IMAAdsLoaderDelegate, IMAAdsManagerDelegate {
 
-    private var _video:RCTVideo
+    private weak var _video: RCTVideo?
 
     /* Entry point for the SDK. Used to make ad requests. */
     private var adsLoader: IMAAdsLoader!
@@ -23,6 +23,7 @@ class RCTIMAAdsManager: NSObject, IMAAdsLoaderDelegate, IMAAdsManagerDelegate {
     }
 
     func requestAds() {
+        guard let _video = _video else {return}
         // Create ad display container for ad rendering.
         let adDisplayContainer = IMAAdDisplayContainer(adContainer: _video, viewController: _video.reactViewController())
 
@@ -54,6 +55,7 @@ class RCTIMAAdsManager: NSObject, IMAAdsLoaderDelegate, IMAAdsManagerDelegate {
     // MARK: - IMAAdsLoaderDelegate
 
     func adsLoader(_ loader: IMAAdsLoader, adsLoadedWith adsLoadedData: IMAAdsLoadedData) {
+        guard let _video = _video else {return}
         // Grab the instance of the IMAAdsManager and set yourself as the delegate.
         adsManager = adsLoadedData.adsManager
         adsManager?.delegate = self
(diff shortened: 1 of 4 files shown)


freeboub:
thank for this PR, it looks good except https://github.com/react-native-video/react-native-video/issues/2744 Unblock the main thread while replacing the source (https://github.com/react-native-video/react-native-video/issues/2744#issuecomment-1219444287)

We already experiment this change, but it causes crash when fast changing source...
can you revert this change and I will merge this soon.

The good fix has be be discussed (I think I know the good fix but this is a big architecture change ...)

cguino:
Thank you, I revert #2744

marcel-happyfloat:
@freeboub the main thread blocking makes this library unusable for us in the real world with bad internet connection trying to scroll a feed of video items. Do you think the fix is safe to use for multiple video elements, with only one fixed source at a time? (or mounting/unmounting with new source). Thank you.
I really appreciate all your work, unfortunately I can't help debugging on the native side.

Trashpants:
Also in the same boat with the blocking main thread, without knowing too much about the internals can we have this as an optional so additional prop such as dangerouslySwapIosVideoSource as a bool on the player component?

freeboub:
I totally understand the issue of startup I revert following change: 1f27ffbc8101044eb59866886fec6fd12223febf
which was causing the crash. I really think it will cause the same issue.
This code also causes an issue when rapidly changing source...
There is clearly a concurrency issue in react native video.
I fixed it on android in my company fork but it is a nightmare to backport, I need to find time to reimplement it here.
moreover the PR of new arch will also conflict with it :/
Let see.
Notice that I advise you to patch the package (with patch-package) or create your own fork if you really want this change.
Personally I prefer latency than crash ...

Issue #3058, opened 2023-03-09
react native Video with flatList (with multiple video) on Scrolling App not responding and crashed the app


iasreact1:
I am using react-native-video-5.2.1, don't know whether use android exo-Player or mediaPlayer. There are so many videos in the flatlist and when get Scrolled it get stuck the UI and app not responding error, Please help what to do for solving it, but if i use the video plaer to another page for playing the video then it working fine but when playing the video within the Flatlist ANR is coming every time & every Device, Please help ...to overcome it..

LucasHimelfarb:
Your UI was blocked because Your UI is likely freezing because the listing is not optimized, either because you are displaying many heavy elements or because there is a design anti-pattern in your code causing multiple renders depending on state variables that are being updated. But without a code sample, it's difficult to know for sure.

iasreact1:
@LucasHimelfarb  i have an api call 10 data per page and there is multiple state variable for video play/pause, mute/unmute etc loading opacity, and if i use a log that will print 10 times for the first rendering ...how can i improve the performance?

krunalinfynno:
i am getting same issue as well.

LucasHimelfarb:
@iasreact1 @krunalinfynno  Im sorry for the late answer. The Flatlist`ve many props to improve perfomance, for example the props "removeClippedSubviews" and "getItemLayout". You can see all props here: https://reactnative.dev/docs/flatlist

If anyone have some issue please share some code to detect common problems

trubnikovr:
@LucasHimelfarb 
I'm using a modal with a FlatList inside it as a horizontal slider. I've noticed that if I slide and the video is still playing, then the app crashes.

freeboub:
If someone can provide a sample app, it would be very useful to reproduce (and fix) the issue!

EuSet:
here is a stack trace if it can be helpful 
![Uploading Screenshot 2024-05-26 at 00.20.42.png…]()

EuSet:
here is a stack trace if it can be helpful
![trace](https://github.com/TheWidlarzGroup/react-native-video/assets/79215069/ba5ab067-a244-4327-bb59-791856ef7338)

EuSet:
here is a stack trace if it can be helpful. version 6.1.1
![trace](https://github.com/TheWidlarzGroup/react-native-video/assets/79215069/187e210d-5485-4ad4-8c62-ee03c0114175)

freeboub:
[...]

PR #3173, opened 2023-07-12
fix ios block ui issue

### LINKED ISSUE #2986: Video freeze UI until video is loaded on IOS, scroll is stuck
# Bug

Happy new year to all !

I am the creator of **SoSkills** app which is available on **Apple** and **Android**
**SoSkills** is a video sharing community application in the gaming ecosystem.
More info here [https://soskills.io](https://github.com/react-native-video/react-native-video/issues/url)

Keep in mind user can swipe through videos on the Home page with FlatList as Instagram short, TikTok etc

When user swipe to the bottom, app is going to load new videos by doing a request to the server for having videos object from BDD with uri property which is the path of the video on Amazon, so the file is not get by my server but directly with react-native-video (uri is with https)
So when user swipe, new item in the Flatlist are rendered and Video component start to init video with new uri from Amazon.
Request take some time when there's an average connexion or bad connexion, UI will freeze, app will be stuck and user can't do anything on the app until video is loaded by react-native-video.

No hidden process or work is doing during the scroll, just react-native-video is blocking the JS thread and UI freeze

## Platform

- Android with Redmi Note 10S phone
- IOS with Iphone 12 pro

## Environment info
Screenshot 2023-01-04 at 15 42 58

React native: 0.67.3
Library version: [v6.0.0-alpha.4]

## Steps To Reproduce

Use average connexion or bad connexion which is the best scenario.
Create a FlatList with with 3 item where item is taking all the viewport so one item by viewport.
Get new data when item + 1 === item.length and get new data, ui will freeze.

## Expected behaviour

UI with no freeze, scroll still possible even if Video component is loading the video as Android with their video player.
On Android, if video is not loaded, user has just to wait but if he wants to scroll he can, without being stopped.

## Reproducible sample code

**SoSkills** is a private company, code is not open source so can't do that but i shared screenshot and video

## Video sample

This example is done with my Iphone 12 pro on 4g which is average so if you don't have good connexion, it's really not good ! 

https://user-images.githubusercontent.com/3983160/210596138-0d291326-06ba-4cae-bfcc-85e260be4add.mp4

vemarav:
closes #2986 #3058

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -563,6 +563,10 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
 
     func applyModifiers() {
+        if let video = _player?.currentItem,
+            video == nil || video.status != AVPlayerItem.Status.readyToPlay {
+            return
+        }
         if _muted {
             if !_controls {
                 _player?.volume = 0


vemarav:
@freeboub tested on fullscreen & flatlist

marcel-happyfloat:
Works in my app 👍. Thank you very much!

vemarav:
> Works in my app 👍. Thank you very much!

make your users happy with delightful experience 🚀

stefanahman:
This solves the issue (on my physical phone), but now I have a problem of streaming videos in the simulator while developing. Anyone knows the issue behind this?

vemarav:
@stefanahman 
> This solves the issue (on my physical phone), but now I have a problem of streaming videos in the simulator while developing. Anyone knows the issue behind this?

Use ios 16.1 or lower on simulator

ColtraneNadler:
what version is this fixed in ? 
im currently using `5.2.1` and it freezes the UI momentarily when starting to play the video.

freeboub:
@ColtraneNadler try with v6.0.0-RC0

EuSet:
faced this issue using 6.1.1 version

himanshut230:
facing this issue in (v6.0.0, v6.1.2) versions. In v6.0.0-alpha.7 it is working fine but there are other issues in this version and it is not stable.

gerson:
I'm having the issue in (v6.0.0, v6.1.2) versions. In v6.0.0-alpha.7 it is working fine cannot go back since there are another issues in this version and I would like to use the stable one, any update?

PR #3179, opened 2023-07-23
[Android] fixes black splashes

### LINKED ISSUE #1128: (Android) Black Screen for a second between loading and video playback

vemarav:
closes #1128 #3148 
#### shutterColor
Apply color to shutter view, if you see black flashes before video start then set shutterColor='transparent'

won't work if useTextureView={false}

black (default)
Platforms: Android

### DIFF
--- a/API.md
+++ b/API.md
@@ -331,6 +331,7 @@ var styles = StyleSheet.create({
 | [selectedAudioTrack](#selectedaudiotrack)                                           | Android, iOS              |
 | [selectedTextTrack](#selectedtexttrack)                                             | Android, iOS              |
 | [selectedVideoTrack](#selectedvideotrack)                                           | Android                   |
+| [shutterColor](#shutterColor)                                                       | Android                   |
 | [source](#source)                                                                   | All                       |
 | [subtitleStyle](#subtitleStyle)                                                     | Android                   |
 | [textTracks](#texttracks)                                                           | Android, iOS              |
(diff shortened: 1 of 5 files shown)


freeboub:
Thank you for the split ! @vemarav

alexhernandez:
Hey All 👋,
Big fan of all the work you do!

Chiming in since I see this ticket is closed but v6-Alpha11 and v6-Beta0 both seem to have this issue still. Is `hideShutterview` `shutterColor` intended only for TextureViews? It currently doesn't seem to work otherwise

vemarav:
> Hey All 👋, Big fan of all the work you do!
> 
> Chiming in since I see this ticket is closed but v6-Alpha11 and v6-Beta0 both seem to have this issue still. Is `hideShutterview` `shutterColor` intended only for TextureViews? It currently doesn't seem to work otherwise

@alexhernandez  Yes, it's intended only for textureview, for surface view you can patch here https://github.com/react-native-video/react-native-video/blob/4d9334b4771f28c51f5410e17702e44e08524d28/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java#L126

if patch worked then use `npx patch-package react-native-video`

PR #3216, opened 2023-08-25
feat: allow customization of tvOS playback expierence

Duell10111:
This PR would allow to customize the tvOS playback expierence as descripted here:
https://developer.apple.com/documentation/avkit/customizing_the_tvos_playback_experience#3820220

The custom title, subtitle and description are all optional and can specified in the source object.
These props are not handled on android side.

Additionally this PR allows to adapt the chapters interface for tvOS, according to the documentation: 
https://developer.apple.com/documentation/avkit/presenting_navigation_markers#2948687
For providing custom chapters, a new prop chapters got introduced.

The changes can be tested in the tvOS example project, by running on an tvOS simulator.


ios/Video/RCTVideo.swift
```diff
@@ -371,7 +372,89 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+        
+        if let description = _source?.description {
+            mapping[.commonIdentifierDescription] = description
+        }
+        
+        if #available(iOS 12.2, *), !mapping.isEmpty {
+            playerItem.externalMetadata = createMetadataItems(for: mapping)
+        }
+        
+        #if os(tvOS)
```
freeboub, line 397: Maybe this #if can be earlier in this function to avoid unnecessary computation on mobile ?

ios/Video/RCTVideo.swift
```diff
@@ -371,7 +372,89 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+        
+        return playerItem
+    }
+    
+    func createMetadataItems(for mapping: [AVMetadataIdentifier: Any]) -> [AVMetadataItem] {
+        return mapping.compactMap { createMetadataItem(for:$0, value:$1) }
+    }
+    
+    #if os(tvOS)
+    private func makeNavigationMarkerGroups(_ chapters: [Chapter]) -> [AVNavigationMarkersGroup] {
```
freeboub: It would be good to extract these new functions in a new file... 
This one is already too big.

freeboub:
@Duell10111  I forgot to publish my review :'( 
I think this PR can be merged soon. please let me know

ios/Video/RCTVideo.swift
Duell10111: I outsourced the functions to a new file

ios/Video/RCTVideo.swift
Duell10111: The external metadata can also be set on iOS to replace the asset metadata, therefore I only restricted the navigationMarker Groups to tvOS as this is only working on tvOS.
https://developer.apple.com/documentation/avfoundation/avplayeritem/1627623-externalmetadata

Duell10111:
@freeboub Thanks for the review, I outsourced the functions to Util files to reduce the code in the RCTVideo class.
I left the externalMetadata feature available on iOS as this could maybe also be helpful on iOS devices.

So from my POV the PR can be merged. :)

PR #3241, opened 2023-09-17
Fix: Not showing video ads when the PIP mode is enabled on iOS

lrusso:
Fix: Not showing video ads when the PIP mode is enabled on iOS.

According to the Google IMA SDK, you cannot show video ads when the PIP mode is enabled (
https://developers.google.com/interactive-media-ads/docs/sdks/ios/client-side/picture_in_picture?hl=en#starting_ads).

### DIFF
--- a/API.md
+++ b/API.md
@@ -667,6 +667,8 @@ Determine whether the media should played as picture in picture.
 * **false (default)** - Don't not play as picture in picture
 * **true** - Play the media as picture in picture
 
+NOTE: Video ads cannot start when you are using the PIP on iOS (more info available at [Google IMA SDK Docs](https://developers.google.com/interactive-media-ads/docs/sdks/ios/client-side/picture_in_picture?hl=en#starting_ads)). If you are using custom controls, you must disable your PIP button when you receive the ```STARTED``` event from ```onReceiveAdEvent``` and show it again when you receive the ```ALL_ADS_COMPLETED``` event.
+
 Platforms: iOS
 
 #### playInBackground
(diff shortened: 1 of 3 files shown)


freeboub:
I understand the necessity of this change, but:
- That would need to be useful to add comment in doc indicating this restriction
- What would happen is application enable pip during playback ? during Ad or during source playback ?

lrusso:
Hi @freeboub 

**That would need to be useful to add comment in doc indicating this restriction**

Yes, that would be a good thing, I could add that note to the documentation file that you think appropriate.

**What would happen is application enable pip during playback ? during Ad or during source playback ?**

There are two escenarios:

1) If you are watching a video (in normal mode) and a video ad appears, the default video controls are going to be hidden, so you won't be able to enable or trigger the pip mode while you are watching the video ad.
2) If you are watching a video (in normal mode) and a video ad appears, the default video controls are going to be hidden, but if you have custom controls, those should be hidden, because if your custom controls are still visible, the end user is going to be able to trigger the pip mode while watching a video ad.

Without this pr/fix, when you are watching a video and you enable the pip mode and you are watching your video on the floating pip window, and then a video ad shows up, that video ad (created by the Google IMA SDK) won't show up on the pip window. The video ad will show up in your app, where the video in normal mode supposed to be. That's why Google says that video ads cannot start while you are using the PIP mode.

freeboub:
Thanks you for details !
Can you add 1 note in pip section https://github.com/react-native-video/react-native-video/blob/master/API.md#pictureinpicture and in ads section: near: $RNVideoUseGoogleIMA usage

Thank you

lrusso:
@freeboub I just added a note regarding the PIP mode on iOS to the API.md file

PR #3245, opened 2023-09-18
feat: expose playback functions

### LINKED ISSUE #3097: Add a play/pause method to programmatically play/pause the video
This is useful in many circumstances. For example when you want to have custom playback controls.

KrzysztofMoch:
## Summary
Allow to play/pause video from component reference

closes #3097

## Example
```js
videoRef.current.pause()
videoRef.current.play()
```

## Test plan
- [x] Functions are working on iOS
- [x] Functions are working on Android


API.md
```diff
@@ -1643,6 +1643,47 @@ this.player.seek(120, 50); // Seek to 2 minutes with +/- 50 milliseconds accurac
 
 Platforms: iOS
 
+### setPlayerStatus
```
freeboub: Is it possible to remove this api from doc ?
OK to use an internal common function, but each new api shall be maintain, more api is officially available more maintenance is necessarry.
Morover can you rename to : setPlayerPauseState. here it is a too generic name (for internal use)

API.md
```diff
@@ -1643,6 +1643,47 @@ this.player.seek(120, 50); // Seek to 2 minutes with +/- 50 milliseconds accurac
...
+Example:
+```
+this.player.setPlayerStatus(true); // Play the video
+this.player.setPlayerStatus(false); // Pause the video
+```
+
+platforms: Android, iOS
+
+### pause
+`pause(): Promise<void>`
```
freeboub, line 1649: do it really need to return a Promise as there is no result ?

Video.js
```diff
@@ -14,8 +14,8 @@ const styles = StyleSheet.create({
   },
 });
 
-const { VideoDecoderProperties } = NativeModules
-export { TextTrackType, FilterType, DRMType, VideoDecoderProperties }
+const { VideoDecoderProperties } = NativeModules;
+export { TextTrackType, FilterType, DRMType, VideoDecoderProperties };
 
 export default class Video extends Component {
```
freeboub, line 21: changes are not related this this feature, can you revert them please ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -116,7 +116,7 @@
 import java.lang.Integer;
 
 @SuppressLint("ViewConstructor")
-class ReactExoplayerView extends FrameLayout implements
+public class ReactExoplayerView extends FrameLayout implements
```
freeboub, line 119: really needed ?

android/src/main/java/com/brentvatne/react/ReactVideoPackage.java
```diff
@@ -9,6 +9,7 @@
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.uimanager.ViewManager;
 
+import java.util.ArrayList;
```
freeboub, line 12: no need of this import

android/src/main/java/com/brentvatne/react/ReactVideoPackage.java
```diff
@@ -25,9 +26,12 @@ public ReactVideoPackage(ReactExoplayerConfig config) {
 
     @Override
     public List<NativeModule> createNativeModules(ReactApplicationContext reactContext) {
-        return Collections.singletonList(
-                new VideoDecoderPropertiesModule(reactContext)
-        );
+        List<NativeModule> modules = new ArrayList<NativeModule>();
```
freeboub, line 29: was their an issue here ?

android/src/main/java/com/brentvatne/react/VideoManagerModule.java
```diff
@@ -0,0 +1,39 @@
+package com.brentvatne.react;
+
+import android.view.View;
+
+import androidx.annotation.NonNull;
+
```
freeboub, line 6: why do you need a new module ? can't you do like presentFullScreen apis ?

freeboub:
additionally, can you remove changes in sample please ?
I agree that linter is not correctly configured in samples, but it makes this change difficult to review.
It would need a dedicated PR to do that (see also: https://github.com/react-native-video/react-native-video/pull/3015)

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
Okelm: let's try to avoid making public classes like this to not expose too much for other developers who will be using this class

API.md
Okelm: If you want to use this API with await and in general in async chains I guess you need a promise as a return type 🤔

Video.js
Okelm: When it comes to adding ; at the end of these two exports my taking would be: don't do unnecessary refactors or code clean ups as this may obscure the real issue we want to solve. 

But in this case it's basically a scouting rule applied "“Always leave the code you are working on a little bit better than you found it.”" in the most minimalistic way so I would personally leave it in the PR

Okelm (changes requested):

API.md
KrzysztofMoch: Yeah, it return promise to let know user that this function is async

Video.js
KrzysztofMoch: @freeboub This is change made by prettier "on save". Do we really need to revert this ? If Yes then should I disable prettier completely ?

android/src/main/java/com/brentvatne/react/VideoManagerModule.java
KrzysztofMoch: Actually there is no `VideoManager` module in android now. To maintain unity between platforms, it seems to me that this is a good solution. But lets keep open this discussion

android/src/main/java/com/brentvatne/react/ReactVideoPackage.java
KrzysztofMoch: With `Collections.singletonList` i could not add `VideoManagerModule` to list. This implementation is from react-native [docs](https://reactnative.dev/docs/native-modules-android?android-language=java)

android/src/main/java/com/brentvatne/react/ReactVideoPackage.java
KrzysztofMoch: Why are using it at line 29 (ReactVideoPackage)

```java
 List<NativeModule> modules = new ArrayList<NativeModule>();
 ```

File: android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java (hunk 4 above)
Comment (KrzysztofMoch) [reply android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java:119]: But we need this to be public so we can import `ReactExoplayerView` in `VideoManagerModule`

File: Video.js [hunk 8]
```diff
@@ -81,6 +81,18 @@ export default class Video extends Component {
     return await NativeModules.VideoManager.save(options, findNodeHandle(this._root));
   }
 
+  pause = async () => {
+    await this.setPlayerPauseState(false);
+  };
+
+  play = async () => {
+    await this.setPlayerPauseState(true);
```
freeboub: sorry but I think naming is not correct here ...
When I call: setPlayerPauseState(true); I am expecting the playback to be paused... It is possible to invert the logic please ?

Video.js
freeboub: only to lines, That's ok let's keep it

API.md
freeboub: Ok, but the question is why is it needed to be async ?
Here you bring more complexity to something which can be straight forward.
if user really need to delay it he use its own promise.
There is nothing interesting with promise: I mean waiting for the promise result doesn't bring interesting information to the user (there is be another event indicating player is paused).

That said I would be curious to have a performance comparaison  between 3 solutions:
- using the pause property
- using direct call (the solution I propose)
- using direct call embedded in promise
- using async call (the solution you implement)

Anyway I agree to keep it like this.
Thank you

Video.js
KrzysztofMoch: Agree - fixed ✅

freeboub (approved):

freeboub:
@KrzysztofMoch can you fix conflicts please ? This is Ok to be merged on my side, Thank you

freeboub (approved):

PR #3246, opened 2023-09-18
Remove explicit SDK version for Windows builds

chrisglein:
Fix SDK build error if 10.0.16299.0 is not installed:
```
 i Build configuration: Debug
 i Build platform: x64
 × Build failed with message 4:10>C:\Users\redactedUser\.nuget\packages\microsoft.ui.xaml\2.8.0\build\Common.targets(10,5): error : Microsoft.UI.Xaml nuget package requires TargetPlatformMinVersion >= 10.0.17763.0 (current project is 16299) [D:\GitHub\redacted\node_modules\react-native-video\windows\ReactNativeVideoCPP\ReactNativeVideoCPP.vcxproj]. Check your build configuration.
```

See https://github.com/react-native-async-storage/async-storage/pull/810 for a similar fix.
And these
- https://github.com/react-native-picker/picker/pull/483
- https://github.com/react-native-clipboard/clipboard/pull/177
- https://github.com/software-mansion/react-native-screens/pull/1890

Previously, after bringing down package dependencies you'd have to retarget the SDK version for the module project.

This change was tested by applying a patch-package of the vcxproj to an RNW app and doing a clean build. Basically, if it builds, it's good. No runtime effect.

- [x] After you open the PR, update the CHANGELOG.md file with an entry pointing to your PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 
 ## Next
+- Windows: fix build error from over-specified SDK version [#3246](https://github.com/react-native-video/react-native-video/pull/3246)
 
 ### Version 6.0.0-alpha.8
 - All: Playing audio over earpiece [#2887](https://github.com/react-native-video/react-native-video/issues/2887)
--- a/windows/ReactNativeVideoCPP/ReactNativeVideoCPP.vcxproj
+++ b/windows/ReactNativeVideoCPP/ReactNativeVideoCPP.vcxproj
@@ -13,13 +13,16 @@
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
     <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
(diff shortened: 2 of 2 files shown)


freeboub:
This is ok for me, thank you for these pull requests. Just a question, does it imply some necessary change on the application side to configure the sdk ? I think no, but I am asking as I am in doubt !

jonthysell (approved):

chrisglein:
> This is ok for me, thank you for these pull requests. Just a question, does it imply some necessary change on the application side to configure the sdk ? I think no, but I am asking as I am in doubt !

Nope, no change needed on the application side. If the app has an SDK version specified (which it absolutely will have) then this module will match that SDK version with this change. Previously it would fight with the app's specified SDK and you'd need to have both SDK versions installed in order to build. Which could have been transparent if you happened to have multiple SDKs installed, or would be a nuisance to prompt you to go install an SDK version you didn't _really_ need (and also be a drain on your disk space 😉).

Issue #3172, opened 2023-07-11
only audio and black screen on both emulators for iso and android


ivan-vilches-5:
## Platform
Which player are you experiencing the problem on:
* iOS
* Android

## Environment info

npx react-native info         
info Fetching system and libraries information...
System:
  OS: macOS 13.4
  CPU: (8) arm64 Apple M1
  Memory: 72.91 MB / 8.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.4.0
    path: /opt/homebrew/bin/node
  Yarn:
    version: 1.22.19
    path: /opt/homebrew/bin/yarn
  npm:
    version: 9.7.2
    path: /opt/homebrew/bin/npm
  Watchman:
    version: 2023.07.03.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.12.1
    path: /Users/ivanvilches/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 22.4
      - iOS 16.4
      - macOS 13.3
      - tvOS 16.4
      - watchOS 9.4
  Android SDK:
    API Levels:
      - "33"
      - "34"
    Build Tools:
      - 30.0.3
      - 33.0.0
      - 34.0.0
    System Images:
      - android-34 | Google APIs ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2022.2 AI-222.4459.24.2221.10121639
  Xcode:
    version: 14.3.1/14E300c
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 11.0.19
    path: /opt/homebrew/opt/openjdk@11/bin/javac
  Ruby:
    version: 3.2.1
    path: /Users/ivanvilches/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.2
    wanted: 0.72.2
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false

## Steps To Reproduce

when i run the project using npx react-native start then i open the emulator for ios and android, in both i can only listen the audio but the video player is on **Black**
...

## Expected behaviour
Watch the video

[...]

Issue #3020, opened 2023-01-30
iOS Black Screen when pull top menu from the phone or if notification or modal pops


Aggroid:
# Bug

## Platform
React-Native iOS
Which player are you experiencing the problem on:
* iOS

## Environment info

React native info output:

info Fetching system and libraries information...
(node:10018) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
System:
    OS: macOS 12.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 145.61 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 19.2.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29, 30, 31, 33
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.0, 30.0.2, 30.0.3, 31.0.0, 33.0.0
      System Images: android-26 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7678000
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
    Java: 13.0.2 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: 7.0.4 => 7.0.4 
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.2 => 0.68.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

Library version: 5.2.1

## Steps To Reproduce

1. Open Video
2. Swipe Menu from the top of the iphone
3.  Blackscreen and cannot get out of the video i need to force quit the application

[...]

Issue #3012, opened 2023-01-27
While the poster is running on android, it comes black on ios.


ahmeturganci:
# Bug
When I want to run a .mp3 source in my iOS project (you can simply think of it as a podcast), I want a constant picture in the background and the source playing in the background all the time. For this

`audioOnly = {true} `
`playInBackground = {true} ` 
`playWhenInactive = {true}  `
`podcast = {a image url with url comes from api}  `

on the bottom of the summary my problem is
While I can use a picture directly from API for podcast on android, it didn't work that way for ios, I just see a black screen.
## Platform
* iOS

## Environment info

React native info output:

```bash

```

Library version: 
react-native-video 5.2.0
react-native 0.63.4 

## Reproducible sample code
```
<VideoPlayer
        isFullScreen={fullScreen}
        statusBarHeight={0}
        hardWareButtonHeight={Platform.OS === 'android' ? Dimensions.get('screen').height - Dimensions.get('window').height : 0}
        progressUpdateInterval={1000}
        muted={videoRef?.state?.muted}
        resizeMode={fullScreen ? 'contain' : 'stretch'}
        style={fullScreen ? styles.backgroundVideoFullScreen : styles.backgroundVideo}
        ref={(ref) => {
          videoRef = ref;
        }}
        onPause={() => {
          clearInterval(interval);
        }}
        onPlay={() => {
          interval = setInterval(() => {
            timer += 1;
          }, 1000);
        }}
        onEnd={() => //some logical stuff}
        onLoad={(event) => {
         // some logical stuff
        }}
        onProgress={(event) => {
          // some logical stuff
        }}
        onExitFullscreen={() => {
           // some logical stuff
        }}
        onEnterFullscreen={() => {
          // some logical stuff
        }}
        source={{
          uri: podCasturl,//here is the,
          headers: {
            access_token: accessToken, // token for reach our api
          },
        }}
        poster={imagePath} // i need to set token in this prop cause android work perfetcly i dont need add any another config but ios i just see black screen
        posterResizeMode="stretch"
      />
```

ahmeturganci:
Basically i just convert image to base64 then fixed

freeboub:
should be retested with liked pr merged

paul-rinaldi:
Can this be closed @freeboub ? It is old and I think it was solved.

PR #3248, opened 2023-09-19
fix: Player shows black screen on re-enter (iOS)

yooasd11:
I've identified an issue on iOS where, when not using PIP (Picture-in-Picture) playback, transitioning the app to the background while a video is playing and then returning to the foreground results in only the audio playing with a black screen.

I believe this is due to the mismatched conditions: setting the player to nil when entering the background with the !_playInBackground condition, and not having a corresponding condition when returning to the foreground.

I'd appreciate it if you could take a look.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -197,7 +197,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     @objc func applicationWillEnterForeground(notification:NSNotification!) {
         self.applyModifiers()
-        if _playInBackground {
+        if !_playInBackground {
             _playerLayer?.player = _player
             _playerViewController?.player = _player
         }


freeboub:
can you confirm this is linked to following tickets please : 
https://github.com/react-native-video/react-native-video/issues/3172
https://github.com/react-native-video/react-native-video/issues/3020
https://github.com/react-native-video/react-native-video/issues/3012
issue seems really similar.

Thank you for the patch !

yooasd11:
[#3172](https://github.com/react-native-video/react-native-video/issues/3172) : This doesn't look like the one I'm facing (Mine is only on iOS, and reproduces when I go background and then come foreground back)
[#3020](https://github.com/react-native-video/react-native-video/issues/3020) : Yes. It seems so.
[#3012](https://github.com/react-native-video/react-native-video/issues/3012) : I'm not sure but it seems related.

And no problem!

Issue #3227, opened 2023-08-31
[iOS] Fairplay with different key per asset not playing


facugu1998:
# Bug

## Platform
Which player are you experiencing the problem on:
* iOS

## Description
When a content has mutliple asset and each asset has different key, for example, skd://{key}/AUDIO for audio playlist, skd://{key}/HD and skd://{key}/SD for video playlist, the content is not being played in most cases and it throws .

```
 {"error": {"code": -19152, "domain": "CoreMediaErrorDomain", "localizedDescription": "The operation couldn’t be completed. (CoreMediaErrorDomain error -19152 - No se pudo completar la operación. (Error de CoreMediaErrorDomain -19152))", "localizedFailureReason": "", "localizedRecoverySuggestion": ""}, "target": 143}
``` 

This issue is related to #2881 but in version 6.0.

As prerequisite you should change:
`let contentId = _drm.contentId ?? loadingRequest.request.url?.host`
to
`let contentId = _drm.contentId ?? loadingRequest.request.url?.absoluteString`

in RCTResourceLoaderDelegate.swift in order to send the complete key string to the license server, otherwise the licenseServer doesn't recognize the sent key.

## Environment info

React native info output:

System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M2
    Memory: 88.39 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.5.1 - /opt/homebrew/bin/node
    Yarn: Not Found
    npm: 9.8.0 - /opt/homebrew/bin/npm
    Watchman: 2023.08.28.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.12.1 - /Users/ingenieria/.rvm/gems/ruby-2.7.5/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 29, 30, 31, 33, 33
      Build Tools: 29.0.2, 30.0.3, 31.0.0, 33.0.0, 33.0.1
      System Images: android-31 | Google APIs ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9477386
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
[...]

Issue #2881, opened 2022-10-03
[iOS] Fairplay with different key per asset [AVAssetResourceLoadingRequest finishLoading] was sent to an instance of AVAssetResourceLoadingRequest that was already finished. Ignoring.


daviCalavera:
# Bug

## Platform
iOS (Tested on devices: iPhone 12 with iOS 16 & iPhone 8 with iOS 15)

## Environment info

React native info output:

```bash
System:
    OS: macOS 12.6
    CPU: (4) x64 Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
    Memory: 1.31 GB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.17.1/bin/npm
    Watchman: 2022.09.26.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 16.0, macOS 12.3, tvOS 16.0, watchOS 9.0
    Android SDK:
      API Levels: 28, 29, 30, 31
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.0, 30.0.2, 30.0.3, 31.0.0, 31.0.0, 31.0.0, 33.0.0, 33.0.0
      System Images: android-28 | Android TV Intel x86 Atom, android-29 | Google Play Intel x86 Atom, android-Q | Android TV Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.3 AI-213.7172.25.2113.9014738
    Xcode: 14.0.1/14A400 - /usr/bin/xcodebuild
  Languages:
    Java: javac 19 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: 0.68.2 => 0.68.2
```

Library version: 5.2.1

## Steps To Reproduce

1. Open the app
2. select FairPlay content with single-drm key
3.  custom getLicense fetch the license and video starts to play.
4. select Fairplay content with multi-drm key (Key for AUDIO stream and a different key for VIDEO).
[...]

PR #3261, opened 2023-09-26
Fix/fairplay different key per asset

facugu1998:
# Summary
As described in #3227, when a content has mutliple asset and each asset has different key, for example, skd://{key}/AUDIO for audio playlist, skd://{key}/HD and skd://{key}/SD for video playlist, the content is not being played.

# Problem found

In `ios/Video/Features/RCTResourceLoaderDelegate.swift`, for every `contentId` the function `handleDrm()` is called. For every call, the `_loadingRequest` object was set to the current call, so the reference to the previous `_loadingRequest` is lost. Also, when the `license` is retrieved from the `getLicense` function, the `setLicenseResult` is called, and sets the current `_loadingRequest` to finished, event though it was already finished.

# Solution description

We replaced the `_loadingRequest` object to a dictionary of `AVAssetResourceLoadingRequest` with key the `contentId` of the request. So, when the license is resolved, it looks for the request in the dictionary and finishes it with the corresponding value. Then, it removes it from the dictionary.

In order to get that working, we had to add the parameter `contentId` to the native functions `setLicenseResult` and `setLicenseResultError`. Also, we had to pass the `contentId` as `loadingRequest.request.url?.absoluteString,` in `_onGetLicense` inside `handleDRM`, so, the getLicense function now receives the full request url and not only the request's host.

# Video Sample

You can test the problem and solution with the following content:

URL: https://cdnw1-stg.zetatv.com.uy/testzapp/bigbuckbunny/playlist.m3u8
licenseServer: https://pm-stg.zetatv.com.uy:8443/api/fairplay/get_license/fbdd541b-0036-4b4c-afef-351b7627ba81/
certificateUrl: https://pm-stg.zetatv.com.uy:8443/fairplay/cert.crt

## Requirements for testing

The onGetLicense function defined on the video component should be the following:

```[javascript]
import { Base64 } from "js-base64";
getLicense: async (spc, skd) => {
          return fetch(licenseServer, {
            method: "POST",
            headers: {
              "Content-Type": "application/octet-stream",
              keyId: skd,
            },
            body: Base64.toUint8Array(spc),
          })
            .then((response) => response.blob())
            .then(async (response) => {
              const base_64 = await blobToBase64(response);
              return base_64.split(",")[1];
            })
            .catch((error) => {
              console.error("Error", error);
            });
        },
      },
```

### DIFF
--- a/Video.js
+++ b/Video.js
@@ -267,15 +267,15 @@ export default class Video extends Component {
         const getLicensePromise = Promise.resolve(getLicenseOverride); // Handles both scenarios, getLicenseOverride being a promise and not.
         getLicensePromise.then((result => {
           if (result !== undefined) {
-            NativeModules.VideoManager.setLicenseResult(result,  data.contentId, findNodeHandle(this._root));
+            NativeModules.VideoManager.setLicenseResult(result,  data.licenseUrl, findNodeHandle(this._root));
           } else {
-            NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('Empty license result',  data.contentId, findNodeHandle(this._root));
+            NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('Empty license result',  data.licenseUrl, findNodeHandle(this._root));
           }
         })).catch((error) => {
-          NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError(error, data.contentId, findNodeHandle(this._root));
+          NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError(error, data.licenseUrl, findNodeHandle(this._root));
         });
       } else {
-        NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('No spc received', data.contentId, findNodeHandle(this._root));
+        NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('No spc received', data.licenseUrl, findNodeHandle(this._root));
       }
     }
   }
--- a/Video.js
+++ b/Video.js
@@ -267,15 +267,15 @@ export default class Video extends Component {
         const getLicensePromise = Promise.resolve(getLicenseOverride); // Handles both scenarios, getLicenseOverride being a promise and not.
         getLicensePromise.then((result => {
           if (result !== undefined) {
-            NativeModules.VideoManager.setLicenseResult(result, findNodeHandle(this._root));
+            NativeModules.VideoManager.setLicenseResult(result,  data.contentId, findNodeHandle(this._root));
           } else {
-            NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('Empty license result', findNodeHandle(this._root));
+            NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('Empty license result',  data.contentId, findNodeHandle(this._root));
           }
         })).catch((error) => {
-          NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError(error, findNodeHandle(this._root));
+          NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError(error, data.contentId, findNodeHandle(this._root));
         });
       } else {
-        NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('No spc received', findNodeHandle(this._root));
+        NativeModules.VideoManager.setLicenseError && NativeModules.VideoManager.setLicenseError('No spc received', data.contentId, findNodeHandle(this._root));
       }
     }
   }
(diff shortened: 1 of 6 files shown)


ios/Video/Features/RCTResourceLoaderDelegate.swift
```diff
@@ -135,7 +158,7 @@ class RCTResourceLoaderDelegate: NSObject, AVAssetResourceLoaderDelegate, URLSes
             ) .then{ spcData -> Void in
                 self._requestingCertificate = true
                 self._onGetLicense?(["licenseUrl": self._drm?.licenseServer ?? "",
-                                     "contentId": contentId,
```
freeboub, line 138: As this is a breaking change, it would be better to keep contentId as is and add a new property for license url 
(loadingRequest.request.url?.absoluteString).
It will make the update easier for client applications

Video.js
```diff
@@ -267,15 +267,15 @@ export default class Video extends Component {
         const getLicensePromise = Promise.resolve(getLicenseOverride); // Handles both scenarios, getLicenseOverride being a promise and not.
         getLicensePromise.then((result => {
           if (result !== undefined) {
-            NativeModules.VideoManager.setLicenseResult(result, findNodeHandle(this._root));
+            NativeModules.VideoManager.setLicenseResult(result,  data.contentId, findNodeHandle(this._root));
```
freeboub: Do this change impact client app ?

freeboub (changes requested):

freeboub:
Thank you for this great enhancement !
I just have doubt on api changes done (as you see in my comments).
As you change api, can you also update documentation to explicite changes you made.

I think you update https://github.com/react-native-video/react-native-video/blob/master/docs/DRM.md#getlicense function
and I am not sure, but the setLicence result changes also need application update ? will / can it be backward compatible with previous version ?

Thank you

ios/Video/Features/RCTResourceLoaderDelegate.swift
facugu1998: We can change the licenseUrl parameter as you defined it in  https://github.com/react-native-video/react-native-video/issues/2881#issuecomment-1416190805 and keep contentId as is.

Video.js
facugu1998: It should not affect the client app because `setLicenseResult` and `setLicenseError` are not called by the client.

facugu1998:
Hello @freeboub, thanks for your comments. We decided to take in account your comment https://github.com/react-native-video/react-native-video/issues/2881#issuecomment-1416190805 and kept `contentId` as is, and use the licenseServer instead, to forward the `skd` url to the client. 

The `getLicense` function passed on the `drm` object receives the same parameters as before, the difference is that if the client does not specify the licenseServer, the `getLicense` function receives the `skd://{key}` url.

freeboub:
@facugu1998 Thank you for the changes, can you please complete DRM.md file, at least to add the new prop: licenseUrl
Thank you

facugu1998:
@freeboub Actually, the parameters were available before the PR. The docs were not updated, how can we describe the function? We thought something like this:

### `getLicense`

`licenseServer` and `headers` will be ignored. You will obtain as argument the `SPC` (as ASCII string, you will probably need to convert it to base 64) obtained from your `contentId` + the provided certificate via `[loadingRequest streamingContentKeyRequestDataForApp:certificateData contentIdentifier:contentIdData options:nil error:&spcError];`. 

Also, you will receive the `contentId` and a `licenseUrl` URL defined as `loadingRequest.request.URL.absoluteString ` or as the `licenseServer` prop if it's passed.
  
  You should return on this method a `CKC` in Base64, either by just returning it or returning a `Promise` that resolves with the `CKC`.

With this prop you can override the license acquisition flow, as an example:

```js
getLicense: (spcString, contentId, licenseUrl) => {
  const base64spc = Base64.encode(spcString);
  const formData = new FormData();
  formData.append('spc', base64spc);
  return fetch(`https://license.pallycon.com/ri/licenseManager.do`, {
      method: 'POST',
      headers: {
          'pallycon-customdata-v2': 'd2VpcmRiYXNlNjRzdHJpbmcgOlAgRGFuaWVsIE1hcmnxbyB3YXMgaGVyZQ==',
          'Content-Type': 'application/x-www-form-urlencoded',
      },
      body: formData
  }).then(response => response.text()).then((response) => {
      return response;
  }).catch((error) => {
      console.error('Error', error);
  });
}
```

freeboub:
@facugu1998 Yes that's Ok for me, thank you. (I agree DRM doc is not really clear in general ;) )

facugu1998:
Hello @freeboub, we updated the DRM documentation. Should we change something else?

freeboub (approved):

freeboub:
@facugu1998 can you please chack following PR please ? https://github.com/react-native-video/react-native-video/pull/3578 I changed the api as it is causing some regressions on existing implementation, do you agree with that ?

facugu1998:
@freeboub hello, I think it's perfect. But in order to make it work I had to do the changes I detailed here https://github.com/react-native-video/react-native-video/issues/3417#issuecomment-1988591585.

PR #3277, opened 2023-10-05
feat(android): add prop to control debug log level

freeboub:
Add custom log level prop for android.
It allows to increase log verbosity on demand.
This is a preliminary PR for futur work.
A toolbox to handle safe parsing of component props has also be created (more job to do to apply it systematically).
Documentation updated.


android/src/main/java/com/brentvatne/common/toolbox/ReactBridgeUtils.kt
```diff
@@ -0,0 +1,142 @@
...
+    fun toStringMap(readableMap: ReadableMap?): Map<String, String?>? {
+        if (readableMap == null) return null
+        val iterator = readableMap.keySetIterator()
+        if (!iterator.hasNextKey()) return null
+        val result: MutableMap<String, String?> = HashMap()
+        while (iterator.hasNextKey()) {
+            val key = iterator.nextKey()
+            result[key] = readableMap.getString(key)
+        }
+        return result
```
KrzysztofMoch, line 84: Is this same as this ?
https://github.com/react-native-video/react-native-video/blob/abdf3988a9173924ba8007003c007b633a2c44f6/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java#L449C5-L464C6

If yes, do we need the old one ?

freeboub: Thank you, I will remove the old one !

freeboub: Done, Thank you !

KrzysztofMoch (approved):
LGTM 👍

PR #3279, opened 2023-10-06
chore: lint project

KrzysztofMoch:
## Summary
Lint project with eslint & prettier (fix CI fails)

## Test plan
_Nothing to test_


src/VideoNativeComponent.ts
```diff
@@ -134,86 +139,93 @@ export type OnSeekData = Readonly<{
...
   error: string;
-}>
+}>;
 
 export type OnAudioFocusChangedData = Readonly<{
   hasFocus: boolean;
-}>
+}>;
 
-export type NativeVideoResizeMode = 'ScaleNone' | 'ScaleToFill' | 'ScaleAspectFit' | 'ScaleAspectFill';
```
freeboub, line 216: Why ? It was better before!

freeboub (changes requested):

freeboub:
OK, let's merge this change asap, @KrzysztofMoch please check my comment in another PR

src/VideoNativeComponent.ts
KrzysztofMoch: It because in native it is not like that - in android it is int and on iOS it is something else. I guess i will refactor this to dont export constants, but create string type and handle it to return responding value on native side

PR #3288, opened 2023-10-11
fix: update types

### LINKED ISSUE #3287: SelectedTracks
# Bug

The latest changes to the player are not compatible with the previous API.
There is no method in Video.tsx for selecting video tracks.
In methods for selecting audio and text tracks, the fields of an selectedTrack now depend on its type (value/index). Changing audo track causes the sound to drop out.

There are no changes at Documentation related to selectedTracks too.

KrzysztofMoch:
## Summary
Fix types

fixes #3287

## Test Plan
- [x] prop `selectedTextTrack` should work
- [x] prop `selectedAudioTrack` should work

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -60,6 +60,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
       fullscreen,
       drm,
       textTracks,
+      selectedVideoTrack,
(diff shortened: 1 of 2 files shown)


KrzysztofMoch:
Test are failing due to CI node version - created PR #3289  to fix it

Issue #2966, opened 2022-12-21
player-accessed-on-wrong-thread


sandeep14:
# Bug

below is error reported on playstore console
Exception java.lang.IllegalStateException: Player is accessed on the wrong thread. See https://exoplayer.dev/issues/player-accessed-on-wrong-thread
  at com.google.android.exoplayer2.SimpleExoPlayer.verifyApplicationThread (SimpleExoPlayer.java)
  at com.google.android.exoplayer2.SimpleExoPlayer.getPlayWhenReady (SimpleExoPlayer.java)
  at com.brentvatne.exoplayer.ReactExoplayerView.pausePlayback (ReactExoplayerView.java)
  at com.brentvatne.exoplayer.ReactExoplayerView.onAudioFocusChange (ReactExoplayerView.java)
  at android.media.AudioManager$ServiceEventHandlerDelegate$1.handleMessage (AudioManager.java:3017)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:246)
  at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run (MessageQueueThreadImpl.java)
  at java.lang.Thread.run (Thread.java:923)

## Platform
android 

Which player are you experiencing the problem on:

* Android

## Environment info
react-version 0.70.6

Library version: 5.2.1

freeboub:
can it be reproduced on master branch ?

kyle-ssg:
I can confirm this is replicable 6.0.0-alpha.6 at least, though in rare circumstances. I have only 1 Sentry report of this 

Redmi Note 10 Pro
Android 11

freeboub:
@kyle-ssg can you provide the callstack please?

kyle-ssg:
![image](https://github.com/react-native-video/react-native-video/assets/8608314/5cdd22f2-457a-439d-9086-824703fbb4fe)

 Hope this helps, thanks so much for all the effort into this library.

freeboub:
This issue will be fixed in 6.0.0-alpha.9

Let me know if you have other crash reposted by crashlytics, it is always interesting !

PadovaY:
hey, this crash happens to me in release mode, mostly on android 13/14 devices. version: 6.2.0.
any suggestion?

stacktrace:
```
Fatal Exception: java.lang.IllegalStateException: Player is accessed on the wrong thread.
Current thread: 'mqt_native_modules'
[...]

PR #3292, opened 2023-10-12
fix: ensure audio volume is changed in UI thread

freeboub:
Fix: ensure audio volum is changed in UIThread when audio duck.
I was not able to test it yet, but to test it you need to generate some notification during playback
shall fix crash : https://github.com/react-native-video/react-native-video/issues/2966


KrzysztofMoch (approved):
Changes looks good - I tested it on physical device (android 13) and it's working 🔥

PR #3299, opened 2023-10-17
remove kotlin-android-extensions

wjaykim:
Removed `kotlin-android-extensions` gradle plugin since it causes build error with following message:
```
The 'kotlin-android-extensions' Gradle plugin is no longer supported. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,6 +1,5 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
-apply plugin: 'kotlin-android-extensions'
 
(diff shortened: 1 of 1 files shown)


freeboub:
This was integrated for a first PR to prepare new architecture integration.
Anyway tested locally sucessfully.
It will be available in 6.0.0-apha.9

Issue #3082, opened 2023-04-01
DRM content fails to play sometimes on Android


efkann:
# Bug

Android Exoplayer fails to play DRM protected content sometimes. This bug is hard to explain, but I'll try my best.

This rarely happens on a bare react native project. But when it comes to expo, you use a expo development client and the content plays on the first try but never works again until you stop and start the dev client.

## Platform
Android

Which player are you experiencing the problem on:

Android Exoplayer

Library version: 6.0.0-alpha.5

## Steps To Reproduce

Think about two screens. 

-HomeScreen
-PlayerScreen

Try to navigate to Player screen from Home screen and go back to Home screen and then again navigate to Player screen. Repeat this 20-25 times and it will fail to play the protected content in one of those tries at least, in bare workflow. But in Expo, it plays the content at first navigation to Player screen but when you go back to Home screen and navigate back to the Player screen, it fails to play it. And it will never be able to play it until you go and restart the dev client.

Actually, the content looks like its playing and you can see the progress, play or pause but the screen is all black. I believe, player thinks its clear content but actually its not so it starts playing it but you cant see the content which also explains the fact that there's not even a single error printed out.

See the `else if` block I'm talking about below.

## Expected behaviour

It should play the protected content 

## Reproducible sample code

Since this issue mostly happens on Expo and rarely happens on bare workflow. I'm not sure what to add here. But I've been debugging this for a while and I might have an idea why this happens.

For testing DRM, you can try these :

sourceUrl: https://storage.googleapis.com/shaka-demo-assets/angel-one-widevine/dash.mpd
licenseUrl: https://cwip-shaka-proxy.appspot.com/no_auth

https://github.com/react-native-video/react-native-video/blob/05864dc64a588f9ffeeded45855eb390fd15b1d8/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L553-L593

At the bottom of the block of code above I don't understand what this `else if` bit is doing actually.

```java
else if (srcUri != null) { 
     initializePlayerSource(self, null); 
 } 
```

Why we're passing null for `drmSessionManager` when `playerNeedsSource` is false ?

I also realized that, `playerNeedsSource` stays false for some reason and I believe it is the reason why the player tries to play DRM content but fails to do so. 

It's very interesting that for my Expo app, I prebuilt the app for Android and removed the`else if` block above and everything just worked perfectly.

## Note

[...]

Issue #3022, opened 2023-02-01
The video stays black and selecting an audio track or video track causes an error.


Arekjaar4:
# Bug

## Platform

Which player are you experiencing the problem on:
Android

## Environment info

React native info output:

```bash
 System:
    OS: Windows 10 10.0.22000
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 3.20 GB / 15.89 GB
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 7.24.1 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      Android NDK: 21.3.6528147
    Windows SDK: Not Found
  IDEs:
    Android Studio: Version  4.2.0.0 AI-202.7660.26.42.7351085
    Visual Studio: Not Found
  Languages:
    Java: 1.8.0_261
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.0.0 => 18.0.0
    react-native: 0.69.6 => 0.69.6
    react-native-windows: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
info React Native v0.71.1 is now available (your project is running on v0.69.6)
```

Library version: "react-native-video": "^6.0.0-alpha.3"

## Steps To Reproduce

These three videos cause me problems when playing them with react-native-video. The links I attach only last 12 hours due to their content.

1. http://umetop.ottc.pro/series/CPR14KRJTP/74W5QAQEZH/462683.mkv
2. http://umetop.ottc.pro/series/CPR14KRJTP/74W5QAQEZH/258331.mkv
3. http://umetop.ottc.pro/movie/CPR14KRJTP/74W5QAQEZH/488270.mkv

The one and two are visible but not audible, and selecting an audio track causes the following error:

{
  "error": Object {
    "errorCode": "24001",
    "errorException": "com.google.android.exoplayer2.ExoPlaybackException: MediaCodecVideoRenderer error, index=0, format=Format(1, null, null, video/x-unknown, null, -1, und, [720, 304, -1.0], [-1, -1]), format_supported=NO_UNSUPPORTED_TYPE",
[...]

PR #3300, opened 2023-10-17
fix : onAttachedToWindow can call multiple times.

lunasky-hy:
Related issue: #3022

Summary : 

When onAttachedToWindow in ReactExoplayerView.java is called multiple times, the video component blacks out and the error occurs

Fix : 

in ReactExolayerView.java
When onAttachedToWindow called multiple times, player initialize function call only one time.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -317,7 +317,9 @@ public class ReactExoplayerView extends FrameLayout implements
     @Override
     protected void onAttachedToWindow() {
         super.onAttachedToWindow();
-        initializePlayer();
+        if(player == null) {
+            initializePlayer();
+        }
     }
 
     @Override


freeboub:
Thank you for the PR.
I am afraid this is not fixing all issues... 
The real root cause of the issue is the asynchronous call inside initializePlayer();
But it not easy to fix.
I am working on this topic, but I have some preliminary work to integrate before the real fix (which will be huge).
I propose you use use patch-package with this fix as a temporary solution.

if some other people push to integrate it, I can merge it as a temporary fix.

freeboub:
Can it also fix : https://github.com/react-native-video/react-native-video/issues/3082 ?

KonstantinosKoulaxizis:
@lunasky-hy I encountered an issue with my app when using a stream with DRM. Everything was functioning correctly on Android mobile, Android TV, iOS, and tvOS. However, in the case of FireTV, I was experiencing a black screen after 0.5 seconds. Your PR helped me resolve this issue. Thanks ❤️

VMBindraban:
Please merge this, seems like a valid temp solution.

freeboub:
@VMBindraban @KonstantinosKoulaxizis @lunasky-hy 
Hello, can you please have a look to this PR : https://github.com/react-native-video/react-native-video/pull/3668
And check if it also fixes you r issue please ?

freeboub:
linked PR closed, let's close this one. Fix will be available in 6.0.0-beta.9

PR #3304, opened 2023-10-18
chore: rework typescript integration

freeboub:
- rework typescript integration
- remove duplicated types
- rework sample to use types
- update package.json
- create some enum


freeboub:
@KrzysztofMoch finally, here it is !

There are 2 points which cn be improved:

1/ I just wondering if there is a possibility to merge: 
export interface VideoNativeProps extends ViewProps {
and 
export interface ReactVideoProps extends ReactVideoEvents {

They both define all the callback available. It will be painful to maintain both interface ...

2/ index.ts and index.d.ts have the same content, not sure if it possible to have it only one time.

KrzysztofMoch:
First of all, I would like to draw attention to the fact that the removal of duplicate types in `VideoNativeComponent` is not the best idea in my opinion. It may look bad now, but in the future, when we start to adopt a new architecture, we will have to type the "Native Component" using [codegen types](https://reactnative.dev/docs/new-architecture-library-intro?fabric-component-backward-compatibility=TypeScript#codegen-helper-types) _(types like Int32, Double etc.)_, which will force us to re-do those interfaces.

package.json
```diff
@@ -5,6 +5,7 @@
     "main": "lib/index",
     "source": "src/index",
     "react-native": "src/index",
+    "types": "./src/index.d.ts",
```
KrzysztofMoch: We will ship package with "built files" (lib folder created by `yarn build`). In lib folder there is already `index.d.ts`, so we can change it to `lib/index.d.ts`
```suggestion
    "types": "lib/index.d.ts",
```

src/index.ts
```diff
@@ -1,11 +1,14 @@
 import Video from './Video';
 
-export {default as FilterType} from './lib/FilterType';
-export {default as VideoResizeMode} from './lib/VideoResizeMode';
-export {default as TextTrackType} from './lib/TextTrackType';
-export {default as DRMType} from './lib/DRMType';
+export {default as ResizeMode} from 'react-native-video/src/types/ResizeMode';
```
KrzysztofMoch: Should be relative path
```suggestion
export {default as ResizeMode} from ./types/ResizeMode';
```

KrzysztofMoch (changes requested):
There are few small corrections. 
Regarding the removal of interfaces, let's talk - we can remove them but we will still have to restore them with the new architecture adoption

src/VideoNativeComponent.ts
```diff
@@ -3,13 +3,45 @@ import type {
...
+  OnReceiveAdEventData,
+  OnSeekData,
+  OnTextTracksData,
+  OnTimedMetadataData,
+  OnVideoAspectRatioData,
+  OnVideoErrorData,
+  OnVideoTracksData,
+} from './types/events';
+import type ResizeMode from './types/ResizeMode';
+import type FilterType from 'react-native-video/src/types/FilterType';
```
KrzysztofMoch: should be relative path
```suggestion
import type FilterType from './types/FilterType';
```

src/VideoNativeComponent.ts
```diff
@@ -3,13 +3,45 @@ import type {
...
+  SubtitleStyle,
+  BufferConfig,
+  SelectedAudioTrack,
+} from './types/video';
 import {NativeModules, requireNativeComponent} from 'react-native';
+import type {
+  OnAudioFocusChangedData,
+  OnAudioTracksData,
+  OnBandwidthUpdateData,
+  OnVideoBufferData,
```
KrzysztofMoch: There is no `OnVideoBufferData` type 🤔 did you mean `OnBufferData` ?

KrzysztofMoch (changes requested):

freeboub:
> First of all, I would like to draw attention to the fact that the removal of duplicate types in `VideoNativeComponent` is not the best idea in my opinion. It may look bad now, but in the future, when we start to adopt a new architecture, we will have to type the "Native Component" using [codegen types](https://reactnative.dev/docs/new-architecture-library-intro?fabric-component-backward-compatibility=TypeScript#codegen-helper-types) _(types like Int32, Double etc.)_, which will force us to re-do those interfaces.

Ok understand, that make sense, Thank you !

freeboub:
> There are few small corrections. Regarding the removal of interfaces, let's talk - we can remove them but we will still have to restore them with the new architecture adoption

Ok, no problem, let's keep them so !

src/index.ts
freeboub: I found 2 more absolute path, I will remove it also

src/VideoNativeComponent.ts
freeboub: good catch, I will fix it

KrzysztofMoch:
@freeboub please restore the deleted types from `VideoNativeComponent`, and we can merge

freeboub:
> @freeboub please restore the deleted types from `VideoNativeComponent`, and we can merge

Done !

KrzysztofMoch:
Nice! last thing - we can remove `src/index.d.ts`

freeboub:
@KrzysztofMoch Ok, I removed it, I also fix the issue linked to onBandWidthUpdate reported by @gizomo 
Can be reviewed.
Notice that I will open another PR for the release build.
I think we just need to add a prepare section in script from package.json to allow build before packaging

KrzysztofMoch:
@freeboub Just found that we don't export `save` function in ref and type for `save` in VideoManagerType is `Promise<void>` even that `save` function should return some data in promise. Cloud you look at it ?

freeboub:
> @freeboub Just found that we don't export `save` function in ref and type for `save` in VideoManagerType is `Promise<void>` even that `save` function should return some data in promise. Cloud you look at it ?

Sure, this is done.
Please notice that I also change pause and resume function to really return a promise as detailed in doc

KrzysztofMoch (approved):
LGTM 👍

PR #3306, opened 2023-10-23
fix: remove false calls at `onPlaybackRateChange`

KrzysztofMoch:
## Summary
Don't emit event `onPlaybackRateChange` if value is same as in previous call - like on android

## Test plan
- [x] run example & check if callback is working correctly

PR #3307, opened 2023-10-23
feat(ios): implement onPlaybackStateChanged callback

KrzysztofMoch:
## Summary
Implement `onPlaybackStateChanged` on iOS

## Note
There are multiple calls of this event - it will be fixed by https://github.com/react-native-video/react-native-video/pull/3306

## Test plan
- [x] run example & check if callback is working correctly


eedeebee:
I don't see evidence of this event being sent on iOS in beta6 fwiw.

Issue #3312, opened 2023-10-24
[Bug]: VideoTracks selection


gizomo:
### What happened?

Multiple defects with video track selection on Android (ExoPlayer):

**1) `VideoTrack.id` is permanently `-1` and shadows `trackId`**

Class `VideoTrack` has both an `id` and a `trackId` field, and `id` is permanently set to `-1`. The function `videoTracksToArray` in `VideoEventEmitter` uses `id` from `VideoTrack` as the `trackId`. As a result, all video tracks on the JS side only have a `trackId` field and it is always set to `-1`.

**2) `exoplayerVideoTrackToGenericVideoTrack` assigns a non-unique trackId**

The function `exoplayerVideoTrackToGenericVideoTrack` takes `trackIndex`, and if `format.id` is null it sets `trackIndex` as the `trackId` field of `VideoTrack`. But `trackIndex` is the index of the video format within the `TrackGroup`. This is wrong because `trackId` will not be unique in that case.

**3) `format.id` must be stringified**

`String.valueOf(format.id)` should be used because `VideoTrack.trackId` is a string, and JS receives `undefined` if a number is passed instead.

**4) `trackId` is effectively useless and `index` selection is broken**

`VideoTrack.trackId` is not used anywhere. Selecting a video track by type `index` actually selects a `TrackGroup`, not a track within that group. So it is impossible to select a video track by `index` for a single video `TrackGroup`. In the case of several video `TrackGroup`s, selection by `index` is not obvious at all, as it selects the group rather than a specific track.

It may be preferable to assign an absolute index for video tracks, or to use a `trackId` composed from the index of the `TrackGroup` and the index of the video format within that group.

### Steps to reproduce

1. Play a source that exposes multiple video tracks (or multiple video track groups) on Android.
2. Read the `videoTracks` array reported to JS and observe that every `trackId` is `-1` (or `undefined`).
3. Attempt to select a video track using `selectedVideoTrack` with type `index`.
4. Observe that the selection targets the `TrackGroup` rather than the specific video track, so a specific track cannot be selected.

### Platforms

Android

freeboub:
Looks like you already fixed these issues. A PR would be welcome! Thank you

gizomo:
As for 4. I'm not sure how it should work and be compatible with the AVPlay.

I found android description of track group and it says:

[...]

PR #3324, opened 2023-10-30
Updated track selection for exoplayer

gizomo:
This request was created in connection with the issue  [#3312](https://github.com/react-native-video/react-native-video/issues/3312)

- fixed trackId for video tracks
- fixed video tracks selection by index and resolution
- fixed audio tracks selection by language and index
- removed track selection by title (title is not unique track property)
- audio/text tracks title is based on label now (contains the human-readable text)

Video audio/video tracks selection is possible only supported formats now.
Selecting video by resolution takes into account multiple formats with exact height to support (saving ads)
Selecting video by index is possible only first group (as the most common case)
Selecting a audio/text track by title has been removed, since track titles may not be unique and in the case of several groups of tracks do not allow the correct selection.
Titles of Audio/text tracks are set based on format.label (the human-readable text).

### DIFF
--- a/android/src/main/java/com/brentvatne/common/api/DRMProps.kt
+++ b/android/src/main/java/com/brentvatne/common/api/DRMProps.kt
@@ -1,0 +1,66 @@
+package com.brentvatne.common.api
+
+import com.brentvatne.common.toolbox.ReactBridgeUtils.safeGetArray
+import com.brentvatne.common.toolbox.ReactBridgeUtils.safeGetString
+import com.facebook.react.bridge.ReadableMap
+import java.util.UUID
+
+/**
+ * Class representing DRM props for host.
+ * Only generic code here, no reference to the player.
+ */
+class DRMProps {
+    /**
+     * string version of configured UUID for drm prop
+     */
+    var drmType: String? = null
+
+    /**
+     * Configured UUID for drm prop
+     */
+    var drmUUID: UUID? = null
+
+    /**
+     * DRM license server to be used
+     */
+    var drmLicenseServer: String? = null
+
+    /**
+     * DRM Http Header to access to license server
+     */
+    var drmLicenseHeader: Array<String> = emptyArray<String>()
+    companion object {
+        private const val PROP_DRM_TYPE = "type"
+        private const val PROP_DRM_LICENSE_SERVER = "licenseServer"
+        private const val PROP_DRM_HEADERS = "headers"
+        private const val PROP_DRM_HEADERS_KEY = "key"
+        private const val PROP_DRM_HEADERS_VALUE = "value"
+
+        /** parse the source ReadableMap received from app */
+        @JvmStatic
+        fun parse(src: ReadableMap?): DRMProps? {
+            var drm: DRMProps? = null
+            if (src != null && src.hasKey(PROP_DRM_TYPE)) {
+                drm = DRMProps()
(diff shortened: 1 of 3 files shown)


KrzysztofMoch (changes requested):
Thanks @gizomo, it looks good 👍
Could you update `trackId` type in docs and [`events.ts`](https://github.com/react-native-video/react-native-video/blob/ad0422d98986aa48e27712951731861ae1dace60/src/types/events.ts#L84C7-L84C23), please?

gizomo:
> Thanks @gizomo, it looks good 👍 Could you update `trackId` type in docs and [`events.ts`](https://github.com/react-native-video/react-native-video/blob/ad0422d98986aa48e27712951731861ae1dace60/src/types/events.ts#L84C7-L84C23), please?

Done

KrzysztofMoch (approved):

stremium:
Does this track selection already work on iOS?

gizomo:
> Does this track selection already work on iOS?

No, it doesn't. I haven't worked on this yet

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Android: ensure audio volume is changed in UI thread [3292](https://github.com/react-native-video/react-native-video/pull/3292)
 - Android: multiple internal refactor and switch to kotlin
 - Android: refactor log management and add an option to increase log verbosity [#3277](https://github.com/react-native-video/react-native-video/pull/3277)
+- Android: updated track selection [#3312](https://github.com/react-native-video/react-native-video/issues/3312)
```
freeboub: shall be moved to Next release ... sorry ...

src/types/events.ts
```diff
@@ -81,7 +81,7 @@ export type OnTextTracksData = Readonly<{
 export type OnVideoTracksData = Readonly<{
   videoTracks: ReadonlyArray<
     Readonly<{
-      trackId: number;
+      trackId: string;
```
freeboub: Is this really necessary ?
It will be a breaking change for applications...

gizomo: I made trackId a string because the 'id' of the Format in Android is a string type. The exoplayer simply assigns the indices of this group as an ids for all formats of the first video/audio group. And despite this they are also strings. Also if there is a second group, then the formats there will receive real ids from the playlist. And real ids may consist of any string characters, not just numbers.

Actually, the trackId has no meaning, because not used anywhere. If someone had previously used the trackId to select video tracks, it looked at least strange. The RNV neither before nor now allows you to select a track by trackId. If the trackId was meant as a video track index, then this is not obvious. However, selecting a video track by index is possible as for other types of tracks.

If there is a real need to assign ids to tracks, then I would rather use a combination of group and format indexes and do this for all track types. Moreover, as far as I understand, iOS uses grouping of tracks too. For example, “0.4” is a track with index 4 from a group with index 0. Using this id, we can accurately determine the track.

YangJonghun:
@freeboub @gizomo 
I think we can close this PR
(This PR was replaced by PR #3778 and merged)

freeboub:
> @freeboub @gizomo 
> I think we can close this PR
> (This PR was replaced by PR #3778 and merged)

Yes more or less, in this patch there is also a change in isTrackSupported I didn't backport... I think there are still some issues with unsupported tracks ... But I don't have such unsupported track in the sample playlist manifest

PR #3327, opened 2023-11-01
[Android] use explicit not-exported flag for AudioBecomingNoisyReceiver

AndyG:
Add `RECEIVER_NOT_EXPORTED` flag support to `AudioBecomingNoisyReceiver` for Android 14 support. See
https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported for details.

Without this fix, apps using this dependency and targeting SDK 34 will crash when we attempt to register this receiver:

```
SecurityException: {package name here}: One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -62,7 +62,6 @@ dependencies {
         exclude group: 'com.android.support'
     }
 
-    // All support libs must use the same version
     implementation "androidx.annotation:annotation:1.7.0"
     implementation "androidx.core:core:1.12.0"
     implementation "androidx.media:media:1.6.0"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -63,9 +63,9 @@ dependencies {
     }
 
     implementation "androidx.annotation:annotation:1.7.0"
-    implementation "androidx.core:core:1.12.0"
+    implementation "androidx.core:core:1.9.0"
     implementation "androidx.media:media:1.6.0"
-    implementation "androidx.activity:activity:1.8.0"
+    implementation "androidx.activity:activity:1.6.0"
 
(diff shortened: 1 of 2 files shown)


KrzysztofMoch (dismissed):
Thanks 🙌

KrzysztofMoch:
@AndyG were you able to compile library without bumping `androidx.core` version? currently we have `1.1.0` but  `RECEIVER_NOT_EXPORTED` seems to be added in `1.9.0`

AndyG:
nope, def want to bump that, will make the necessary changes, test, and ping when ready. thanks for taking a look!

AndyG:
@KrzysztofMoch I bumped that dependency and some related dependencies to keep them roughly in sync. I was able to use this source on a local build of my own app. Please let me know if you have any further concerns -- excited to help here but my react-native build tools experience is limited to be totally candid so a second opinion is extremely valuable.

KrzysztofMoch:
Thanks, I do some tests and we need to change some versions in order to keep kotlin version at `1.7.0`:
- `androidx.core:core:1.12.0` -> `androidx.core:core:1.9.0`
- `androidx.activity:activity:1.8.0` -> `androidx.activity:activity:1.6.0`

And we need to set minimal kotlin version to `1.7.0` (currently `1.6.20`) but I will do it in separate PR.

@freeboub I have checked and this (`1.7.0`) is recommended version in react native library template

AndyG:
okay thanks for the heads up. I have made those two changes in this PR, but not the kotlin version. Let me know how I can best support you to get this merged and released!

freeboub (approved):

Issue #1726, opened 2019-08-27
How to get bitrate and bandwidth of video


pjreact:
How can I get bitrate and bandwidth of video? I need to manage some stat.

CHaNGeTe:
https://github.com/react-native-community/react-native-video#onbandwidthupdate

joemewes:
this only works on Android? any solutions for iOS where we want to track this data for analytics purpose. thanks.

asallem94:
I want to know this answer for iOS too

ericlifs:
Any updates on supporting this on ios? I noticed that the prop is commented out here https://github.com/react-native-video/react-native-video/blob/master/ios/Video/RCTVideo.m#L801. Anyone knows another video player library that supports this feature?

Thanks

Fedeorlandau:
Any updates on this? https://github.com/react-native-video/react-native-video/pull/1745 the PRa was created 2 years ago.

freeboub:
onBandwidthUpdate notification added for alpha.10 version

PR #3331, opened 2023-11-02
feat(ios): add onBandwidthUpdate event

KrzysztofMoch:
## Summary
Add onBandwidthUpdate event
_backport of #1745_

## Test plan
- [x] example app should work

### DIFF
--- a/ios/Video/Features/RCTPlayerObserver.swift
+++ b/ios/Video/Features/RCTPlayerObserver.swift
@@ -7,8 +7,7 @@ protocol RCTPlayerObserverHandlerObjc {
     func handleDidFailToFinishPlaying(notification:NSNotification!)
     func handlePlaybackStalled(notification:NSNotification!)
     func handlePlayerItemDidReachEnd(notification:NSNotification!)
-    // unused
-//    func handleAVPlayerAccess(notification:NSNotification!)
+    func handleAVPlayerAccess(notification:NSNotification!)
 }
 
 protocol RCTPlayerObserverHandler: RCTPlayerObserverHandlerObjc {
@@ -24,7 +23,7 @@ protocol RCTPlayerObserverHandler: RCTPlayerObserverHandlerObjc {
 }
 
 class RCTPlayerObserver: NSObject {
-
+    
     weak var _handlers: RCTPlayerObserverHandler?
     
     var player:AVPlayer? {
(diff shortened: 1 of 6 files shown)


ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -24,7 +23,7 @@ protocol RCTPlayerObserverHandler: RCTPlayerObserverHandlerObjc {
 }
 
 class RCTPlayerObserver: NSObject {
-
+
```
freeboub:
```suggestion

```

src/VideoNativeComponent.ts
```diff
@@ -148,12 +148,15 @@ export type OnProgressData = Readonly<{
...
-export type OnBandwidthUpdateData = Readonly<{
-  bitrate: number;
-  width: number;
-  height: number;
-  trackId: number;
-}>;
+export type OnBandwidthUpdateData = Readonly<
+  | {
+      bitrate: number;
+      width: number;
```
freeboub: can it be clearer to define it as width?: number ?

src/types/events.ts
```diff
@@ -127,12 +127,15 @@ export type OnAudioFocusChangedData = Readonly<{
...
-export type OnBandwidthUpdateData = Readonly<{
-  bitrate: number;
-  width: number;
-  height: number;
-  trackId: number;
-}>;
+export type OnBandwidthUpdateData = Readonly<
+  | {
+      bitrate: number;
+      width: number;
```
freeboub, line 133: same here width?: number;

ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -179,10 +178,11 @@ class RCTPlayerObserver: NSObject {
     
     func attachPlayerEventListeners() {
         guard let _handlers = _handlers else {return}
-
+
```
freeboub:
```suggestion

```

ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -118,7 +117,7 @@ class RCTPlayerObserver: NSObject {
         _playerPlaybackLikelyToKeepUpObserver?.invalidate()
         _playerTimedMetadataObserver?.invalidate()
     }
-
+
```
freeboub:
```suggestion

```

src/VideoNativeComponent.ts
KrzysztofMoch: Here must be - forgot that code gen don't support types like this. Will fix this

src/types/events.ts
KrzysztofMoch: But here i would like to leave it like this, it will be easier for user - if width is not null it means that height is also not null. Correct me if I am wrong

src/VideoNativeComponent.ts
KrzysztofMoch: Done ✅

ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -105,7 +103,7 @@ class RCTPlayerObserver: NSObject {
     
     func addPlayerItemObservers() {
         guard let playerItem = playerItem, let _handlers = _handlers else { return }
-
+
```
freeboub:
```suggestion

```

src/types/events.ts
freeboub: ok, no problem, and yes you are right, if wifth is ull height is null also

freeboub:
1 small additional clean needed, sorry :)

freeboub (approved):

freeboub:
merged for alpha.10 ticket: https://github.com/react-native-video/react-native-video/issues/1726

Issue #3333, opened 2023-11-05
Package is heavy, It includes exemple and docs folders


freeboub:
# Bug

Package is too heavy, It includes exemple and docs folders

## Platform
* All

## Environment info

Library version: 6.0.0-alpha.9

## Steps To Reproduce

packaging is currently done with np tool

## Expected behaviour

Lighter package because yarn is already a black hole

freeboub:
merged in alpha.10

PR #3342, opened 2023-11-08
feat: add release-it

freeboub:
Integrate release-it tool for new release management

Next version can be generated with:
npx release-it --preRelease=alpha

should address: https://github.com/react-native-video/react-native-video/issues/3333

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,11 @@
         "test": "echo no test available"
     },
     "files": [
+        ...
+        "!docs",
+        "!examples",
+        ...
+    ]
         "android",
         "ios",
         "windows",
--- a/package.json
+++ b/package.json
@@ -47,11 +47,6 @@
         "test": "echo no test available"
     },
     "files": [
-        ...
-        "!docs",
-        "!examples",
-        ...
-    ]
         "android",
         "ios",
         "windows",
@@ -62,6 +57,8 @@
         "!android/buildOutput_*",
         "!android/local.properties",
         "!ios/build",
-        "!**/*.tsbuildinfo"
+        "!**/*.tsbuildinfo",
+        "!docs",
+        "!examples"
     ]
 }
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     },
     "devDependencies": {
         "@react-native/eslint-config": "^0.72.2",
+        "@release-it/conventional-changelog": "^7.0.2",
         "@types/jest": "^28.1.2",
         "@types/react": "~18.0.0",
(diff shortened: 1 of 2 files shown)


evoactivity:
I've been thinking about adding changesets here, it works a lot better for community projects as you don't have to enforce every PR to follow conventional commits, it makes keeping a changelog very simple as peoples changesets get added automatically. It can be completely automated through github actions so anyone with merge access on the repo will be able to make a release. Releases can goto github and npm.

https://github.com/changesets/changesets

I have a repo using automated changesets here if you would like to see how I have it setup
https://github.com/ektotv/playlist

KrzysztofMoch:
@evoactivity Thanks for your feedback! 
But I think that for us it will be much easier to use [release-it](https://github.com/release-it/release-it) with [changelog plugin](https://github.com/release-it/conventional-changelog). To enforce that every PR follow conventional commits we can use [github action](https://github.com/amannn/action-semantic-pull-request). We (I and @freeboub) have already used release-it so it will be easier for us to configurate and use it

package.json
```diff
@@ -41,6 +43,7 @@
         "prepare": "yarn build",
         "xbasic": "yarn --cwd examples/basic",
         "docs": "yarn --cwd docs build",
+        "release": "release-it",
         "test": "echo no test available"
     },
     "files": [
```
KrzysztofMoch, line 49: we should make sure that next release won't have docs and example 😅
```suggestion
    "files": [
        ...
        "!docs",
        "!examples",
        ...
    ]
```

KrzysztofMoch (changes requested):
we need update `package.json` to exclude `docs` and `examples` and we are ready to go

freeboub:
> we need update `package.json` to exclude `docs` and `examples` and we are ready to go

I added, but I think this is not mandatory, all the !XXX are subfolders of included folders.

freeboub:
> I've been thinking about adding changesets here, it works a lot better for community projects as you don't have to enforce every PR to follow conventional commits, it makes keeping a changelog very simple as peoples changesets get added automatically. It can be completely automated through github actions so anyone with merge access on the repo will be able to make a release. Releases can goto github and npm.
> 
> https://github.com/changesets/changesets
> 
> I have a repo using automated changesets here if you would like to see how I have it setup https://github.com/ektotv/playlist

Thank you for the clue, I agree it would be interesting to have github integration, but I had this release-it integration under the hand, it was very easy.

KrzysztofMoch (approved):
LGTM 👍

Issue #3346, opened 2023-11-14
 Cannot find module 'react-native-video'


alan-nakagiri:
# Bug

Cannot find module 'react-native-video' when running tests and eslint for versions v.6.0.0-alpha9 or alpha10.

It looks like some changes were made to the "main" directory listed in `package.json`. The package.json file points to `lib/index.js` however I do not see a lib directory in the repo when pulling from npm.

Do I need to add something to my config files to resolve this or was the build command not run before publishing?

freeboub:
Hi, someone reports me on slack that the alpha.10 package is not usable. I will try to fix it tonight.
But alpha.9 should be ok.

chayin:
I had this problem with video caching and here is my fix

in `react-native-video.podspec`

```
- ss.source_files = "ios/Video/**/*.{h,m,swift}" "ios/VideoCaching/**/*.{h,m,swift}"
+ s.source_files = "ios/*/**/*.{h,m,swift}"
```

I likewise fixed another video caching issue in `/ios/VideoCaching/RCTVideoCachingHandler.swift`
```
-    func dvAssetLoaderDelegate(loaderDelegate:DVAssetLoaderDelegate!, didLoadData data:NSData!, forURL url:NSURL!) {
+    func dvAssetLoaderDelegate(_ loaderDelegate: DVAssetLoaderDelegate!, didLoad data: Data!, for url: URL!) {
         _videoCache.storeItem(data as Data?, forUri:url.absoluteString, withCallback:{ (success:Bool) in
             DebugLog("Cache data stored successfully 🎉")
         })
      }
```

Plus, `canImport(RCTVideoCache)` doesn't work and i have absolutely no idea.

@freeboub

freeboub:
@alan-nakagiri I Just rebuild alpha.11 with an additional fix to ensure package is properly build.
Test on the sample successfully on ios & android.
Let me know if this is Ok for you

freeboub:
@chayin Thank you very much for the report.
Can we just follow up in this ticket please: https://github.com/react-native-video/react-native-video/issues/3217 
Thank you, and let's close this issue

alan-nakagiri:
@freeboub Looks like the new release works thanks!

Issue #3217, opened 2023-08-26
Call to undeclared function 'RCTLog'; ISO C99 and later do not support implicit function declarations


xts-bit:
# Bug

I was following the docs for iOS installation with video caching, But after i followed it didi pod install everything worked fine but as soon i started my project with npm start and run on an iOS simulator it crashes and i found these error in my Xcode

/Users/user/Desktop/myProject/node_modules/react-native-video/ios/VideoCaching/RCTVideoCache.m:33:7 Call to undeclared function 'RCTLog'; ISO C99 and later do not support implicit function declarations

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info
macOS 

Library version: 6.0.0-alpha.7
Device: iPhone 14 pro 

## Steps To Reproduce
npm start 
i
## Expected behaviour

1. It should run my project

## Reproducible sample code
```
# Resolve react_native_pods.rb with node to allow for hoisting
require Pod::Executable.execute_command('node', ['-p',
  'require.resolve(
    "react-native/scripts/react_native_pods.rb",
    {paths: [process.argv[1]]},
  )', __dir__]).strip

platform :ios, min_ios_version_supported
use_modular_headers!
prepare_react_native_project!
pod 'react-native-video/VideoCaching', :path => '../node_modules/react-native-video/react-native-video.podspec'
$RNVideoUseGoogleIMA=true
# If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.
# because `react-native-flipper` depends on (FlipperKit,...) that will be excluded
#
# To fix this you can also exclude `react-native-flipper` using a `react-native.config.js`
# ```js
# module.exports = {
#   dependencies: {
#     ...(process.env.NO_FLIPPER ? { 'react-native-flipper': { platforms: { ios: null } } } : {}),
# ```
flipper_config = ENV['NO_FLIPPER'] == "1" ? FlipperConfiguration.disabled : FlipperConfiguration.enabled

linkage = ENV['USE_FRAMEWORKS']
if linkage != nil
  Pod::UI.puts "Configuring Pod with #{linkage}ally linked Frameworks".green
  use_frameworks! :linkage => linkage.to_sym
end

target 'myProject' do
  config = use_native_modules!

  # Flags change depending on the env values.
  flags = get_default_flags()

  use_react_native!(
    :path => config[:reactNativePath],
    # Hermes is now enabled by default. Disable by setting this flag to false.
[...]

PR #3348, opened 2023-11-15
Fix(ios): another video caching fix

freeboub:
Add fix propose by @chayin in: https://github.com/react-native-video/react-native-video/issues/3346
Linked to original issue: https://github.com/react-native-video/react-native-video/issues/3217

* fix podspec for video caching
* fix conditionnal include usage
* fix dvAssetLoaderDelegate  prototype issue

I was able to build and run with video caching.
I see VideoCaching log displayed
save() api still not working on my side

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -14,7 +14,7 @@
     },
     "hooks": {
         "before:init": [
-            "rm -Rf lib", 
+            "rm -Rf lib tsconfig.tsbuildinfo",
             "yarn install --frozen-lockfile --non-interactive --production=false", 
             "yarn run lint", 
             "yarn run build"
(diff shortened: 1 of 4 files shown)


.release-it.json
```diff
@@ -14,7 +14,7 @@
     },
     "hooks": {
         "before:init": [
-            "rm -Rf lib", 
+            "rm -Rf lib tsconfig.tsbuildinfo",
```
KrzysztofMoch, line 17: Maybe ignore this file by adding to "files" section in package.json ?

KrzysztofMoch (approved):

.release-it.json
freeboub: This is already merged in fact :) it was the issue with the previous release! Once the typescript is built, it creates this file like a cache to remember what has been built. 
In case we build, remove /lib and build again, the last build doesn't generate output due to this tsc cache file

KrzysztofMoch: Okay ✅

Issue #3345, opened 2023-11-13
OnVideoErrorData Typed Incorrectly in 6.0.0-alpha.9


brianlenz:
# Bug
The type definition of `OnVideoErrorData` doesn't match the shape of the error returned in Android. I haven't compared it against iOS.

The Android data is generated here:

https://github.com/react-native-video/react-native-video/blob/master/android/src/main/java/com/brentvatne/common/react/VideoEventEmitter.java#L370

The type is defined here:

https://github.com/react-native-video/react-native-video/blob/master/src/types/events.ts#L120

## Platform
* Android

## Environment info
Library version: `6.0.0-alpha.9`
Device: Android Pixel 4 (emulator)

## Steps To Reproduce
1. Do something that generates an error in Android. For us, this happens at the end of a video playing in an emulator when it resets to the start.

## Expected behaviour
1. The error shape matches the shape of the `OnVideoErrorData`.

## Actual Error Shape
Here's the actual shape of the error object when there is an error in Android:
```
{
  "error": {
    "errorStackTrace": "...",
    "errorCode": "24003",
    "errorException": "...",
    "errorString": "..."
  }
}
```

The issue is that `errorStackTrace`, `errorCode`, `errorException`, and `errorString` are nested _inside_ of a containing `error` object, whereas the type definition defines it at the top level:

```
export type OnVideoErrorData = Readonly<{
  errorString: string;
  errorException: string;
  errorStackTrace: string;
  errorCode: string;
  error: string;
}>;
```

The `error: string` also appears to be incorrect in this case for Android.

freeboub:
Thank you for the report @brianlenz 
I provided a fix for this issue, let me know if it looks good for you.

brianlenz:
@freeboub yep, #3349 should fix the issue! Thanks 🙏

freeboub:
Merged, will be available in alpha.12 (or beta.1 🤞)

vvdodiya:
Any solution?
I am getting the same error. I am using react-native-video version: "6.4.3" 
Thanks in advance.

PR #3349, opened 2023-11-15
fix: update onError definition to match implementation

freeboub:
Fix onError typescript definition.
It now match native implementation, but this clearly need to be refactored and unified between platforms
Should address: https://github.com/react-native-video/react-native-video/issues/3345


KrzysztofMoch (approved):
LGTM 👍

PR #3354, opened 2023-11-16
fix: Fix types for `ReactVideoSource` to also allow `require(..)` sources

mrousavy:
We need to be careful here to not use `any`, so I used `NodeRequire` - which is afaik present in all React Native environments as a type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,6 @@
 ## Next
 - Android, iOS: add onVolumeChange event #3322
 
-
 ### Version 6.0.0-alpha.9
 - All: add built-in typescript support [#3266](https://github.com/react-native-video/react-native-video/pull/3266)
 - All: update documentation generation [#3296](https://github.com/react-native-video/react-native-video/pull/3296)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,6 @@
 
 ## Next
 - Android, iOS: add onVolumeChange event #3322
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
 
 ## Next
 - Android, iOS: add onVolumeChange event #3322
+- Fix types for ReactVideoSource to also allow require(..) sources
```
freeboub: you can remove this change for now, I will fix PR template, but thank for it 🙏

freeboub:
```suggestion
```

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
...
+
```
freeboub:
```suggestion
```

freeboub:
This Pr doesn't pass the linter step.
you can check locally with "yarn lint" command.
Can you please check ?

notice that ideally we should move require option inside uri prop, but that is another subject !

freeboub (changes requested):

mrousavy:
Oh my bad I only checked typescript output - will fix linting now!

freeboub:
@mrousavy I rework the fix here: https://github.com/react-native-video/react-native-video/pull/3361/files can you please check if it works for you ?
Thank you

freeboub:
My Pr has been merged, don't hesitate to comment on it if you see any issues with it !
Thank you !

Issue #2693, opened 2022-05-25
Move startup props into a single configuration


armands-malejevs:
# Feature Request

Create a new prop where all current props that are only used during initialisation and cannot be changed during runtime into a single configuration prop.

## Why it is needed

This will make it more clear to developer as to which props can be changed during runtime and which props can only be provided when creating the player. This will also simplify the initialisation logic to get rid of `postDelayed` calls etc.

## Tasks

- [x] textTracks (external subtitle tracks)
- [x] contentStartTime
- [X] drm
- [x] Ads props
- [x] bufferConfig
- [x] minLoadRetryCountModifier

wood1986:
@freeboub and @armadilio3. Could you guys please come a list of props which is only for the player initialization? Any change to these props will be ignored after the initialization

`fullscreen`
`controls`

wood1986:
I prefer breaking the people build in alpha after fixing this. I mean throw an exception when people still using original fullscreen and controls. Do you guys have any concerns?

freeboub:
@wood1986 I already flag following props as 'startProps':
- textTracks (external subtitle tracks)
- bufferConfig
- contentStartTime
- drm
- header
- useTunnelingPlayback (a change from my working branch I need to backport)

The good news is that we should be able to implement this change without changing external API.
in Video.js, we can get public input paramters and reformat then to provide it to native.
That's my plan, but I have a lack of time :)

freeboub:
+ localSourceEncryptionKeyScheme for ios shall be moved to source.drm.localSourceEncryptionKeyScheme

freeboub:
> * localSourceEncryptionKeyScheme for ios shall be moved to source.drm.localSourceEncryptionKeyScheme

Done in: https://github.com/TheWidlarzGroup/react-native-video/pull/4136 thank you @KrzysztofMoch

freeboub:
reviewing the description of the ticket, and it seems that headers property implementation is missing: https://thewidlarzgroup.github.io/react-native-video/component/props#headers
I cannot find implementation :/

PR #3355, opened 2023-11-16
feat: implement startPosition

### LINKED ISSUE #2778: Implement startPosition prop to set the start time of a video
# Feature Request

Add a new prop for iOS and Android to be able to set the start time a video opens at. I am open to doing the PR for this, but per the contribution guide I'm opening an issue first.

## Why it is needed

Sometimes users may want to have a video start at a time other than 0:00. It's possible to manually run a .seek in the onLoad prop, but on Android in particular for bigger video files, there is noticeable lag in this starting.  

## Possible implementation

In my local project we used patch-package to make this work on android. In ReactExoplayerView.java in initializePlayerSource, the start time can be used if resumePosition is not present.

Logic to set the start time in getVideoTrackInfoFromManifest would likely be needed to initialize start time as well, something like 
final long startTime = (this.startTime > 0 ? this.startTime : this.contentStartTime) * 1000 - 100; // s -> ms with 100ms offset
But I don't understand that section quite as well.

I haven't looked into the swift side yet and don't have as much knowledge there, but I believe in the setter that would be created, you could set pendingSeek and pendingSeekTime if startTime is passed in, then handleReadyToPlay would automatically seek to the correct time when the player is ready.

### Code sample

After setting the startTime variable, snippet from our android patch-package:
```
@@ -624,11 +625,14 @@ class ReactExoplayerView extends FrameLayout implements
             }
         }
 
-        boolean haveResumePosition = resumeWindow != C.INDEX_UNSET;
-        if (haveResumePosition) {
+        boolean useStartTime = this.startTime > 0;
+        boolean useResumePosition = resumeWindow != C.INDEX_UNSET && resumePosition > 0;
+        if (useResumePosition) {
             player.seekTo(resumeWindow, resumePosition);
+        } else if (useStartTime) {
+            player.seekTo(this.startTime * 1000); // sec to ms
         }
-        player.prepare(mediaSource, !haveResumePosition, false);
+        player.prepare(mediaSource, !useResumePosition && !useStartTime, false);
         playerNeedsSource = false;
```

For iOS, I haven't done much native work but something like this may work

@objc
  func setStartTime(_ startTime:Float) {
      if startTime > 0 {
          _pendingSeek = true
          _pendingSeekTime = startTime
      }
  }

YangJonghun:
#### Update the changelog
- feat: implement `startPosition` prop (defined within source prop)
- (Breaking Change) `source.startTime` have been renamed to `source.cropStart`
- (Breaking Change) `source.endTime` have been renamed to `source.cropEnd`

- this change should test with `source.cropStart`
  - If the `source.cropStart` prop is applied, it will be applied from that point forward.
- this change should test with SSAI

#### Describe the changes
- start any video with given position by `startPosition` prop (ignore zero or negative numbers)
- fix #2778


docs/pages/component/props.md
```diff
@@ -160,10 +160,12 @@ Note on iOS, controls are always shown when in fullscreen mode.
 Note on Android, native controls are available by default.
 If needed, you can also add your controls or use a package like [react-native-video-controls](https://github.com/itsnubix/react-native-video-controls) or [react-native-media-console](https://github.com/criszz77/react-native-media-console), see [Useful Side Project](/projects).
 
+Platforms: Android, iOS
+
 ### `contentStartTime`
 The start time in ms for SSAI content. This determines at what time to load the video info like resolutions. Use this only when you have SSAI stream where ads resolution is not the same as content resolution.
 
-Platforms: Android, iOS
+Platforms: Android
```
freeboub, line 168: good catch !

freeboub:
Point 1:

this change should test with source.startTime
--> should not be an issue, but the startPosition will be applied from startTime
Not sure this is clean, source.startTime / endTime are cutting the video so source.startTime become position 0. This position will be the base for applied startPosition.

this change should test with SSAI: -> I agree


Point 2: 

Can I propose you to move this property inside the source object ?
Why am I asking that ? 
- What is the expected behavior if we just change startPosition without changing ? -> should it restart the stream with startPosition or should it just seek ?
- do it have sense to change source and keep current startPosition ? (I think no)

the background of this request is also the real time issue your are pointing in the media3 PR. the postDelayed is necessary to ensure all props are well applied. If we correctly want to remove this delay when starting the stream, all props necessary to start the playback shall be stored in the same property.

Let me know if this is not clear !

(else the code looks good to me!)

YangJonghun:
> should not be an issue, but the startPosition will be applied from startTime

I agree (I'll make sure to include that in the documentation as well)

> this change should test with SSAI: -> I agree

Unfortunately, I don't have any sample to test

> Can I propose you to move this property inside the source object ?

I deeply agree (I think DRM should be included for the same reason (all props necessary to start the playback shall be stored in the same property.))

I agree with all your points, but I think `startTime` and `startPosition` can be confusing because of their names (as an aside, I think `contentStartTime` is also unintuitive).

So, while it might be a breaking change, I was wondering what you think of the following modification.
```tsx
source={{
  uri: "..."
  startPosition: 0,
  crop: {
    startTime: 0,
    endTime: 0
  }
}}
```

freeboub:
You are right...
For ads sample, It should be easy to test with the basic sample!
Regarding to naming, I agree this is confusing.
I can propose you cropStart and cropEnd, it can be an easier change. (be carefull you need to warn this is a breaking API change ! ) but your proposal works for me also !

> I deeply agree (I think DRM should be included for the same reason (all props necessary to start the playback shall be stored in the same property.))

Yes ! to be follow up in: https://github.com/react-native-video/react-native-video/issues/2693

docs/pages/component/props.md
```diff
@@ -656,18 +658,24 @@ type: 'mpd' }}
 The following other types are supported on some platforms, but aren't fully documented yet:
 `content://, ms-appx://, ms-appdata://, assets-library://`
 
+#### Start playback at a specific point in time
+
+Provide an optional `startPosition` for video. Value is in milliseconds. If the `cropStart` prop is applied, it will be applied from that point forward.
+(If it is zero or negative, it is ignored)
```
freeboub: It should not be ignored if set to 0 !

ios/Video/RCTVideo.swift
```diff
@@ -1177,6 +1183,14 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
             _pendingSeek = false
         }
 
+        if _startPosition > 0 {
```
freeboub:
```suggestion
        if _startPosition >= 0 {
```

ios/Video/RCTVideo.swift
```diff
@@ -1177,6 +1183,14 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+            setSeek([
+                "time": NSNumber(value: _startPosition),
+                "tolerance": NSNumber(value: 100)
+            ])
+            _startPosition = -1
```
freeboub, line 1191: is it really need to force _startPosition to -1 ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -702,7 +703,12 @@ private void initializePlayerSource(ReactExoplayerView self, DrmSessionManager d
         if (haveResumePosition) {
             player.seekTo(resumeWindow, resumePosition);
         }
-        player.prepare(mediaSource, !haveResumePosition, false);
+        if (startPositionMs > 0) {
```
```suggestion
        if (startPositionMs >= 0) {
```

docs/pages/component/props.md
freeboub:
```suggestion
(If it is negative or undefined or null, it is ignored)
```

freeboub (changes requested):
We need to handle 0 as a seek position.
Some video (like live streams) which starts to the end of playback by default.
I think this prop should be also applied in that case.

To be exact, on my product there is a "restart" option on live stream.
It take the EPG and create a kind of VOD with it. This VOD has a close start (always start at the same point), but an open end (video chuck are added according to live updates).

In taht case, exoplayer consider this video as a live stream, thus it starts the video at the end, where, when customer client on restart, he wants video to start from the beggining.

(And thank you for the rework of the prop)

docs/pages/component/props.md
YangJonghun: You are right! I'll fix it

ios/Video/RCTVideo.swift
YangJonghun: I was assuming that the playback would work fine, then an error would occur, and then it would work again.

When the source is unchanged and the lifecycle is as follows, 
```
(1) readyToPlay → (2) failed → (3) readyToPlay
```
I forced it to `-1` because I would expect `startPosition` to not work in state (3).

But I'm not very experienced with iOS, and I don't know much about it. If my assumptions are wrong, feel free to let me know and I'll correct above code.

freeboub (approved):

KrzysztofMoch (approved):
LGTM 👍

ihopeyouwin:
I am wondering what if we do videofile imports through require, and still need to use cropstart/end feature, in that case we cannot directly get access to uri of that file. Maybe it will be better if you consider making it more flexible?

freeboub:
> I am wandering what if we do videofile imports through require, and still need to use cropstart/end feature, in that case we cannot directly get access to uri of that file. Maybe it will be better if you consider making it more flexible?

You are right! I think we should allow (or move ) the require usage inside uri prop of the source!

Issue #3370, opened 2023-11-22
Error when build with beta.0 version with IMA enable


maxinminax:
# Bug

## Platform
Which player are you experiencing the problem on:
* Android

## Environment info

Library version: 6.0.0-beta.0

## Steps To Reproduce

1. enable google ima `RNVUseExoplayerIMA = true`
2. build app
3. error
`error: package com.google.android.exoplayer2.ext.ima does not exist`
`error: package ImaAdsLoader does not exist`

avencat:
I have the same issue on my project, if you want to fix it before this PR is merged you can use `patch-package` to apply the fix on your code: https://github.com/react-native-video/react-native-video/pull/3372

avencat:
My bad my PR wasn't really fixing the issue but this does! 🙂 https://github.com/react-native-video/react-native-video/commit/fa23442268d048c36672901c876275bec80b9956

freeboub:
@maxinminax I just finish another pull request, if you want to have a look at it 🙏

freeboub:
beta.1 has been released. This issue shall now be fixed !

PR #3372, opened 2023-11-22
fix(android): remove ads files only if not used

avencat:
This fixes the usage of ads on Android (see #3370). Since beta0, the ads specific files were excluded only if we said we wanted them to be included.

You can test it by compiling the basic example project and adding this to `build.gradle` file in `buildScript.ext`:
```gradle
RNVUseExoplayerIMA = true
```

I also had to update the `kotlinVersion` variable (I've set it to 1.9.20, also in the `build.gradle` file).


freeboub:
@avencat Thank you for the PR, but I think this is not the good solution
I fixed the build issue here with this patch: https://github.com/react-native-video/react-native-video/pull/3376/commits/fa23442268d048c36672901c876275bec80b9956 See the PR: https://github.com/react-native-video/react-native-video/pull/3376

avencat:
Yes you're absolutely right, this fixes both the build issue and a crash when launching the video! 😉 I will close this PR then! 🙂

PR #3373, opened 2023-11-22
fix: harmonize IMAAdEventType between iOS and Android

avencat:
https://react-native-video.github.io/react-native-video/component/props#adtagurl (it appears the link needs to be fixed too) and iOS and Android parts of https://react-native-video.github.io/react-native-video/installation

```Podfile
$RNVideoUseGoogleIMA = true
```

```gradle
        RNVUseExoplayerIMA = true
```

Then, when an ad completes, you should receive the event `COMPLETED` instead of `COMPLETE` on your [onReceiveAdEvent](https://react-native-video.github.io/react-native-video/component/events#onreceiveadevent) listener on iOS (should already be the case before this PR on Android)

### DIFF
--- a/ios/Video/Features/RCTIMAAdsManager.swift
+++ b/ios/Video/Features/RCTIMAAdsManager.swift
@@ -160,7 +160,7 @@ class RCTIMAAdsManager: NSObject, IMAAdsLoaderDelegate, IMAAdsManagerDelegate, I
                 result = "CLICK";
                 break;
             case .COMPLETE:
-                result = "COMPLETE";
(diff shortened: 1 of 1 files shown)


freeboub (approved):
⚠️  breaking change

avencat:
@freeboub do you need me to do something about this breaking change or it's to notify it in the Changelog maybe?

freeboub:
> @freeboub do you need me to do something about this breaking change or it's to notify it in the Changelog maybe?

@avencat I will handle it, thank you for the change !

PR #3376, opened 2023-11-22
fix(android): ads build and enable ads in android sample

freeboub:
Ensure Sample builds correctly with IMA enable
Related issue: https://github.com/react-native-video/react-native-video/issues/3370

- Fix bad ima import
- Fix missing rework in ima stub
- Apply kotlin workaround
- Apply androidx.core workaround
- enable RNVUseExoplayerIMA = true in sample by default

### how to test
go to sample directory
yarn android
--> you should be able to see ads on big bug bunny sample
comment RNVUseExoplayerIMA = true
yarn android
--> you should NOT be able to see ads on big bug bunny sample (but playback works anyway !)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -103,7 +104,6 @@
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.google.ads.interactivemedia.v3.api.AdEvent;
-import com.google.android.exoplayer2.ext.ima.ImaAdsLoader;
```
avencat, line 106: I think we can remove the local file `com/google/android/exoplayer2/ext/ima/ImaAdsLoader.java` too, right? We can also remove the local `com/google/ads/interactivemedia/v3/AdEvent.java` file and use google's one

android/build.gradle
```diff
@@ -143,14 +143,15 @@ repositories {
 
 def media3_version = safeExtGet('media3Version')
 def kotlin_version = safeExtGet('kotlinVersion')
+def androidxCode_version = safeExtGet('androidxCore')
```
avencat:
```suggestion
def androidxCore_version = safeExtGet('androidxCore')
```
Isn't that supposed to be androidxCore_version ?

android/build.gradle
avencat, line 1: We should be able to remove from line 103 to 112 of this file as we don't need to let these files in the project anymore! 😉

android/build.gradle
freeboub: You are right, let's keep homogeneity ! Thank you

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Yes, but no :)
In fact these files are "stubs".
As there is no conditional compilation in Java, we added a stub to be able to avoid including google IMA package.
These files are here to stub google package. There is maybe a better way to do that instead of this stub, but it works like that for now ...

freeboub: notice also that adding IMA libs in the package imply to configure it, mainly when you push your app on store (AD_ID). This is not expected to handle that by default

android/build.gradle
freeboub: For the same reason of the first point, we need to keep the stub for now !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
avencat: Oh ok! Sorry didn't think through it! 😅 I was using ads and forgot that we could actually not use them at all and so these imports would break, thanks for the explanation! 🙂

examples/basic/android/app/src/main/java/com/videoplayer/MainApplication.java
```diff
@@ -30,7 +30,7 @@ protected List<ReactPackage> getPackages() {
           @SuppressWarnings("UnnecessaryLocalVariable")
           List<ReactPackage> packages = new PackageList(this).getPackages();
           // Packages that cannot be autolinked yet can be added manually here, for example:
-          // packages.add(new MyReactNativePackage());
+          packages.add(new ReactVideoPackage());
```
freeboub, line 33: @KrzysztofMoch This one was missing ! but now it work fine, thank uyou for the tips of removing react-native-video from package.json !

freeboub:
ping @YangJonghun ! here is aditionnal rework from you media3 PR !

avencat (approved):

KrzysztofMoch (approved):

PR #3378, opened 2023-11-23
feat(ad): add adData to onReceiveAdEvent

avencat:
https://react-native-video.github.io/react-native-video/component/props#adtagurl and iOS and Android parts of https://react-native-video.github.io/react-native-video/installation

```Podfile
$RNVideoUseGoogleIMA = true
```

```gradle
        RNVUseExoplayerIMA = true
```

Then, you should be able to do:
```typescript
import { AdEvent, OnReceiveAdEventData } from 'react-native-video'

const MyComponent = () => {
  const onReceiveAdEvent = ({
    adData,
    event
  }: OnReceiveAdEventData) => {
    switch (event) {
      case AdEvent.STARTED:
      case AdEvent.CONTENT_PAUSE_REQUESTED:
        setIsReadingAds(true)
        break
      case AdEvent.COMPLETED:
      case AdEvent.CONTENT_RESUME_REQUESTED:
      case AdEvent.SKIPPED:
        setIsReadingAds(false)
        break
      case AdEvent.LOG:
        captureException(new Error('Ads LOG event'), adData)
        break
      default:
        break
    }
  }

  return (
    <Video
      onReceiveAdEvent={onReceiveAdEvent}
    />
  );
};
```


freeboub (changes requested):
looks good, few comments:

1/  Don't you think it would be better not to add adData WritableMap if it is empty ? It would just be less data transfered in the bridge.
Then adData would be optional in ts

2 / I let you review, but if you don't do the change I think it would be better to add @NonNull to  Map<String, String> adData

3/ can you move "adData" to a const String please ?

4/ you also need to update documentation !

Thank you

avencat:
> looks good, few comments:
> 
> 1/ Don't you think it would be better not to add adData WritableMap if it is empty ? It would just be less data transfered in the bridge. Then adData would be optional in ts
> 
> 2 / I let you review, but if you don't do the change I think it would be better to add @nonnull to Map<String, String> adData
> 
> 3/ can you move "adData" to a const String please ?
> 
> 4/ you also need to update documentation !
> 
> Thank you

Hey @freeboub, I think we're good, I just didn't get the 3rd point so if I still need changes, let me know! 😉

freeboub (approved):

avencat:
@freeboub I added getAdData to stub file, I'm not sure if I did it correctly, can you please have a look? https://github.com/react-native-video/react-native-video/pull/3378/commits/3a04d476d472101e3aec8ccbdf881298741ee972

avencat:
@freeboub I'm wondering if I should just rename the `adData` prop to `data` maybe? it would make an object like this:
```Typescript
{
  event: AdEvent,
  data: Record<string, string>
}
```

What are your thoughts about this?

I'm thinking of this because of #3381, I don't think that the error on `ERROR` event should be stored in `adData` but `data` makes it more accurate, don't you think?

freeboub:
> @freeboub I'm wondering if I should just rename the `adData` prop to `data` maybe? it would make an object like this:
> ```Typescript
> {
>   event: AdEvent,
>   data: Record<string, string>
> }
> ```
> 
> What are your thoughts about this?
> 
> I'm thinking of this because of #3381, I don't think that the error on `ERROR` event should be stored in `adData` but `data` makes it more accurate, don't you think?

I agree, no need of ad prefix!

avencat:
Ok, I updated the code! 🙂

freeboub:
> Ok, I updated the code! 🙂

Did you miss git push ?

avencat:
Capture d’écran 2023-11-27 à 20 59 47
Well… No, I did git push but there's an issue with my commit, it seems Github is still processing it…

avencat:
I really don't understand what's going on, I tried pushing an empty commit then reset to the previous commit and push back but it didn't fix it, the commit is available on my branch: https://github.com/avencat/react-native-video/commit/3e5b413a6a2933c4c1d3c6ef9cf9ac987a3e90e5

freeboub (approved):

freeboub:
Looks good for me, thank you !

PR #3380, opened 2023-11-23
fix(android): fix leak caused by removing lifecycle listener too early

BasixKOR:
This PR fixes some cases where the lifecycle event listeners are removed before it receives `onHostDestroy`, causing the app to not release the resource when some other methods such as `setBufferConfig`.

### Possible leak scenario
1. `<Video />` mounts. `ReactExoplayerView` is created.
2. `bufferConfig` prop is changed, calling the `ReactExoplayerViewManager.setBufferConfig`, `ReactExoplayerView.setBufferConfig`, `ReactExoplayerView.releasePlayer` in order.
3. `releasePlayer` removes the activity lifecycle listener, causing it to not receive `onHostDestroy` properly.
4. After activity is destroyed, the view doesn't know it's destroyed due to the listener removal, causing it to hold the resources that should be handled by `releasePlayer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 
+# [6.0.0-beta.1](https://github.com/react-native-video/react-native-video/compare/v6.0.0-beta.0...v6.0.0-beta.1) (WIP)
+* **android:** fix leak caused by removing lifecycle listener too early ([#3380](https://github.com/react-native-video/react-native-video/pull/3380))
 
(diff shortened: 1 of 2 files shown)


freeboub:
It make sense, looks like it will fix a lot of strange behavior !
Thank you

freeboub (approved):

KrzysztofMoch (approved):

YangJonghun:
@BasixKOR 
안녕하세요!
I made simple PR because there is a bug that doesn't do lifecycle unsubscribe when component unmount.
It would be nice if you could review it together to see if there is a problem. :)

PR #3489

PR #3381, opened 2023-11-24
feat: add onAdError event listener

avencat:
This adds an onAdError event listener so you can have ad errors when your ads fail to load. I managed to test it on Android by providing a fake url to adTagUrl, but not on `iOS`, I don't know how to trigger this event listener… I already use this listener on my project thanks to patch package but thought some people might find it useful so here it is! 😄

https://react-native-video.github.io/react-native-video/component/props#adtagurl and iOS and Android parts of https://react-native-video.github.io/react-native-video/installation

```Podfile
$RNVideoUseGoogleIMA = true
```

```gradle
        RNVUseExoplayerIMA = true
```

Then, you should be able to do:
```typescript
import { AdEvent, OnAdErrorData } from 'react-native-video'

const MyComponent = () => {
  const onAdError = ({
    code,
    message,
    type
  }: OnAdErrorData) => {
    console.log({
      code,
      message,
      type
    })
  }

  return (
    <Video
      onAdError={onAdError}
    />
  );
};
```


freeboub:
code looks good, I just doubting on the added api.
don't you think the error should be added to the main ads callback onReceiveAdEvent ?
It will allow to manage all ads event with only 1 callback..

Can you also confirm you tried to build with IMA dependancy disabled ?

avencat:
Hi @freeboub! I didn't test without IMA dependencies, I just fixed the issues, I think we can add an error event to `onReceiveAdEvent` if you prefer to manage it that way, yes, I'll make the changes! 😉

avencat:
As I said in https://github.com/react-native-video/react-native-video/pull/3378#issuecomment-1827704985 I use `onReceiveAdEvent` but with a different data name (I used `data` instead of `adData` to be more accurate), do you want me to merge this PR in #3378?

freeboub:
> As I said in https://github.com/react-native-video/react-native-video/pull/3378#issuecomment-1827704985 I use `onReceiveAdEvent` but with a different data name (I used `data` instead of `adData` to be more accurate), do you want me to merge this PR in #3378?

No I really prefer small PRs! Thank you for the proposal!

freeboub:
@avencat can you juste merge master branch please ?
I added actions to build all samples (with or without ads).
I will just ensure that PR doesn't break build !
Thanks

freeboub (changes requested):
Additionally,  can you add the new even in doc please ?

freeboub (approved):

PR #3395, opened 2023-12-02
chore: lint project

KrzysztofMoch:
## Summary
Lint and format code (all changes are made by linters)
Most changes are made by linters. The only one made by me is https://github.com/react-native-video/react-native-video/pull/3395/commits/51a329e358130bbfdbd10a2acffef48dbece18f7 & https://github.com/react-native-video/react-native-video/pull/3395/commits/e98a2c00104fbef735f6ae1da33d4f383eb1bae0

## Testplan
- [x] should pass tests


KrzysztofMoch:
Nice, all green ✅ @freeboub @YangJonghun

android/src/main/java/com/brentvatne/common/API/SubtitleStyle.kt
```diff
@@ -1,4 +1,4 @@
-package com.brentvatne.common.API
+package com.brentvatne.common.api
```
freeboub, line 1: It makes sense, but I think you should also rename the folder ?
I think a simple git mv would fix this

freeboub:
It would be better if you move back indent to 4 spaces, here we are loosing all the history :/

YangJonghun:
@KrzysztofMoch
Looks good to me. I'd also suggest using Spotless, although it's not required.

KrzysztofMoch:
> It would be better if you move back indent to 4 spaces, here we are loosing all the history :/

Done ✅

KrzysztofMoch:
> @KrzysztofMoch Looks good to me. I'd also suggest using Spotless, although it's not required.

Thanks for suggestion 🙌, it looks good. But Spotless does not support swift, so implementing it only for android is not worth it IMO (Unless I haven't read about what useful magic feature ✨)

ios/Video/Features/RCTIMAAdsManager.swift
```diff
@@ -1,230 +1,209 @@
 #if USE_GOOGLE_IMA
-import Foundation
```
freeboub, line 2: This change is not very good ... Do you think there is a way to avoid that ?
This is not a blocking point !

freeboub (changes requested):

android/src/main/java/com/brentvatne/common/API/SubtitleStyle.kt
KrzysztofMoch: agree, I missed that

KrzysztofMoch: Done ✅

freeboub (approved):

PR #3408, opened 2023-12-07
fix(android): ensure adTagUrl can be reset

freeboub:
In sample app, once we play the ads sample, next video playback will still use the ads even if it is not requested.
I just improve handling of setAdTagUrl to allow setting value to null.

Can be tested with basic sample app


freeboub:
ping @avencat as you also work with ads

avencat (approved):

PR #3412, opened 2023-12-08
iOS: Use loadTracksWithMediaType so we can see errors

cpoile:
#### Summary

This PR adjusts how text tracks are fetched. Previously we were using `tracksWithMediaType` which has been deprecated  for quite awhile, and it has a funny side effect: failed gets are silently ignored. This caused a problem with one of our servers because it wasn't configured correctly and we didn't understand why subtitles were not being loaded on iOS. Apply this change revealed the problem in our sever config, which we fixed, and now everything works. 

I hope by submitting this upstream, it will help save someone else time in the future. :)

Thank you for all the hard work on this great project.

~~Note: Apologies for the noise trimming whitespace -- it's something my editors are doing automatically on save (even emacs).~~ Fixed.

Please let me know if you would like this applied to the master branch as well and I can submit another PR for that.

### DIFF
--- a/ios/Video/RCTVideo.m
+++ b/ios/Video/RCTVideo.m
@@ -35,26 +35,26 @@ static int const RCTVideoUnset = -1;
   NSURL *_videoURL;
   BOOL _requestingCertificate;
   BOOL _requestingCertificateErrored;
-  
+
   /* DRM */
   NSDictionary *_drm;
   AVAssetResourceLoadingRequest *_loadingRequest;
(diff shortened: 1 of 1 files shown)


freeboub:
Hello @cpoile, thank you for the PR, but we don't plan any new release on 5.2 branch.
Is this change need on V6 version ?
Thank you

KrzysztofMoch:
Closed due to inactivity

PR #3425, opened 2023-12-18
feat: add visionOS support

KrzysztofMoch:
## Summary
Add experimental visionOS support
### TODO:
- [x] Update podspec
- [x] ~~Add visionOS example~~ (since the template visionOS will change yet, it makes no sense to add example now)
- [x] Add visionOS support for [promises](https://github.com/google/promises) - https://github.com/google/promises/pull/224
- [x] Migrate/Fix/Disable deprecated/unavailable functions that are no available in visionOS
  - `AVPlayer`:
    - [x] `AVPlayer.allowsExternalPlayback` - won't work
    - [x] `AVPlayer.isExternalPlaybackActive` - won't work
    - [x] `AVPlayer.preventsDisplaySleepDuringVideoPlayback` - won't work
  - `AVPlayerItem`: - #3449
    - [x] `AVPlayerItem.timedMetadata`
  - `UIApplication` && `UIScreen`: - #3448
    - [x] `UIApplication.shared.statusBarOrientation`
    - [x] `UIApplication.shared.keyWindow`
    - [x] `UIScreen.main.bounds`
  - `AVAsset`: - #3444
    - [x] `AVAsset.tracks(withMediaType:)` 
    - [x] `AVAsset.mediaSelectionGroup(forMediaCharacteristic:)`
  - `Other`:
    - [x] `AVVideoComposition.init(asset:applyingCIFiltersWithHandler:)` - #3444
    - [x] `PlayerViewController.isReadyForDisplay` - won't work
    - [x] `AVAssetResourceLoadingRequest.streamingContentKeyRequestData` - skipped, will be added in future

PR #3438, opened 2023-12-22
Used video instance in a new fullscreen view controller

nicolaosm:
Created Demo (iOS) of how we can make use `presentFullscreenPlayer` Method to open an instance of the existing video in a new fullscreen.

This is one of the most missing features in React Native Video and this approach should provide the functionality needed.

The functionality will allow us to show an embedded video in a fullscreen view using the same video instance.

This PR is a POC of the approach used that works.

The approach is basically takes the AVPlayerLayer(player: _player) instance and adds it to a new UIViewController. This way the same video instance will play in the new UIViewController which can be used to show the video in fullscreen.

The POC needs improvement (for landscape rotation, and custom controls via react native). The video controls I added in the POC are from swift side and need to be removed.

Also, a similar approach needs to be used for Android.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -777,6 +777,8 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         _chapters = chapters
     }
 
+    //Added Custom Player View Controller
+    let customPlayerViewController = UIViewController()
     @objc
     func setFullscreen(_ fullscreen: Bool) {
         if fullscreen && !_fullscreenPlayerPresented && _player != nil {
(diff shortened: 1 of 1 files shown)


KrzysztofMoch:
Hey @nicolaosm, I am currently doing visionOS support (#3425). When I will be done, i will look at it

nicolaosm:
Superb! Thank you @KrzysztofMoch

jordangrant-stadiumlive:
@nicolaosm wondering - does this double declaration of dismissFullscreenPlayer look wrong to you? I was patching something and this stuck out: https://github.com/react-native-video/react-native-video/blob/be80d70ab33750f046396a9da5ca9fb460dca311/ios/Video/RCTVideoManager.m#L84

nicolaosm:
Not sure @jordangrant-stadiumlive! Sorry, I'm not a swift dev. Maybe @KrzysztofMoch has some insight?

Thanks!

KrzysztofMoch:
So I have done some research and there is private API that allows us to present fullscreen with 4 lines of code 👀
We cant go with your proposal as we would like to use native controls 😕

```swift
let selectorName = "enterFullScreenAnimated:completionHandler:"
let selectorToForceFullScreenMode = NSSelectorFromString(selectorName)

if playerViewController.responds(to: selectorToForceFullScreenMode) {
  playerViewController.perform(selectorToForceFullScreenMode, with: true, with: nil)
}
```

PR #3449, opened 2023-12-27
feat(ios): update timed metadata handler

KrzysztofMoch:
## Summary
Part of https://github.com/react-native-video/react-native-video/pull/3425
update timed metadata handler in order to support visionOS

## Test plan
 - [x] Tested in example app
 - [x] CI passed

### DIFF
--- a/ios/Video/Features/RCTPlayerObserver.swift
+++ b/ios/Video/Features/RCTPlayerObserver.swift
@@ -50,7 +50,7 @@ class RCTPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate {
             removePlayerItemObservers()
         }
         didSet {
-            guard let playerItem = playerItem else { return }
+            guard let playerItem else { return }
 
             addPlayerItemObservers()
 
@@ -110,7 +110,7 @@ class RCTPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate {
     }
 
     func addPlayerObservers() {
-        guard let player = player, let _handlers = _handlers else {
+        guard let player, let _handlers else {
             return
         }
 
@@ -125,7 +125,7 @@ class RCTPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate {
     }
 
     func addPlayerItemObservers() {
-        guard let playerItem = playerItem, let _handlers = _handlers else { return }
+        guard let playerItem, let _handlers else { return }
         _playerItemStatusObserver = playerItem.observe(\.status, options: [.new, .old], changeHandler: _handlers.handlePlayerItemStatusChange)
         _playerPlaybackBufferEmptyObserver = playerItem.observe(
             \.isPlaybackBufferEmpty,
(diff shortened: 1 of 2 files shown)


ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -50,9 +50,14 @@ class RCTPlayerObserver: NSObject {
...
-                addPlayerItemObservers()
-            }
+            guard let playerItem = playerItem else { return }
+
+            addPlayerItemObservers()
+
+            // handle timedMetadata
+            let metadataOutput = AVPlayerItemMetadataOutput()
+            playerItem.add(metadataOutput)
+            metadataOutput.setDelegate(self, queue: .global())
```
freeboub: Are you 100% of the .global(). In my understanding, it will handle the update in UI thread, if possible I think we should move it to another thread. Thanks !

KrzysztofMoch: CleanShot 2023-12-28 at 13 03 28@2x
If i undterstand this correctly, `.global()` is what we want to use. `.main` will be for UI

okwasniewski: Every example I found was actually using the main UI queue. https://dcordero.medium.com/hls-timed-metadata-with-avplayer-9e20806ef92f

If you intend to use background queue specify quality of service so those calls get higher priority

ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -50,9 +50,14 @@ class RCTPlayerObserver: NSObject {
             removePlayerItemObservers()
         }
         didSet {
-            if playerItem != nil {
-                addPlayerItemObservers()
-            }
+            guard let playerItem = playerItem else { return }
```
okwasniewski:
```suggestion
            guard let playerItem else { return }
```

KrzysztofMoch:
Thank you @okwasniewski 🙌

ios/Video/Features/RCTPlayerObserver.swift
KrzysztofMoch: Done, moved to main queue ✅

freeboub (approved):

ios/Video/Features/RCTPlayerObserver.swift
```diff
@@ -98,8 +103,14 @@ class RCTPlayerObserver: NSObject {
         }
     }
 
+    func metadataOutput(_: AVPlayerItemMetadataOutput, didOutputTimedMetadataGroups groups: [AVTimedMetadataGroup], from _: AVPlayerItemTrack?) {
+        for metadataGroup in groups {
+            _handlers?.handleTimeMetadataChange(timedMetadata: metadataGroup.items)
+        }
```
okwasniewski: nit: I think you should unwrap the `_handlers` first as it doesn't make sense to iterate when it's nil

okwasniewski (approved):

PR #3464, opened 2024-01-09
fix: support opacity properly

hannojg:
Thanks for opening a PR!

#### Update the documentation

/

#### Update the changelog

/

In the example app set the styles of a video component to opacity: 0.5. The whole video component should now be 50% visible. Without that change, only the video content would have 50% transparency while the video view itself would be a back background.

✅

#### Describe the changes

Texture views are by default opaque. To enable styling the opacity of a video when using texture view we have to disable opaqueness.


KrzysztofMoch (approved):
LGTM 👍, Thanks!

PR #3487, opened 2024-01-19
Fabric (New Architecture) codegen support

yungblud:
#### Update the documentation

#### Update the changelog
I've completed fully functioning codegen generate.

I've added `yarn codgen` script to package.json, so you guys can test codegen by run that script.

* Focused on codegen types.

#### Describe the changes
* moved `src/VideoNativeComponent.ts` to `src/specs/VideoNativeComponent.ts`
* modified react codegen types for running codegen successfully


src/specs/VideoNativeComponent.ts
```diff
@@ -0,0 +1,563 @@
...
+}>;
+// TODO fix type after RN 0.73
+// export type OnTimedMetadataData = Readonly<{
+//   metadata: ReadonlyArray<
+//     Readonly<{
+//       value?: string;
+//       identifier: string;
+//     }>
+//   >;
+// }>;
```
KrzysztofMoch: What do you mean by "fix type after RN 0.73" ? 0.73 is already out 🤔

examples/FabricExample/tsconfig.json
```diff
@@ -4,9 +4,10 @@
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig.json to read more about this file */
     "paths": {
-      "react-native-video": ["../../Video.js"]
+      "react-native-video": ["../../src/index.ts"],
+      "react": [ "./node_modules/@types/react" ]
```
KrzysztofMoch, line 8: This is very good fix 👏

src/specs/VideoNativeComponent.ts
yungblud: oh that means, we have to bump up "react-native" version on top-level's package.json to above 0.73.
Right now, it's below 0.73. (in devDependencies)

So if we want to support array parameters on event callback functions via codegen, I think we can only fully support fabric after RN 0.73 version.

yungblud: But right now, this PR is only for support codegen.
Not implemented native codes yet.
So I just commented out and didn't consider react-native 0.73 on this PR.

examples/FabricExample/tsconfig.json
yungblud: Thanks!

yungblud:
@KrzysztofMoch Let me know if I have to fix or implement other things in this PR.
I've only focused for implementing codegen support in this PR.
Cheers.

src/specs/VideoNativeComponent.ts
yungblud: But, I think it's my bad. It'll affect JS layer's event prop types too.
I will make a fix commit soon.

yungblud: @KrzysztofMoch 
I think these two commits will fix this issue.
https://github.com/react-native-video/react-native-video/pull/3487/commits/0c4bd9ce749d1e6c5f501e2ce4901836bc326314
https://github.com/react-native-video/react-native-video/pull/3487/commits/c9f6a9254592e66c3b9382aa2bebf35dc1ed7c10

yungblud:
@KrzysztofMoch One more issue I found is "headers" on src.
In codegen types, we cannot support `Record<string, string>` type.
So we have to convert them as Array which has {key: string, value: string}.

I did it like this.
https://github.com/react-native-video/react-native-video/pull/3487/files#diff-70773b05560f0897b24b4baa0ed114158fd50e075c0c76de43cc30b90043eb80R16

and created one more util function to fix type issue.
https://github.com/react-native-video/react-native-video/pull/3487/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R9

finally, I implemented that util function to these.
https://github.com/react-native-video/react-native-video/pull/3487/files#diff-272334bc4a560000e8736860b4e6980d3f865b9abea3baee3bfee41ab1894bd6R149
https://github.com/react-native-video/react-native-video/pull/3487/files#diff-272334bc4a560000e8736860b4e6980d3f865b9abea3baee3bfee41ab1894bd6R168

But it'll make android issue like bottom screen shot.
스크린샷 2024-01-20 오후 1 59 33

So I think if we want to fully support codegen, we need to change android's native codes about headers prop?
I want to know your opinion about this issue.
Cheers.

KrzysztofMoch:
Yeah, seems like we can't use `Record<string, string>` with codegen. We should pass it as an array and then on native side convert it to Map to decrease number of changes in codebase
https://github.com/react-native-video/react-native-video/blob/3858a15b4268ae54d5b97c036d86b05aaf31bcf9/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java#L152-L161
So, change expected type to `Array` and convert it to `Map<String, String>`, WDYT ?

yungblud:
@KrzysztofMoch 
Yeah, we already have code about that.
I will show you code fragment about that changes of `setSrc`'s parsing headers part.

```java
        Map<String, String> headers = new HashMap<>();
        ReadableArray propSrcHeadersArray = (src.hasKey(PROP_SRC_HEADERS)) ? src.getArray(PROP_SRC_HEADERS) : null;
        if (propSrcHeadersArray != null) {
            if (propSrcHeadersArray.size() > 0) {
                for (int i = 0; i < propSrcHeadersArray.size(); i++) {
                    ReadableMap current = propSrcHeadersArray.getMap(i);
                    String key = current.hasKey("key") ? current.getString("key") : null;
                    String value = current.hasKey("value") ? current.getString("value") : null;
                    if (key != null && value != null) {
                        headers.put(key, value);
                    }
                }
            }
        }
```

And also we should consider setDRM too. (but approach would be similar)

So if you are okay about that code fragment, I will make an additional commit about this issue.

src/types/video.ts
```diff
@@ -177,7 +156,7 @@ export enum PosterResizeModeType {
 export type AudioOutput = 'speaker' | 'earpiece';
 
 export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
-  source?: ReactVideoSource;
+  source: ReactVideoSource;
```
freeboub: source can be undefined. It is used to stop the player without releasing it. you should keep ? or put | undefined

src/types/video.ts
```diff
@@ -217,9 +196,9 @@ export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
   repeat?: boolean;
   reportBandwidth?: boolean; //Android
   resizeMode?: EnumValues<VideoResizeMode>;
-  selectedAudioTrack?: SelectedTrack;
-  selectedTextTrack?: SelectedTrack;
-  selectedVideoTrack?: SelectedVideoTrack; // android
+  selectedAudioTrack?: SelectedTrackType;
```
freeboub: I don't undestand why you did this change ... 
SelectedTrack is an object representing all possible values to select a track where  SelectedTrackType is an enum representing which kind of track selection can be used

src/Video.tsx
```diff
@@ -118,11 +121,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
       [posterResizeMode],
     );
 
-    const src = useMemo(() => {
-      if (!source) {
```
freeboub, line 124: the is !source shall be kept I think (see my other comment)

src/utils.ts
```diff
@@ -5,6 +5,14 @@ import type {ReactVideoSource, ReactVideoSourceProperties} from './types/video';
 
 type Source = ImageSourcePropType | ReactVideoSource;
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function generateHeaderForNative(obj?: Record<string, any>) {
+  if (!obj) {
+    return [];
```
freeboub, line 11: Are you sure about this change ?
native code is made to handle an object, not an array ...

src/specs/VideoNativeComponent.ts
```diff
@@ -0,0 +1,536 @@
+/* eslint-disable @typescript-eslint/ban-types */
```
freeboub, line 1: Is it mandatory to move this file ?
It makes review harder ... (I think you remove and add instead of git mv)

freeboub (changes requested):

freeboub:
And thank you for extracting in smaller PR ! 🙏

src/types/video.ts
yungblud: Hmm I couldn't fully understand your point.
So you want to use typescript's enum value as selectedAudioTrack's type?

src/utils.ts
yungblud: Ah, It's because of codegen issue.
Codegen cannot parse `Record<string, string>` type.
So that's why I made this util.
And through this utils function, it converts `headers` as array.

src/Video.tsx
yungblud: oh yeah it can be possible. Thanks.
it was my fault.
Fixed on this commit
https://github.com/react-native-video/react-native-video/pull/3487/commits/7944195e9dabcc6d4d9d6c7519903742ccb8df80

src/types/video.ts
yungblud: Fixed on this commit
https://github.com/react-native-video/react-native-video/pull/3487/commits/7944195e9dabcc6d4d9d6c7519903742ccb8df80

src/specs/VideoNativeComponent.ts
yungblud: It's because of handling `{}` on codegen.
I wanted to use Readonly<object> but it cannot be possible.

But, I found another issue on OnReceiveAdEventData while I am looking this issue.
OnReceiveAdEventData['data'] type should be handled as array type I guess.
Because codegen cannot parse Record<string, string> type.
I think I just handle that type as just {} and I missed it.

yungblud: Or, we have to define OnReceiveAdEventData['data'] type as Readonly object if we can define what key and value are expected.

yungblud:
> And thank you for extracting in smaller PR ! 🙏

Thanks my pleasure.

src/specs/VideoNativeComponent.ts
KrzysztofMoch: @yungblud I think the point is that instead of deleting and adding a file (using `git add`), you should use `git mv` to move the file - this keeps the diff between files

KrzysztofMoch:
>So if you are okay about that code fragment, I will make an additional commit about this issue. 

@yungblud for me its okay 🙌

src/specs/VideoNativeComponent.ts
yungblud: @KrzysztofMoch Thanks. That was my mistake.
I will try that!

yungblud: Hmm, I think I cannot do mv file clearly.
I tried multiple times but it's difficult to show just "mv" on file changes.

src/types/video.ts
yungblud: Ah, mate.
I think I just deleted typescript enum types and replaced them from codegen types to recycle codegen types.
Just naming and few types change for codegen types.
I think there would have no problem?

Let me know your opinion.
@freeboub

yungblud:
@KrzysztofMoch @freeboub 
67c648ef079c22e975c63004bde6c4f6667e6882
Can you guys check this new commit?
This commit is for [this issue](https://github.com/react-native-video/react-native-video/pull/3487#issuecomment-1901751343)

And I would be happy if you guys check again my comments.
Thanks.

src/types/video.ts
freeboub: Here is doc:
selectedAudioTrack
Configure which audio track, if any, is played.

selectedAudioTrack={{
  type: Type,
  value: Value
}}

SelectedAudioTrack is an object
SelectedAudioTrack.type is the enum you use.

Is that clearer ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -128,18 +128,19 @@ public void setDRM(final ReactExoplayerView videoView, @Nullable ReadableMap drm
         if (drm != null && drm.hasKey(PROP_DRM_TYPE)) {
             String drmType = ReactBridgeUtils.safeGetString(drm, PROP_DRM_TYPE);
             String drmLicenseServer = ReactBridgeUtils.safeGetString(drm, PROP_DRM_LICENSESERVER);
-            ReadableMap drmHeaders = ReactBridgeUtils.safeGetMap(drm, PROP_DRM_HEADERS);
+            ReadableArray drmHeadersArray = (drm.hasKey(PROP_DRM_HEADERS)) ? drm.getArray(PROP_DRM_HEADERS) : null;
```
freeboub: you should be able to use safeGetArray here :)

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -157,7 +158,20 @@ public void setSrc(final ReactExoplayerView videoView, @Nullable ReadableMap src
         int cropEndMs = ReactBridgeUtils.safeGetInt(src, PROP_SRC_CROP_END, -1);
         String extension = ReactBridgeUtils.safeGetString(src, PROP_SRC_TYPE, null);
 
-        Map<String, String> headers = src.hasKey(PROP_SRC_HEADERS) ? ReactBridgeUtils.toStringMap(src.getMap(PROP_SRC_HEADERS)) : new HashMap<>();
+        Map<String, String> headers = new HashMap<>();
+        ReadableArray propSrcHeadersArray = (src.hasKey(PROP_SRC_HEADERS)) ? src.getArray(PROP_SRC_HEADERS) : null;
```
freeboub: safeGetArray can also be used here

src/utils.ts
freeboub: No problem with this internal refactor!
I saw you changed parsing on android, I think you should also do it on ios ?

freeboub (changes requested):
few minor comment.
Regarding to the sample, won't it make sens to merge the new arch with basic sample?
I don't request it now, I am just asking ! I am afraid it will be painful to maintain multiple sample.

src/types/video.ts
yungblud: We use https://github.com/react-native-video/react-native-video/pull/3487/files#diff-70773b05560f0897b24b4baa0ed114158fd50e075c0c76de43cc30b90043eb80R75 this type from 'specs/VideoNativeComponent.tsx'

스크린샷 2024-01-30 오전 10 38 17

We cannot use typescript's enum type with codegen. It can't be possible.
So I recycled above codegen type for Selected**Track prop types.

I think you want to consider DX for developers by handling `type` as enum.
I think it can be possible, but I'm not sure. I will have a try.

src/utils.ts
yungblud: I will have a look on iOS.
Thanks for the comments!

yungblud:
> few minor comment. Regarding to the sample, won't it make sens to merge the new arch with basic sample? I don't request it now, I am just asking ! I am afraid it will be painful to maintain multiple sample.

Yeah maybe it would be happier. (if we can)
Let me consider them too. Cheers.
@freeboub

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
yungblud: That's a good point. I'll try to fix them.

src/utils.ts
yungblud: @freeboub I handled iOS data structure to this commit https://github.com/react-native-video/react-native-video/pull/3487/commits/6c0fe6885fcbeca107174fe3040ee4812eab74b9
Can you have a check please?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
yungblud: Cool
Have a check on this commit https://github.com/react-native-video/react-native-video/pull/3487/commits/e69666d2458b262e52d2ba01775426a8e224cf4b

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
yungblud: Cool
Have a check on this commit https://github.com/react-native-video/react-native-video/pull/3487/commits/e69666d2458b262e52d2ba01775426a8e224cf4b

src/types/video.ts
yungblud: @freeboub Can you check this commit?
I've handled some issue about this comments.
https://github.com/react-native-video/react-native-video/pull/3487/commits/d1a858a590339b48d5430fbe8b8769f8f1921bd6

I think we have to choose between
* remain docs and modify some types like above commit
* change docs and give up enum types

Let me know your opinion. that commit can be reverted.

src/types/video.ts
```diff
@@ -80,7 +80,7 @@ export enum SelectedTrackType {
 
 export type SelectedTrack = {
   type: SelectedTrackType;
-  value?: string | number;
```
freeboub: Value can be a number in the code ... it is used when we select by index.
Is it problematic for codegen ? I review last commit and it seems working with Int Type !

yungblud: Ah, all types in `src/types/video.ts` are for JS layer's prop types.
The reason I removed `number` is I did want to sync with codegen types of SelectedTrack.
Such as this https://github.com/react-native-video/react-native-video/pull/3487/files/d1a858a590339b48d5430fbe8b8769f8f1921bd6#diff-70773b05560f0897b24b4baa0ed114158fd50e075c0c76de43cc30b90043eb80R85

스크린샷 2024-01-31 오전 10 19 48

I think Codegen do not allow mixed types.

yungblud: Maybe we have to change docs content of this types.
I am not sure what is the best option. But codegen do not allow mixed types.
So maybe we have to change docs?
Let me know your opinion @freeboub

freeboub: ok, I can propose you this change:
- keep current behavior fo external API
- move the value to string only in native bindings (then string To Int in native code)
- do the wrapping in Video.js to convert the int to string

This is the ideal change. but is you really prefer keep it as string only (even in external API), just notice this is a breaking change. (but you need anyway to change the native code)

yungblud: okay I will have a check.
Thanks for the suggestions!

yungblud: @freeboub hello, sorry for the late reply.
I've implemented your suggestions on this commit https://github.com/react-native-video/react-native-video/pull/3487/commits/2ef456a23206201b444c50bd316d42677048dca0

Can you have a look?
If I didn't catch your suggestions, let me know.

freeboub (approved):

yungblud:
@freeboub Hey, thanks for the reviewing my PR.
I've resolved merge conflicts from master branch.

freeboub:
This PR is OK for me @KrzysztofMoch can you please double check ?
BTW @KrzysztofMoch do you think we shall merge it before or after next release ? (I would say just after)

KrzysztofMoch:
Okay, will check that
Yes, I think we should merge it and release in next release for our safety

KrzysztofMoch (approved):
I have tested both iOS & android in basic example - and it's work perfectly
Now @yungblud you can leave it to use, we will release in our time 😄
Thanks for your effort!

@yungblud I have one question, I guess that with this change we will drop support for older versions of react-native. Do you know maybe what is the older version that we can support?

yungblud:
> I have tested both iOS & android in basic example - and it's work perfectly Now @yungblud you can leave it to use, we will release in our time 😄 Thanks for your effort!
> 
> @yungblud I have one question, I guess that with this change we will drop support for older versions of react-native. Do you know maybe what is the older version that we can support?

@KrzysztofMoch Hey, thanks for detailed review of my PR.
Hmm, if you want to consider deprecate older versions of react-native, I think minimum "codegen available" version would be the best option.

https://reactnative.dev/blog/2022/03/30/version-068#opting-in-to-the-new-architecture
I think 0.68.0 would be the first version that RN was released with "New Architecture" support.
So I would suggest set minimum version of RN to 0.68.0 for react-native-video.

freeboub:
I am ready to merge, but there are still conflicts ...
Please merge with master, it will be the next merged PR !

yungblud:
@freeboub Hello! Conflicts were simple. had no problem.
I've resolved it by merging upstream/master.

KrzysztofMoch:
> it will be the next merged PR !

@freeboub I merged it to avoid next possible conflicts (if you are not happy with it please revert)

freeboub:
@KrzysztofMoch I am happy with that :)

Issue #3488, opened 2024-01-19
In Android, ReactExoplayerView instance doesn't eliminate after JS component unmount.


YangJonghun:
# Bug

## Platform
* Android

## Environment info

Library version: 6.0.0-beta.4
Device: Android

## Steps To Reproduce
1. add below Log within `onHostPause` method
```
Log.d(TAG, "onHostPause: " + this.hashCode());
```
2. mount `<Video />` component
3. move to background
4. check log print properly (via Logcat)
```
ReactExoplayerView      <package name>      D  onHostPause: <instance hash code A>
```
5. move to foreground
6. unmount `<Video />` component you created in step 2
7. mount new `<Video />` component
8. move to background
9. you can check stacked log
```
ReactExoplayerView      <package name>      D  onHostPause: <instance hash code A>
ReactExoplayerView      <package name>      D  onHostPause: <instance hash code B>
```
10. repeat steps 5~9, you can see the logs overlap by the number of iterations.

## Expected behaviour
1. memory leak
2. calling unnecessary lifecycle callback

## FYI
This is not an issue in v5.2.1., something seems to have gone wrong with the update process so far

YangJonghun:
PR #3380 missed unmount case
so I made PR #3489

KrzysztofMoch:
Thanks for PR!, will be included in `6.0.0-beta.5`

VMBindraban:
@KrzysztofMoch Do you have an ETA for the `6.0.0-beta.5` release?

KrzysztofMoch:
Hey @VMBindraban, I am planing to make release this week (so max 5 days 😅)

PR #3489, opened 2024-01-19
fix(android): remove lifecycle listener after component unmount

YangJonghun:
#### Update the changelog
- fix(android): remove lifecycle listener after component unmount (resolve issue #3488)

- already described via issue #3488

#### Describe the changes
- remove lifecycle listener after component unmount


KrzysztofMoch (approved):
LGTM, good catch 👍

BasixKOR (commented):
Seems like a better place to handle this. Thanks for the heads-up!

PR #3490, opened 2024-01-20
fix(android): re-layout controls after fullscreen dismiss

### LINKED ISSUE #3377: Android - Controls gets misplaced on screen when closing fullscreen in landscape mode.
# Bug
Android video controls are visible outside the area of video when I exit full screen from landscape mode. (Note - I added a code to make app in portrait mode when closing full screen. This is a required functionality).

## Platform
* Android

## Environment info

Library version: 6.0.0-beta.0
Device: Android 12

## Steps To Reproduce
1. Click full screen button
2. rotate device in landscape mode
3. click full screen button again to exit fullscreen.
4. notice the controls get placed outside the video view.

## Expected behaviour
1. Controls should be in place

## Reproducible sample code

```
<Video
            key={data.id}
            ref={videoRef}
            source={{uri: data.url}}
            style={styles.videoStyle}
            onError={console.log}
            repeat
            playInBackground={false}
            playWhenInactive={true}
            muted={false}
            resizeMode="contain"
            progressUpdateInterval={1000}
            onReadyForDisplay={() => setShowLoader(false)}
            ignoreSilentSwitch="ignore"
            volume={1.0}
            // fullscreenOrientation="landscape"
            preventsDisplaySleepDuringVideoPlayback={true}
            onFullscreenPlayerDidDismiss={() => {
              if (Platform.OS === 'android') {
                Orientation.lockToPortrait();
              }
            }}
            onFullscreenPlayerWillPresent={() => {
              Orientation.unlockAllOrientations();
            }}
            controls={true}
          />
```

## Video sample
[Video Sample](https://drive.google.com/file/d/1NVsIO8Z89OXabw4EfaIymoJCFh2hS4DA/view?usp=sharing)

KrzysztofMoch:
## Summary
re layout controls view after exit from fullscreen to update controls positions and weird bug when controls are out of video

fixes #3377

## Test plan
- [x] Tested in example
- [x] App build in CI

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -387,7 +387,7 @@ public class ReactExoplayerView extends FrameLayout implements
      */
     private void togglePlayerControlVisibility() {
         if (player == null) return;
-        reLayout(playerControlView);
+        reLayoutControls();
         if (playerControlView.isVisible()) {
             playerControlView.hide();
         } else {
@@ -498,6 +498,11 @@ public class ReactExoplayerView extends FrameLayout implements
         view.measure(MeasureSpec.makeMeasureSpec(getMeasuredWidth(), MeasureSpec.EXACTLY),
                 MeasureSpec.makeMeasureSpec(getMeasuredHeight(), MeasureSpec.EXACTLY));
         view.layout(view.getLeft(), view.getTop(), view.getMeasuredWidth(), view.getMeasuredHeight());
+    }
+
+    private void reLayoutControls() {
+        reLayout(exoPlayerView);
+        reLayout(playerControlView);
     }
 
     private class RNVLoadControl extends DefaultLoadControl {
@@ -725,7 +730,8 @@ public class ReactExoplayerView extends FrameLayout implements
         player.prepare();
         playerNeedsSource = false;
 
-        reLayout(exoPlayerView);
+        reLayoutControls();
+
         eventEmitter.loadStart();
         loadVideoStarted = true;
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1995,6 +1995,7 @@ public void setFullscreen(boolean fullscreen) {
             if (controls && fullScreenPlayerView != null) {
                 fullScreenPlayerView.dismiss();
                 reLayout(exoPlayerView);
+                reLayout(playerControlView);
```
freeboub: Thank you for the fix, I think this is a correct change to apply.
BTW can I propose you additional modifications:
- I think there can be some corner case where  playerControlView can be null, can you please test it before accessing to it
-  There is another call to 'reLayout(exoPlayerView);' in initializePlayerSource I think it make sens to apply this change at this place also.
- Then these 2 relayout calls should be put in a function :)
Thank you !

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: > I think there can be some corner case where playerControlView can be null, can you please test it before accessing to it

in `reLayout` we are checking if passed view is not null

> - There is another call to 'reLayout(exoPlayerView);' in initializePlayerSource I think it make sens to apply this change at this place also.
> - Then these 2 relayout calls should be put in a function :)
Thank you !

Done ✅

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -500,6 +500,11 @@ private void reLayout(View view) {
         view.layout(view.getLeft(), view.getTop(), view.getMeasuredWidth(), view.getMeasuredHeight());
     }
 
+    private void reLayoutControls() {
+        reLayout(exoPlayerView);
+        reLayout(playerControlView);
```
freeboub, line 505: Please check playerControlView is not null !

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: But reLayout function has this check

freeboub (approved):

PR #3494, opened 2024-01-23
fix(android): check disableFocus when state is ready

YangJonghun:
#### Update the changelog
fix(android): check disableFocus when state is ready

#### Describe the changes
check `disableFocus` prop when state is ready (currently only works on playback toggle)

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1117,7 +1117,9 @@ public class ReactExoplayerView extends FrameLayout implements
                     if (playerControlView != null) {
                         playerControlView.show();
                     }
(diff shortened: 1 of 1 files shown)


freeboub:
I don't understand why you do this change ...
Audio focus and keep screen on are 2 differents features, controlled independently.
No need of this change in my point of view

YangJonghun:
@freeboub 
You're right, I was just trying to sync the behavior with [this code](https://github.com/react-native-video/react-native-video/blob/master/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L1015-L1017).
I think it would be better to remove the `if (!disableFocus)` from the existing code, what do you think?

freeboub:
@YangJonghun I agree, removing the other reference makes more sense !

YangJonghun:
@freeboub I fixed it. thanks!

KrzysztofMoch (approved):

PR #3496, opened 2024-01-23
fix(android): remove pausePlayback when audio focus loss event

YangJonghun:
#### Update the documentation
fix(android): remove pausePlayback when audio focus loss event

#### Update the changelog
fix(android): remove pausePlayback when audio focus loss event

#### Describe the changes
pausePlayback is added within PR #1897  (commit [1ab82f7](https://github.com/react-native-video/react-native-video/pull/1897/commits/1ab82f773d57efb9310e04faa8b94da8c1efbda8))
but, it makes unintended pauses can be happen with below flow
source ready(auto play) -> pause -> play

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -925,8 +925,6 @@ public class ReactExoplayerView extends FrameLayout implements
                 case AudioManager.AUDIOFOCUS_LOSS:
                     view.hasAudioFocus = false;
                     view.eventEmitter.audioFocusChanged(false);
-                    // FIXME this pause can cause issue if content doesn't have pause capability (can happen on live channel)
-                    view.pausePlayback();
                     view.audioManager.abandonAudioFocus(this);
                     break;
                 case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:


freeboub (approved):
I agree with that, this behavior is an application choice !

KrzysztofMoch (approved):

freeboub:
@KrzysztofMoch @YangJonghun I think again of this patch, I think it will cause some regressions ...
What would be the behaviour when user receive an incoming call ?
Users will have to patch their application to restaure this default behavior...
Do you agree to revert it ?
Ideally we should make this behavior configurable by application with a new prop !

YangJonghun:
@freeboub
Your suggestion is very good.
but usually, we have to pause with AUDIOFOCUS_LOSS event. the problem is developer doesn't have any chance to resume their playback.
I think simple revert would continue the problem, so I was wondering about adding the below code to the AUDIOFOCUS_GAIN event handler.
```
            switch (focusChange) {
                case AudioManager.AUDIOFOCUS_LOSS:
                    view.hasAudioFocus = false;
                    view.eventEmitter.audioFocusChanged(false);
                    view.pausePlayback();
                    view.audioManager.abandonAudioFocus(this);
                    break;
                case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:
                    view.eventEmitter.audioFocusChanged(false);
                    break;
                case AudioManager.AUDIOFOCUS_GAIN:
                    view.hasAudioFocus = true;
                    view.eventEmitter.audioFocusChanged(true);
                    if (!view.isPaused) {
                        view.startPlayback();
                    }
                    break;
                default:
                    break;
            }
```

freeboub:
@YangJonghun Ok, I understand correctly the request know!
You are right, in case the playback has been paused due to audioFocusLoss, we should resume it (but only in that precise case!)
I will revert the change for now.
Ideally, this behavior shall be configurable by application...

Issue #3399, opened 2023-12-04
IOS Memory Leak


kristiantoska:
# Bug

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info

Library version: 6.0.0-beta.0
Device: iPhone 14 Pro Max (Emulator)

## Steps To Reproduce

1. Show video component with an online video (in the example i am using the following 30mb sample video: https://sample-videos.com/video123/mp4/480/big_buck_bunny_480p_30mb.mp4)
2. Unmount the video component
3. Re-Mount the component
4. The Memory usage of the app grows by about the size of the video in my case, and keeps growing with each repeat
5. Profiling also seems to report several memory leaks, with most of the extra memory allocation going into `VM: CoreMedia HTTP cache`
<img width="540" alt="Screenshot 2023-12-04 at 19 25 17" src="https://github.com/react-native-video/react-native-video/assets/20017227/ca1a31e4-6a99-4da8-8acf-c916ddd8d54b">

## Expected behaviour

1. Video data is cleared from memory whenever the `Video` component is unmounted

## Reproducible sample code

Full repo [here](https://github.com/kristiantoska/RNVideo-Memory-Leak-Test): Clean RN init + RNVideo@6.0.0-beta.0 install only

```
import React, {useState} from 'react';
import {SafeAreaView, TouchableOpacity, Text, View} from 'react-native';
import Video from 'react-native-video';

function App() {
  const [hidden, setHidden] = useState(false);

  return (
    <SafeAreaView style={{flex: 1}}>
      <View style={{height: 300, width: '100%'}}>
        {!hidden && (
          <Video
            style={{height: 300, width: '100%'}}
            source={{
              uri: 'https://sample-videos.com/video123/mp4/480/big_buck_bunny_480p_30mb.mp4',
            }}
            controls
            paused={false}
          />
        )}
      </View>

      <TouchableOpacity
        style={{padding: 20, backgroundColor: hidden ? 'red' : 'green'}}
        onPress={() => setHidden(!hidden)}>
        <Text>Video: {hidden ? 'Unmounted' : 'Mounted'}</Text>
      </TouchableOpacity>
    </SafeAreaView>
  );
}

export default App;

```

## Video sample

https://github.com/react-native-video/react-native-video/assets/20017227/beecff73-7bdf-4acf-a70e-c3d6ea49286a

unubold1:
yes i worry about that too. How to fix this memory leak.

[...]

PR #3506, opened 2024-01-29
fix(ios): fix pip memory leak

### LINKED ISSUE #3505: iOS - Picture in picture preventing video instances being released after unmount, eventually crashes
# Bug

I am using a VirtualizedList to show videos. In my code, the video component is only mounted if its list item is in view - the item above and below show an Image preview. Consequently, there is only one Video player instance mounted at any one time.

However, after swiping down the list approximately 10 times (the exact number varies), the video player crashes and I get this error:

```
Error Domain=AVFoundationErrorDomain Code=-11819 "Cannot Complete Action" UserInfo={NSLocalizedRecoverySuggestion=Try again later., NSLocalizedDescription=Cannot Complete Action, NSUnderlyingError=0x283cf3d80 {Error Domain=NSOSStatusErrorDomain Code=-16978 "(null)"}}
```

I have found this error occurs when the maximum number of AV instances is breached on iOS. The same issue has been raised in [another bug report](https://github.com/react-native-video/react-native-video/issues/3399) where the author proposed this solution which solves the issue:

>After looking into this for a while i noticed that the issue is caused by the PIP IOS support. I managed to fix the issue on the latest releases by commenting this line on native as I was not using PIP: 
https://github.com/react-native-video/react-native-video/blob/01d7bedb4160a2be55708d45721eb7b12407519d/ios/Video/RCTVideo.swift#L143

When this line is commented out, the problem goes away. So there is something about PIP that is prevent video instances from being released, even if PIP is not in use.

## Platform
* iOS

## Environment info

Library version: 6.0.0-beta.4 (has been an issue in all 6.0.0 beta versions
Device: iPhone 11, iOS 17. Also iPhone 15 simulator, iOS 17

## Steps To Reproduce

1. Create a FlatList and mount a Video instance on each list item
2. Hide the Video instance when the list item is not in view
3. Scroll through the list
4. Video crashes

## Expected behaviour

1. Video player does not crash

KrzysztofMoch:
## Summary
Fix memory leak caused by pip (`RCTPictureInPicture`).
Seems like RCTPictureInPicture hold strong reference to RCTVideo. With this change I was able to call `deinit` on component unmount

should fix #3505 #3399

## Test Plan
- [x] Tested in example

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -140,7 +140,14 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         _eventDispatcher = eventDispatcher
 
         #if os(iOS)
-            _pip = RCTPictureInPicture(self._onPictureInPictureStatusChanged, self._onRestoreUserInterfaceForPictureInPictureStop)
+            _pip = RCTPictureInPicture(
+                { [weak self] in
+                    self?._onPictureInPictureStatusChanged()
+                },
+                { [weak self] in
+                    self?._onRestoreUserInterfaceForPictureInPictureStop()
+                }
+            )
         #endif
 
         NotificationCenter.default.addObserver(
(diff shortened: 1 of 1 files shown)


freeboub (approved):

xiao99xiao (commented):
~Since `_pip` is declared inside a platform condition statement, `_pip = nil` should be wrapped by such condition as well.~
Just noticed it's fixed in the latest code.

PR #3530, opened 2024-02-13
fix(android): onSeek called instantly

### LINKED ISSUE #3529: [BUG]: onSeek event called instantly on Android
### Version

v6 (Beta)

Android

### Architecture

Old architecture

### What happened?

`onSeek` event is call instantly after `videoRef.seek(time)` without waiting for the Video to load on Android.

In iOS, the event is called after loading the video and started playing at the seeked position.

In this method, the `onSeek` event is called after immediately
https://github.com/react-native-video/react-native-video/blob/f0a237e9c268cf585dc0c580b8585c6fb6c336ee/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L1890-L1895

This behaviour was implemented on this [commit](https://github.com/react-native-video/react-native-video/commit/daf5e595eced2b2f6580e8d8a6205f14facdec46#diff-330fa59e5d22938724c39d72d19e9b97777ea691ed2a37c503a3f948c085ba42L1695-L1700)

Resetting back to the previous behaviour works properly. I could not find a reason why the new way was implemented. So any insight would be helpful.

### Reproduction

On the [Basic Example](https://github.com/react-native-video/react-native-video/tree/master/examples/basic), add the `onSeek` props to check. Aound here https://github.com/react-native-video/react-native-video/blob/f0a237e9c268cf585dc0c580b8585c6fb6c336ee/examples/basic/src/VideoPlayer.tsx#L734-L753

GyanendroKh:
Fixes #3529

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1889,8 +1889,8 @@ public class ReactExoplayerView extends FrameLayout implements
 
     public void seekTo(long positionMs) {
         if (player != null) {
+            seekTime = positionMs;
             player.seekTo(positionMs);
-            eventEmitter.seek(player.getCurrentPosition(), positionMs);
         }
     }
 


KrzysztofMoch (approved):
Thanks! LGTM 👍

PR #3535, opened 2024-02-14
feat: move require (local files) to `source.uri`

KrzysztofMoch:
## Summary
From now "local files" that are imported via `require` will be passed in `source.uri`

### Motivation
This will unblock video config for local files

### Changes
changed `resolveAssetSourceForVideo` function (JS side change only) 

## Test plan
- [x] Tested in example app

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,9 +1,7 @@
 import type {Component, RefObject, ComponentClass} from 'react';
 import {Image, findNodeHandle} from 'react-native';
-import type {ImageSourcePropType} from 'react-native';
 import type {ReactVideoSource, ReactVideoSourceProperties} from './types/video';
 
-type Source = ImageSourcePropType | ReactVideoSource;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function generateHeaderForNative(obj?: Record<string, any>) {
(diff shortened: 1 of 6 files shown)


freeboub (changes requested):
No change on native side ?
Can you also update the sample to be able to test easily ? 
I would also expect a documentation update !

src/utils.ts
```diff
@@ -1,18 +1,24 @@
...
     };
   }
+
+  if (typeof source.uri === 'number') {
+    return {
+      ...source,
+      uri: Image.resolveAssetSource(source.uri).uri,
+    };
+  }
+
```
KrzysztofMoch: @freeboub In stand of making breaking change I will leave old option and add reference in docs to use new way. With this we won't have to make breaking change

> No native side change ?

Yes as you can see, we earlier converted `require` into `{ uri: "some uri to local file" }` so basically I've just added one if statement

freeboub:
@KrzysztofMoch finally successfully tested ! :)
what do you think of this addtionnal commit ? https://github.com/react-native-video/react-native-video/pull/3584/commits/9ab8712e12eab084a56c7640d9c49fc91dd597d7

freeboub (approved):

KrzysztofMoch:
> what do you think of this addtionnal commit ? https://github.com/react-native-video/react-native-video/commit/9ab8712e12eab084a56c7640d9c49fc91dd597d7

Good idea, I have added this

PR #3543, opened 2024-02-26
docs: update installation section

KrzysztofMoch:
## Summary
Update installation section, as static linkage is no more needed

## Test plan
- [x] Tested in example


KrzysztofMoch:
@freeboub Cloud you check if I am right ?

freeboub (approved):

PR #3555, opened 2024-03-01
fix: added nodeModulesPath on metro.config.js of basic example app

yungblud:
## Summary
I couldn't run example basic app.
I've got this error on metro CLI when I run example basic app.
`error: Error: Unable to resolve module @babel/runtime/helpers/interopRequireDefault`

I figured out why I got this error.
I think it's because of metro config.
So I've implemented `nodeModulesPaths` on metro.config.js

### Changes
* example/basic/metro.config.js - nodeModulesPaths

## Test plan


KrzysztofMoch (approved):
Thank you!

Issue #3480, opened 2024-01-16
iOS: onError (AVFoundationErrorDomain) being thrown when changing selectedTextTrack


brycnguyen:
# Bug
Originally I had text track selection working fine in https://github.com/react-native-video/react-native-video/commit/8f1bdb7c36cb3b42925312b9a9952432f4735ba5. One of the other commits in the release is causing this to break.

{"error": {"code": -11800, "domain": "AVFoundationErrorDomain", "localizedDescription": "The operation could not be completed", "localizedFailureReason": "An unknown error occurred (-12783)", "localizedRecoverySuggestion": ""}, "target": 3557}

Being thrown for onError whenever I switch selected text track

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info
Library version: 6.0.0-beta.4
Device: iPhone 14 simulator

## Steps To Reproduce
1. Pass in an array of text tracks to react-native-video
2. Change the selectedTextTrack
3. onError throws error mentioned in the ticket
4. Video does not play

## Expected behaviour
1. Changing the selected text track should continue to play the video

gkueny:
I tried to check this issue, but without any luck for now.

However, if it helps others, here's a minimal repo to reproduce the problem https://github.com/gkueny/VideoErrorThrownOnSelectedTrackReproduction


@brycnguyen with my minimal reproduction mentioned issue is still present when I use `yarn add react-native-video@https://github.com/react-native-video/react-native-video#f1bdb7c36cb3b42925312b9a9952432f4735ba5` (https://github.com/react-native-video/react-native-video/commit/8f1bdb7c36cb3b42925312b9a9952432f4735ba5)

> Originally I had text track selection working fine

did you use the commit in question?
Or just via a patch on one of the versions you use? (& if so which one ?)

KrzysztofMoch:
@gkueny Thank you for sharing repo ❤️
@brycnguyen I have tried v6 beta 4 & 5 and it's not working - not sure what is the issue, because error message is not show any readable info 🫠

side loading subtitles is not liked by apple (there is no official api to do it 😐)

I will try to make some investigation but I am not promise that i will make this work

gkueny:
@KrzysztofMoch I think I have a guess but not sure for now. 
[...]

PR #3557, opened 2024-03-01
fix(ios) add to validTextTracks only if we successfully insertTimeRage

gkueny:
insertTimeRage can fail & if we add failed textTrack to our validTextTracks array, video can crash later on selectTextTrack
we also add en empty textTrack only we we have validTextTrack

related to https://github.com/react-native-video/react-native-video/issues/3480

## Summary

Error https://github.com/react-native-video/react-native-video/issues/3480 with sideLoaded textTrack happen because 

```swift
textCompTrack.insertTimeRange
```

fail, but we already have added textTrack to our `validTextTracks` array.
Later, when we `setSelectedTextTrack` this cause video to crash.

### Motivation

This change do not fix the side loaded textTrack issue, but prevent video crash because of them

### Changes

- append `textTrack` to  `validTextTracks`  only if `textCompTrack.insertTimeRange` run successfully.
- avoid add empty `textTrack` if we have no `validTextTracks` handled (act the same if use give no `textTracks`) 
- print error to give some hint to developer

## Test plan

- Use https://github.com/gkueny/VideoErrorThrownOnSelectedTrackReproduction
- Install react-native-video with patch `yarn add react-native-video@https://github.com/gkueny/react-native-video.git#f0e32d45846003f14680a5292e42a263df2f52b3`
- Run project
- Assert that video run correctly, without (bad?) subtitle...

### DIFF
--- a/ios/Video/Features/RCTVideoUtils.swift
+++ b/ios/Video/Features/RCTVideoUtils.swift
@@ -299,23 +299,32 @@ enum RCTVideoUtils {
                     if let textTracks {
                         for i in 0 ..< tracks.count {
                             guard let track = tracks[i]?.first else { continue } // fix when there's no textTrackAsset
-                            validTextTracks.append(textTracks[i])
 
                             let textCompTrack: AVMutableCompositionTrack! = mixComposition.addMutableTrack(withMediaType: AVMediaType.text,
                                                                                                            preferredTrackID: kCMPersistentTrackID_Invalid)
-                            try? textCompTrack.insertTimeRange(
-                                CMTimeRangeMake(start: .zero, duration: videoAsset.timeRange.duration),
-                                of: track,
-                                at: .zero
-                            )
+
+                            do {
+                                try textCompTrack.insertTimeRange(
+                                    CMTimeRangeMake(start: .zero, duration: videoAsset.timeRange.duration),
+                                    of: track,
+                                    at: .zero
+                                )
+                                validTextTracks.append(textTracks[i])
+                            } catch {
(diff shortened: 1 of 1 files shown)


ios/Video/Features/RCTVideoUtils.swift
```diff
@@ -299,23 +299,30 @@ enum RCTVideoUtils {
...
+                            do {
+                                try textCompTrack.insertTimeRange(
+                                    CMTimeRangeMake(start: .zero, duration: videoAsset.timeRange.duration),
+                                    of: track,
+                                    at: .zero
+                                )
+                                validTextTracks.append(textTracks[i])
+                            } catch {
+                                continue
+                            }
```
gkueny: one downside here, is we fail silently here. What do you recommend ?

KrzysztofMoch: I think we can leave TODO and later when we will "upgrade errors" we will throw here 
Eventually we cloud warn user that something is failing

KrzysztofMoch:
You cooked 🔥 let me test it 👀

KrzysztofMoch (approved):
Tested and It's working 🚀
Thank You!

gkueny:
> Tested and It's working 🚀 Thank You!

Just to be sure that there's no misunderstanding, my PR do not really fix https://github.com/react-native-video/react-native-video/issues/3480#issuecomment-1976149056 because sideloaded subtitle will not be shown to user

I just prevented the crash 


I still haven't found out why `textCompTrack.insertTimeRange` throw an error 😞

PR #3577, opened 2024-03-10
feat: implement capture method

YangJonghun:
## Summary
Implement capture method

### Motivation
Actually this feature(save image file) doesn't related player feature, but it is so much hard to implement within JS side or other native module. so I create this PR.
(If it doesn't fit your development plan, please close it.)

### Changes
add method named `capture`

## Test plan
just call `await videoRef.current.capture()` without below props
- useSurfaceView
- useSecureView
- drm

and check below setup
- https://github.com/react-native-video/react-native-video/pull/3577/files#diff-aa8251da49a770bf1cb8679474b0eb22feff34b6eb7d4104bb838299ea2f2999R67-R91

### DIFF
--- a/android/src/main/java/com/brentvatne/common/toolbox/CaptureUtil.kt
+++ b/android/src/main/java/com/brentvatne/common/toolbox/CaptureUtil.kt
@@ -16,7 +16,6 @@ import com.facebook.react.bridge.ReactApplicationContext
 import java.io.IOException
 import java.io.OutputStream
 
-
 object CaptureUtil {
     @JvmStatic
     fun capture(reactContext: ReactApplicationContext, view: View) {
--- a/android/src/main/java/com/brentvatne/common/toolbox/CaptureUtil.kt
+++ b/android/src/main/java/com/brentvatne/common/toolbox/CaptureUtil.kt
@@ -1,0 +1,93 @@
+package com.brentvatne.common.toolbox
+
+import android.content.ContentValues
+import android.graphics.Bitmap
+import android.net.Uri
+import android.os.Build
+import android.os.Handler
+import android.os.Looper
+import android.provider.MediaStore
+import android.view.PixelCopy
+import android.view.SurfaceView
+import android.view.TextureView
+import android.view.View
+import androidx.media3.common.util.Util
+import com.facebook.react.bridge.ReactApplicationContext
+import java.io.IOException
+import java.io.OutputStream
+
+
+object CaptureUtil {
+    @JvmStatic
+    fun capture(reactContext: ReactApplicationContext, view: View) {
+        val bitmap: Bitmap?
+        if (view is TextureView) {
+            bitmap = view.bitmap
+            try {
+                saveImageToStream(bitmap, reactContext)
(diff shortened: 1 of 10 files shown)


ios/Video/RCTVideo.swift
```diff
@@ -68,6 +68,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     private var _presentingViewController: UIViewController?
     private var _pictureInPictureEnabled = false
     private var _startPosition: Float64 = -1
+    private var _playerOutput: AVPlayerItemVideoOutput?
```
KrzysztofMoch, line 70: Does it increase memory usage significantly ?

YangJonghun: As I know, it depends on resolution or bitrate

src/Video.tsx
```diff
@@ -265,6 +266,13 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
       return VideoManager.setPlayerPauseState(false, getReactTag(nativeRef));
     }, []);
 
+    const capture = useCallback(() => {
+      if (drm) {
+        throw Error('"capture" method can not be called with "drm" prop');
```
freeboub, line 271: throwing an exception is a bit violent ? console.warn should be enough ?

docs/pages/component/methods.mdx
```diff
@@ -64,6 +64,31 @@ Future:
 - Will support more formats in the future through options
 - Will support custom directory and file name through options
 
+### `capture`
+
+<PlatformsList types={['iOS', 'Android']} />
+
+`capture(): Promise<void>`
```
freeboub, line 71: Is it possible to manage the generate file here ?
2 solutions:
- allows to give a path / file path in parameters
- give the generated file path in promise result

Because, here, I cannot know where the file is generated

android/src/main/java/com/brentvatne/common/toolbox/CaptureUtil.kt
```diff
@@ -0,0 +1,92 @@
...
+import android.net.Uri
+import android.os.Build
+import android.os.Handler
+import android.os.Looper
+import android.provider.MediaStore
+import android.view.PixelCopy
+import android.view.SurfaceView
+import android.view.TextureView
+import android.view.View
+import androidx.media3.common.util.Util
```
freeboub, line 14: This import should not be in common folder : import androidx.media3.common.util.Util
Any way to remove it ?

freeboub (changes requested):

src/Video.tsx
freeboub: And BTW, you may defined an empty object in drm... I think you can just remove the test and handle it correctly in native code !

YangJonghun:
@freeboub @KrzysztofMoch 
On second thought, I don't think it's necessary for this feature to be in this library.
I'd like to close this PR if you don't mind, as there is a memory growth issue, so feel free to close it.
(However, if someone asks for a capture feature, I'd like to point them to this PR so they can use it).

PR #3582, opened 2024-03-11
Fix(ios): fix real time issue when fast zapping

freeboub:
## Summary
fix real time issue when fast zapping
When user changes stream url very quicky, sometimes stream is not released and multiple playback are ongoing in the same time 

### Motivation
Have a correct behavior ...

### Changes
Save information when a stream start is in progress and buffer next request to apply it later 
+ add safety checks in the sample
+ fix delay implementation (current implement doesn't work, but it is not really used ...)

## Test plan
very fast channel up in the sample


KrzysztofMoch (approved):
LGTM, well done 🙌

Issue #3579, opened 2024-03-11
[Feature]: Add ability to show/hide subtitles while still selecting a text track.


coofzilla:
### Description

Introduce the ability to independently select and hide text tracks, enabling developers to access subtitle data without displaying it during playback. 

### Why it is needed ?

This feature is crucial for apps that use subtitles to overlay additional information ( e.g., sports scores ) on videos without showing the subtitle text itself. Currently, disabling the `selectedTextTrack` also stops the subtitle data from being accessible; which, limits the creative use of subtitles for overlays. 

1. Increases the flexibility and utility of the `react-native-video` package, aligning with web standards that recommend using subtitles for more than just language support.

2. Allows developers to create a more tailored viewing experience, especially in sports and educational content, where additional information needs to be displayed without interfering with the video content.

> The primary purpose of [WebVTT](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API) files is to add text overlays to a `<video>`

### Possible implementation

```js
<Video
  source={{uri: "video-url"}}
  selectedTextTrack={{
    type: "title",
    value: "English",
  }}
  showSelectedTextTrack={true} // New prop to control subtitle rendering
  onTextTrackDataChanged={event => {
    // Handle the subtitle data
  }}
/>
```

### Code sample

📲 iOS
```swift
// Inside the AVPlayerItem setup
let legibleOutput = AVPlayerItemLegibleOutput()
legibleOutput.suppressesPlayerRendering = !showSelectedTextTrack // Conditional based on the new prop
```

🤖 Android
```java
private void updateSubtitleViewVisibility() {
    subtitleLayout.setVisibility(this.showSelectedTextTrack ? View.VISIBLE : View.GONE);
}
```

freeboub:
Now I correctly understand your use case :)
BTW, I confirm that it would be better in terms of api to use subtitleStyle.opacity to achieve that feature.
you need to:
- add links in the documentation to onTextData -> the goal is to understand the link between features.
- in case of opacity 0 you can put tracks as not rendered like in your sample code.
- if opacity cannot be changed (on ios), you can precise that it must be 0 or 1

coofzilla:
Thank you for the response! once https://github.com/react-native-video/react-native-video/pull/3566 gets merged, I'll work on getting that implemented :) 

Do you think we can add this to the roadmap?

freeboub:
[...]

Issue #3580, opened 2024-03-11
[INFRA]: Github Action Cache


KrzysztofMoch:
## Description

There is something wrong with node_modules cache, it appears on PR check in two scenarios
- in eslint job
- in android build

To fix it I need to manually remove cache from actions and re-run action which is not the best
Will try to fix it, otherwise I will remove cache for node modules

PR #3583, opened 2024-03-11
feat: implement opacity to control visibility of subtitles

coofzilla:
## Summary
Add ability to control subtitle visibility while still maintaining ability of selecting a text track.
### Motivation
see https://github.com/react-native-video/react-native-video/issues/3579
### Changes
Updated `subtitleStyle` to include an `opacity` property; which, can control the visibility of subtitles.

Updated documentation for the new property and provided a link `onTextTrackDataChanged` -> `subtitleStyles` to clarify intent on how to control visibility of subtitles.

## Test plan

✅ Tested on example app:

image

Simply set the following props:

```ts
          selectedTextTrack={{
            type: SelectedTrackType.LANGUAGE,
            value: 'en',
          }}
          subtitleStyle={{
            opacity: 0,
          }}
```

Then observe that the `subtitleTracks` log from `onTextTrackDataChanged` and _do not_ display them :)

Subtitles displayed as expected when set to 1:

image

If not explicitly set, they will be displayed:

image

Type safe, must be 0 or 1
image

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -117,7 +117,12 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
             subtitleLayout.setFixedTextSize(TypedValue.COMPLEX_UNIT_SP, style.getFontSize());
         }
         subtitleLayout.setPadding(style.getPaddingLeft(), style.getPaddingTop(), style.getPaddingRight(), style.getPaddingBottom());
-        subtitleLayout.setVisibility(style.getOpacity() == 0 ? View.GONE : View.VISIBLE);
+        if (style.getOpacity() != 0) {
+            subtitleLayout.setAlpha(style.getOpacity());
+        } else {
+            subtitleLayout.setVisibility(View.GONE);
+        }
+
     }
 
     public void setShutterColor(Integer color) {
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -117,6 +117,7 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
             subtitleLayout.setFixedTextSize(TypedValue.COMPLEX_UNIT_SP, style.getFontSize());
         }
         subtitleLayout.setPadding(style.getPaddingLeft(), style.getPaddingTop(), style.getPaddingRight(), style.getPaddingBottom());
+        subtitleLayout.setVisibility(style.getOpacity() == 0 ? View.GONE : View.VISIBLE);
     }
 
     public void setShutterColor(Integer color) {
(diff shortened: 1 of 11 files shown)


coofzilla:
@KrzysztofMoch I believe the `build android` actions need their cache cleaned and `re-run` they passed https://github.com/react-native-video/react-native-video/actions/workflows/build-android.yml right before I did the lint to remove a white space.

coofzilla:
https://github.com/react-native-video/react-native-video/issues/3580 seems like this might be related.

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -117,6 +117,7 @@ public void setSubtitleStyle(SubtitleStyle style) {
             subtitleLayout.setFixedTextSize(TypedValue.COMPLEX_UNIT_SP, style.getFontSize());
         }
         subtitleLayout.setPadding(style.getPaddingLeft(), style.getPaddingTop(), style.getPaddingRight(), style.getPaddingBottom());
+        subtitleLayout.setVisibility(style.getOpacity() == 0 ? View.GONE : View.VISIBLE);
```
freeboub: I think you should be able to use  
subtitleLayout.setAlpha(style.getOpacity()); if opacity is != 0
It would be more flexible

freeboub (commented):
please see my comment.
I think you can also add the ios change in this PR ?!

coofzilla:
> I think you can also add the ios change in this PR ?!

actually, yeah I can add those changes, I'm going to update it now.

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
coofzilla: done :) implemented as follows:

```java
if (style.getOpacity() != 0) {
    subtitleLayout.setAlpha(style.getOpacity());
} else {
    subtitleLayout.setVisibility(View.GONE);
}
```

coofzilla:
@freeboub Added the `ios` changes, please review :)

Tested:

✅ Opacity set to 0:

image_720


✅ Opacity set to 0.5( not supported; but, no errors ) this is due to limitations with alpha support on `legibleOutput`:
image_480



✅ Opacity set to 1:

image_480-1

coofzilla:
and one more verification for good measure. 

tested with:

```ts
  selectedTextTrack={{
    type: SelectedTrackType.LANGUAGE,
    value: 'en',
  }}
  subtitleStyle={{
    opacity: 0,
  }}
```

image

KrzysztofMoch (approved):
LGTM 👍

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -117,6 +117,12 @@ public void setSubtitleStyle(SubtitleStyle style) {
             subtitleLayout.setFixedTextSize(TypedValue.COMPLEX_UNIT_SP, style.getFontSize());
         }
         subtitleLayout.setPadding(style.getPaddingLeft(), style.getPaddingTop(), style.getPaddingRight(), style.getPaddingBottom());
+        if (style.getOpacity() != 0) {
+            subtitleLayout.setAlpha(style.getOpacity());
```
freeboub, line 121:
```suggestion
            subtitleLayout.setAlpha(style.getOpacity());
            subtitleLayout.setVisibility(View.VISIBLE);
```
if you change opacity from 0 to 1 it will not be visible !

freeboub (changes requested):
one small remark, else it looks good to me !

coofzilla:
> one small remark, else it looks good to me !

great catch, thank you! I think the cache needs to be cleared for the Build Android, and then it should be good for merge? 🥹

freeboub:
@coofzilla Yes this is OK for me!
let's wait for @KrzysztofMoch for the builder cache issue

freeboub (approved):

PR #3589, opened 2024-03-19
fix(ios): fix startPosition, cropStart and cropEnd to handle float values correctly

YangJonghun:
## Summary
fix startPosition, cropStart and cropEnd to handle float values correctly

### Changes
startPosition, cropStart, and cropEnd now accept float values and work correctly

## Test plan

### DIFF
--- a/ios/Video/DataStructures/VideoSource.swift
+++ b/ios/Video/DataStructures/VideoSource.swift
@@ -5,7 +5,7 @@ struct VideoSource {
     let isAsset: Bool
     let shouldCache: Bool
     let requestHeaders: [String: Any]?
-    let startPosition: Int64?
+    let startPosition: Float64?
     let cropStart: Int64?
     let cropEnd: Int64?
     // Custom Metadata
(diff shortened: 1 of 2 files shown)


ios/Video/DataStructures/VideoSource.swift
```diff
@@ -51,9 +51,9 @@ struct VideoSource {
         } else {
             self.requestHeaders = nil
         }
-        self.startPosition = json["startPosition"] as? Int64
-        self.cropStart = json["cropStart"] as? Int64
-        self.cropEnd = json["cropEnd"] as? Int64
+        self.startPosition = json["startPosition"] as? Float64
+        self.cropStart = (json["cropStart"] as? Float64).flatMap { Int64(round($0)) }
```
freeboub, line 55: can you explain why you need .flatMap { Int64(round($0)) } ?
and not in startPosition ?

freeboub (commented):
Looks good, but I really think you need to do the same for android.
may be merged in 2 times @KrzysztofMoch what do you think of it ?
(I check ts typings, they are OK)

ios/Video/DataStructures/VideoSource.swift
YangJonghun: @freeboub 
Thanks for review!

> Looks good, but I really think you need to do the same for android.

Android already works properly with float value.

> can you explain why you need .flatMap { Int64(round($0)) } ?
> and not in startPosition

`startPosition` used within seek function and seek need `CMTimeMakeWithSeconds` that accept float value only.
however `cropStart(cropEnd)` is used make `CMTimeMake`. and it accept Int value only.
so, I used flatMap for safe type casting (If use `json["cropStart"] as? Int64` will be return nil when pass float value)

freeboub: OK, thank you, BTW on android we have this code:
int cropStartMs = ReactBridgeUtils.safeGetInt(src, PROP_SRC_CROP_START, -1);
int cropEndMs = ReactBridgeUtils.safeGetInt(src, PROP_SRC_CROP_END, -1);

It can work but it will be 1ms accurate instead of 1microsec.
Do you agree with that ? (not necessary for the PR)

YangJonghun: @freeboub 
iOS code also works in ms
`rounte($0)` makes rounded value

KrzysztofMoch (approved):

PR #3594, opened 2024-03-21
fix(android): track selection parameter has change in last release.

### LINKED ISSUE #3592: [BUG]: Error while updating 'SelectedTextTrack'
### Version

v6 (Beta)

Android

### Architecture

Old architecture

### What happened?

When I try to change value of the prop selectedTextTrack, the app crashes and throws this error :  

Screenshot_1711015021

### Reproduction

repository link

### Reproduction

Simple implementation of the Video component with textTracks, and controlled selectedTextTrack prop.

freeboub:
## Summary
Move all setSelectedTrack to String instead of Dynamic
Fix regression introduced by: https://github.com/react-native-video/react-native-video/pull/3487

### Motivation
Typing of native module changed from String|Int to String only
Will fix: https://github.com/react-native-video/react-native-video/issues/3592

### Changes
- Move Dynamics to Strings on Java side
- Fix crash issue in sample

## Test plan
Tested on sample by hardcoding the value like in the opened ticket: https://github.com/react-native-video/react-native-video/issues/3592

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -205,11 +205,11 @@ public class ReactExoplayerView extends FrameLayout implements
     private String extension;
     private boolean repeat;
     private String audioTrackType;
-    private Dynamic audioTrackValue;
+    private String audioTrackValue;
     private String videoTrackType;
-    private Dynamic videoTrackValue;
+    private String videoTrackValue;
     private String textTrackType;
-    private Dynamic textTrackValue;
+    private String textTrackValue;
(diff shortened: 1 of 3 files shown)


freeboub:
ping @yungblud

KrzysztofMoch (approved):

yungblud:
> ping @yungblud 

@freeboub sorry for the late reply.
Seems like the bug was from my PR.
I will have a look

freeboub:
> > ping @yungblud 
> 
> @freeboub sorry for the late reply.
> Seems like the bug was from my PR.
> I will have a look

@yungblud it's ok, it is more for information (just to explain why I was afraid to merge very big PR). This PR fix the issue, thank you !

PR #2722, opened 2022-06-15
Memory management improvements

armands-malejevs:
#### Describe the changes
Add improvements to support more control over the video tracks and when to use back buffer.

This PR adds a new prop `limitMaxResolution` which when enabled will remove any video tracks that have a larger resolution than the device screen. This should improve memory usage and shouldn't impact the experienced quality of content.

The second improvement is a new property in the buffer which allows setting the minimum required memory for the back buffer. Since the back buffer is a nice to have feature, but isn't required for a good experience it is good to have it disabled on low end devices to get a little bit more memory available to the UI and forward buffer.

PR #3595, opened 2024-03-21
feat(android): allow limit max video resolution to screen resolution

KrzysztofMoch:
## Summary
port of 
- #2722

### Motivation
- Allows to limit max video resolution to screen resolution - this should help on low end devices 
- Allows to limit bandwidth used on CDN side

### Changes
- added new prop `limitMaxResolution`
- updated docs

## Test plan
- [x] CI pass

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -94,6 +94,7 @@ import androidx.media3.extractor.metadata.id3.Id3Frame;
 import androidx.media3.extractor.metadata.id3.TextInformationFrame;
 import androidx.media3.ui.LegacyPlayerControlView;
 
+import com.brentvatne.common.api.LimitMode;
 import com.brentvatne.common.api.ResizeMode;
 import com.brentvatne.common.api.SubtitleStyle;
 import com.brentvatne.common.api.TimedMetadata;
@@ -122,6 +123,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
@@ -187,7 +189,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private boolean hasDrmFailed = false;
     private boolean isUsingContentResolution = false;
     private boolean selectTrackWhenReady = false;
-    private boolean limitMaxResolutionToScreenSize = false;
+    private String limitMaxResolutionToScreenSize = LimitMode.DISABLED;
 
     private int minBufferMs = DefaultLoadControl.DEFAULT_MIN_BUFFER_MS;
     private int maxBufferMs = DefaultLoadControl.DEFAULT_MAX_BUFFER_MS;
@@ -617,16 +619,6 @@ public class ReactExoplayerView extends FrameLayout implements
         self.trackSelector.setParameters(trackSelector.buildUponParameters()
                 .setMaxVideoBitrate(maxBitRate == 0 ? Integer.MAX_VALUE : maxBitRate));
 
-        Runtime runtime = Runtime.getRuntime();
-        long usedMemory = runtime.totalMemory() - runtime.freeMemory();
-        long freeMemory = runtime.maxMemory() - usedMemory;
-        int backBufferMs = backBufferDurationMs;
-
-        if(freeMemory < (self.enableBackBufferAvailableMemory * 1000 * 1000)) {
-            DebugLog.w("LoadControl", "Available memory is less than required to enable back buffer, setting to 0ms!");
-            backBufferMs = 0;
-        }
-
         DefaultAllocator allocator = new DefaultAllocator(true, C.DEFAULT_BUFFER_SEGMENT_SIZE);
         RNVLoadControl loadControl = new RNVLoadControl(
                 allocator,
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -613,6 +618,16 @@ private void initializePlayerCore(ReactExoplayerView self) {
         self.trackSelector.setParameters(trackSelector.buildUponParameters()
                 .setMaxVideoBitrate(maxBitRate == 0 ? Integer.MAX_VALUE : maxBitRate));
 
+        Runtime runtime = Runtime.getRuntime();
+        long usedMemory = runtime.totalMemory() - runtime.freeMemory();
+        long freeMemory = runtime.maxMemory() - usedMemory;
+        int backBufferMs = backBufferDurationMs;
+
+        if(freeMemory < (self.enableBackBufferAvailableMemory * 1000 * 1000)) {
+            Log.w("LoadControl", "Available memory is less than required to enable back buffer, setting to 0ms!");
```
```suggestion
            DebugLog.w("LoadControl", "Available memory is less than required to enable back buffer, setting to 0ms!");
```

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -16,8 +16,11 @@
 import android.os.Looper;
 import android.os.Message;
 import android.text.TextUtils;
+import android.util.DisplayMetrics;
+import android.util.Log;
```
```suggestion
```

Please use DebugLog instead of Log

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -613,6 +618,16 @@ private void initializePlayerCore(ReactExoplayerView self) {
         self.trackSelector.setParameters(trackSelector.buildUponParameters()
                 .setMaxVideoBitrate(maxBitRate == 0 ? Integer.MAX_VALUE : maxBitRate));
 
+        Runtime runtime = Runtime.getRuntime();
```
freeboub: I don't think this change is necessary in fact...
There are some strange memory management which should not be needed

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -916,6 +931,11 @@ private void releasePlayer() {
             mainHandler.removeCallbacks(mainRunnable);
             mainRunnable = null;
         }
+
+        Runtime runtime = Runtime.getRuntime();
+        if(runtime != null) {
+            runtime.gc();
```
freeboub: please remove this also, calling gc() mannually is not a good practice

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1266,6 +1295,17 @@ private ArrayList<VideoTrack> getVideoTrackInfo() {
         return videoTracks;
     }
 
+    private int getScreenShortestSide(ThemedReactContext context) {
+        DisplayMetrics displayMetrics = new DisplayMetrics();
```
freeboub: looks good, but I also wondering if we should limit / or allow to limit the size to video area resolution instead of screen resolution. 
This can be done later, but maybe we should move setLimitMaxResolution to have a string parameter to allow:
limitMaxResolution='screen' or limitMaxResolution='videoArea'

wdyt ?

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: Good idea 💡

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -195,6 +200,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private double minBufferMemoryReservePercent = ReactExoplayerView.DEFAULT_MIN_BUFFER_MEMORY_RESERVE;
     private Handler mainHandler;
     private Runnable mainRunnable;
+    private double enableBackBufferAvailableMemory = -1f;
```
freeboub, line 203:
```suggestion
```
This one is useless now

docs/pages/component/props.mdx
```diff
@@ -283,6 +283,21 @@ Controls whether the ExoPlayer shutter view (black screen while loading) is enab
...
+
+Controls whether to limit the available videotracks to not exceed device screen resolution or video surface. If enabled the video track list will be limited to resolutions that are smaller or equal to the screen resolution of the device (this also affects automatic video track selection).
+
+- **disabled (default)** - Don't limit the video track resolution.
+- **screen** - Limit the video track resolution to the device screen resolution.
+- **videoSurface** - Limit the video track resoultion to the video surface.
+
+Example:
+```
+limitMaxResolutionToScreenSize={true}
```
freeboub, line 298:
```suggestion
limitMaxResolutionToScreenSize={'screen'}
```

freeboub (changes requested):
few minor change proposed, but I am wondering if the feature really works.
I was expecting to see a call like: setMaxVideoSize
https://developer.android.com/reference/androidx/media3/common/TrackSelectionParameters.Builder#setMaxVideoSize(int,int)

because here you are only patch the report to the UI, but not track selection.
were you able to test the change ?
Thanks

src/types/video.ts
```diff
@@ -180,6 +180,8 @@ export enum PosterResizeModeType {
 
 export type AudioOutput = 'speaker' | 'earpiece';
 
+type LimitMaxResolutionMode = 'screen' | 'videoArea' | 'disabled';
```
freeboub, line 183: no need to export ?

KrzysztofMoch:
@freeboub I will close this one - We need to think a new way to do it

Issue #3501, opened 2024-01-25
onPictureInPictureStatusChanged isActive called twice


nick-mccomb-easygo:
# Bug
When changing `pictureInPicture` to `true` `onPictureInPictureStatusChanged` is called and `isActive` is `true`.

When exiting PIP mode (by pressing the restore button on the PIP video) `onPictureInPictureStatusChanged` is called and `isActive` is `false`.

Then, straight away, `onPictureInPictureStatusChanged` is called and `isActive` is `true`, even though the video is not in PIP mode.

How to stop `onPictureInPictureStatusChanged` firing twice when `pictureInPicture` is set to `false`?

## Platform
Which player are you experiencing the problem on:
* iOS

## Environment info

Library version: 6.0.0-beta.4
Device: iPhone 15 Pro Max (real device)

## Steps To Reproduce

1. Make video go into PIP mode by setting `pictureInPicture` to `true`
2. Make video return from PIP mode by pressing the restore button on the PIP video
3. Observe the `isActive` prop in `onPictureInPictureStatusChanged` (it fires twice)

## Expected behaviour

1. `onPictureInPictureStatusChanged` should fire once when video goes into PIP mode and `isActive` should be `true`
1. `onPictureInPictureStatusChanged` should fire once when video goes out of PIP and `isActive` should be `false`
1. `onPictureInPictureStatusChanged` should not be called again (until `pictureInPicture` is set to true again)

gkueny:
I reproduce the issue too.
I will try to investigate what happen

gkueny:
See https://github.com/gkueny/PiPStatusChangedFireTwice for reproduction.

I have a fix. I will submit a PR this weekend

gkueny:
@KrzysztofMoch fixed by https://github.com/react-native-video/react-native-video/pull/3601

KrzysztofMoch:
Thanks ❤️, let's close

PR #3601, opened 2024-03-23
fix(ios): fix PiP callback

### LINKED ISSUE #3598: [BUG][iOS]: onRestoreUserInterfaceForPictureInPictureStop is never called
### Version

v6 (Beta)

iOS

### Architecture

Old architecture

### What happened?

When we exit  PiP Mode,  `onRestoreUserInterfaceForPictureInPictureStop` is never called

### Reproduction

https://github.com/gkueny/PiPStatusChangedFireTwice

### Reproduction

- click on button to enter PiP mode
- Exit PiP mode
- see that `onRestoreUserInterfaceForPictureInPictureStop` is never called

gkueny:
## Summary

`onPictureInPictureStatusChanged` PiP callback was called twice after PiP exit with bad value, this should be fixed.

`onRestoreUserInterfaceForPictureInPictureStop` was never called,  this should be fixed.

### Motivation

should fix #3598 & #3501

### Changes

- We rename internal call back to`onPictureInPictureEnter` & `onPictureInPictureExit` for better understanding 
- We do not call `onPictureInPictureStatusChanged` twice on PiP exit
- We do call `onRestoreUserInterfaceForPictureInPictureStop` on PiP exit

## Test plan

- Install & run https://github.com/gkueny/PiPStatusChangedFireTwice
- Reproduce the issues 
- Install patch 
```sh
yarn add react-native-video@https://github.com/gkueny/react-native-video.git#20d1576c373ba81718c6788498e26cacd4d963e3
```
- Check that issues are fixed

### DIFF
--- a/ios/Video/Features/RCTPictureInPicture.swift
+++ b/ios/Video/Features/RCTPictureInPicture.swift
@@ -6,27 +6,33 @@ import React
 
 #if os(iOS)
     class RCTPictureInPicture: NSObject, AVPictureInPictureControllerDelegate {
-        private var _onPictureInPictureStatusChanged: (() -> Void)?
+        private var _onPictureInPictureEnter: (() -> Void)?
+        private var _onPictureInPictureExit: (() -> Void)?
         private var _onRestoreUserInterfaceForPictureInPictureStop: (() -> Void)?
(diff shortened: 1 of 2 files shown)


gkueny:
I will test this change in our project. I will adjust this PR if needed after that

PR #3605, opened 2024-03-25
fix: remove `setNativeProps` usage

KrzysztofMoch:
## Summary
Moved prop seek to `VideoManager` function as `setNativeProps` is deprecated
There will be separate PR for windows as we don't have `VideoManager` on it

### Motivation
`setNativeProps` is deprecated and in order to make it work on [Interop Layer](https://github.com/reactwg/react-native-new-architecture/discussions/135) I moved it to component reference function

### Changes
- Remove `seek` prop
- Added `seek` function to `VideoManager`

## Test plan
- [x] CI pass


src/Video.tsx
```diff
@@ -229,17 +229,25 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
 
       Platform.select({
         ios: () => {
-          nativeRef.current?.setNativeProps({
-            seek: {
+          VideoManager.seek(
```
freeboub: I am afraid of impact on windows regarding to typings and ...
BTW, won't it be cleaner to extract 
const callSeekFunction = () => {
          VideoManager.seek(
            {
              time,
              tolerance: tolerance || 0,
            },
            getReactTag(nativeRef),
          );

and then
Platform.select({
        ios: callSeekFunction,
        android: callSeekFunction,
....

freeboub (approved):

freeboub (approved):

PR #3606, opened 2024-03-25
fix: fix getLicense function's type definition

YangJonghun:
## Summary
fix getLicense function's type definition

### Motivation
fix getLicense function's type definition

## Test plan

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -435,10 +435,10 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           const data = event.nativeEvent;
           if (data && data.spcBase64) {
             const getLicenseOverride = drm.getLicense(
-              data.spcBase64,
-              data.contentId,
               data.licenseUrl,
               data.loadedLicenseUrl,
+              data.contentId,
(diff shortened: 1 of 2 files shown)


freeboub:
This change will introduce a breaking change, I think we can avoid it by changing declaration in ts ?
BTW, if we agree to introduce a breaking change, it would be better to put all props in an object ! can be planned for V7 if you want

YangJonghun:
> This change will introduce a breaking change, I think we can avoid it by changing declaration in ts ? BTW, if we agree to introduce a breaking change, it would be better to put all props in an object ! can be planned for V7 if you want

@freeboub 
I don't want breaking change. so I revert previous commit and just fix type definition.
In my thoughts, If we need to change this function, getLicense function should take an object type argument as shown below.
```js
{ spcBase64, contentId, licenseUrl, loadedLicenseUrl }
```
but it isn't neccesary.

freeboub:
Thank you, let's follow up here: https://github.com/react-native-video/react-native-video/issues/3615

freeboub (approved):

PR #3608, opened 2024-03-26
fix: inject onGetLicense prop properly for detect user defined or not

YangJonghun:
## Summary
fix: inject onGetLicense prop properly for detect user defined or not

### Motivation
RCTVideo's onGetLicense property is always not nil even user doesn't define getLicense function

## Test plan


freeboub:
@YangJonghun please fix linter issue

freeboub (approved):

KrzysztofMoch (approved):

PR #3631, opened 2024-03-30
feat!(ios): remove native dependency `promises`

KrzysztofMoch:
⚠️ **Breaking - min iOS, tvOS version bumped**
## Summary
Remove native iOS dependency `promises` and migrate to async/await

### Motivation
We internally discussed that we should remove this dependency as we can go without it. `promises` also required privacy manifest, so generally it should be easier for us in future to go without it.

### Changes
 **min version bumped** 
 - iOS 9.0 -> 13.0
 - tvOS 10.0 -> 13.0

## Test plan
  - [x] on iOS
  - [x] on tvOS
  - [x] on visionOS
- [x] CI pass


KrzysztofMoch:
It is generally ready for review, but I need to test tvOS an visionOS now

ios/Video/Features/RCTVideoDRM.swift
```diff
@@ -76,67 +64,63 @@ enum RCTVideoDRM {
...
-            guard let certificateData else {
-                reject(RCTVideoErrorHandler.noCertificateData)
-                return
+        var certificateData: Data?
+        do {
+            certificateData = try Data(contentsOf: certificateURL)
+            if base64Certificate != nil {
+                certificateData = Data(base64Encoded: certificateData! as Data, options: .ignoreUnknownCharacters)
             }
+        } catch {}
```
freeboub, line 96: maybe add a log here in case of error ?

freeboub (commented):
were you able to test this PR on a content with DRM ?

ios/Video/Features/RCTVideoDRM.swift
KrzysztofMoch: I wasn't able to find any example stream 🥲

KrzysztofMoch:
No i wasn't able to find any example stream

KrzysztofMoch:
@YangJonghun @facugu1998 do you have DRM (FairPlay) stream and could test it ?

YangJonghun:
@freeboub @KrzysztofMoch 
DRM works fine

PR #3636, opened 2024-03-31
fix: fix codegen types

KrzysztofMoch:
## Summary
Fix codegen by changing types to supported one as it does not support string unions

### Motivation
Fix compatibility with codegen 

### Changes
- Updated native view types

## Test plan
- [x] Tested locally
- [x] CI pass

### DIFF
--- a/react-native-video.podspec
+++ b/react-native-video.podspec
@@ -2,6 +2,8 @@ require "json"
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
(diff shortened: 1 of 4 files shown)


freeboub (approved):

KrzysztofMoch:
So folks from expo team told me that we cannot use codegen with Interop Layer - so I propose to temporary disable it. We will turn on it after we migrate to new architecture
Are you okay with it @freeboub ?

freeboub:
@KrzysztofMoch changes looks good to me. Don't you think the codegen removal (temporary) should be done in another PR ? It would be easier to revert.
BTW, is it really still a draft ?

KrzysztofMoch:
Sure so I will make separate PR for disabling it 🙌

PR #3647, opened 2024-04-03
chore: disable codegen

KrzysztofMoch:
## Summary
We can not use Interop Layer with codegen - to be reverted once we integrate new architecture

### Motivation
Add proper support for Interop Layer

### Changes
- removed codegen

## Test plan
- [x] Tested locally
- [x] CI pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,10 +64,5 @@
         "!**/*.tsbuildinfo",
         "!docs",
         "!examples"
-    ],
-    "codegenConfig": {
-        "name": "RNCVideo",
-        "type": "components",
-        "jsSrcsDir": "./src/specs"
-    }
+    ]
 }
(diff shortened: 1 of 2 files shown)


yungblud:
@KrzysztofMoch I am wondering what is Interop Layer that you described.
Can you explain the reason you merged this PR more specifically?

KrzysztofMoch:
> I am wondering what is Interop Layer that you described.

Interop Layer allow to use Old Arch components on New Arch - it's created by facebook to support community with New Arch migration - you can read more [here](https://github.com/reactwg/react-native-new-architecture/discussions/135)

> Can you explain the reason you merged this PR more specifically?

I am not sure now, but there was an issue with build when Interop Layer and Codegen was enabled at the same time - and folks from expo told me to disable it until we complete new arch migration

Issue #3644, opened 2024-04-03
[BUG]: Video is Playing even after video is reached its end , in IOS


Salman-Raj:
### Version

v6 (Beta)

### What platforms are you having the problem on?

iOS

### Architecture

Old architecture

### What happened?

I started playing one Video in iOS Simulator, and even if the video reached its end, it played again continuously, in android, it is working fine

this is my code

class VideoPlayer extends PureComponent<VideoPlayerProps, IState> {

  private videoRef = React.createRef<Video>();
    blurListener: any;
  playerRef: any;
  subscription: any;
  focusSubscription: any;
  constructor(props: VideoPlayerProps) {
    super(props);
    this.videoRef = React.createRef();
    this.state = {
      videoPaused: false,
      playPuse: true,
      duration: "00:00",
      unformatedcurrentTime: "",
      unformatedDuration: responsiveWidth(49),
      currentDuration: 0,
      currentTime: "00:00",
      fillPrecentage: 0,
      videoControls: true,
      fullscreenAndroid: false,
      videoPercentage: 0,
      currentPlayingVideoId: 0,
      currentTimeUnformated: 0,
      loading: false,
      fullyWatched: false,
      orientation: Dimensions.get('screen').width > Dimensions.get('screen').height ? 'landscape' : 'portrait',
      videoEnded:false,
      appState: AppState.currentState
    }
  }

  handleFullScreen = () => {
    this.props.setFullScren?.(this.props.source, true, this.props.index);
  };

  handleExitFullScreen = () => {
    this.props.setFullScren?.(null, false, this.props.index);
  };

  onReadyForDisplay = () => {
    this.setState({ loading: false });
  };

  handleVideoEnd = () => {
    console.log("video ended")
    try{
    if (this.videoRef.current) {
      this.videoRef.current?.setState({paused: true})
      this.videoRef.current?.seek(0);
      setTimeout(()=>{
      this.setState({videoPaused:true,videoEnded:true,playPuse:true})
      },100)
    }
    }
    catch(error:string|unknown){
    console.log(error)
    }
  };

  render() {
    const { videoControls,unformatedDuration, currentTime,currentTimeUnformated } = this.state;
    const { source, fullscreen, videoId } = this.props
    return (
              <Video
                volume={1.0}
                ref={this.videoRef}
                source={{ uri: source }}
                style={(Platform.OS === "android" && fullscreen) ? styles.videoFullscreen : styles.video}
                resizeMode="contain"
[...]

PR #3650, opened 2024-04-03
fix(ios): do not save pause state before seeking

freeboub:
## Summary
Do not save paused state before seeking.
If user change paused prop during seek, it will not be applied (as a pause is done before seeking)

### Motivation
we can have a kind of race condition if user change paused prop during the seek, the initial value will be applied
address ticket: https://github.com/react-native-video/react-native-video/issues/3644

### Changes
- improve sample
- Do not save paused state before seeking and apply expected paused state at the end of seek

## Test plan
Tested by changing the sample to seek at 0 and pause in onEnd callback
see comment: https://github.com/react-native-video/react-native-video/issues/3644#issuecomment-2035089186


KrzysztofMoch (approved):
LGTM
But how about android? Is there also pause state saving? (just asking)

freeboub:
I just fixed conflict, regarding to android implmentation, this issue is not available (diffferent behavior)

PR #3655, opened 2024-04-05
fix(ios): apply PictureInPicture state on start

### LINKED ISSUE #3654: [BUG]: Can't disable Picture in Picture
### Version

v6 (Beta)

iOS

### Architecture

Old architecture

### What happened?

When I run app with following code, PiP is showing when exiting application even if it is disabled
```tsx
    <Video
      source={{
        uri: 'https://www.w3schools.com/html/mov_bbb.mp4',
      }}
      style={{width: '100%', height: 300}}
      controls
      playInBackground
      pictureInPicture={false}
     />
</View>
```

### Reproduction

repository link

### Reproduction

**Need to be run on real device**
1. Play video
2. Exit application
3. PiP is showing even it is disabled

KrzysztofMoch:
## Summary
Fixed PiP on iOS where it couldn't possible to disable PiP

### Motivation
- fixes #3654 

### Changes
- apply pip state on init

## Test plan
- [x] Tested locally
- [x] CI pass

### DIFF
--- a/examples/basic/index.js
+++ b/examples/basic/index.js
@@ -1,0 +1,2 @@
+// Without this file, the example will not build on physical devices
+import './src/index';
--- a/examples/basic/ios/Podfile.lock
+++ b/examples/basic/ios/Podfile.lock
@@ -935,11 +935,49 @@ PODS:
   - React-Mapbuffer (0.74.0-rc.4):
     - glog
     - React-debug
-  - react-native-video (6.0.0-beta.6):
+  - react-native-video (6.0.0-beta.8):
+    - DoubleConversion
+    - glog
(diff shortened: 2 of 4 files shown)


freeboub (approved):

YangJonghun:
@KrzysztofMoch 
Hi!
`pictureInPicture` prop changed from false to true, it assign new `_pip` every time. I think it may cause some bug.
Also, the assignment of `allowsPictureInPicturePlayback` within `init` is duplicated with `createPlayerViewController`.

KrzysztofMoch:
@YangJonghun you right 👀 good catch! - will make PR to address this

Issue #3642, opened 2024-04-02
[BUG]: basic sample project in repo on iOS doesn't show textTracks


paulrinaldi:
### Version

v6 (Beta)

### What platforms are you having the problem on?

iOS

### Architecture

Old architecture

### What happened?

I ran the basic sample app (on `master`) and was trying to find a working example of the textTracks on iOS since I can't get it to work on my own sample repo.
I expected to see subtitles/captions since the title of this one says subtitles.

<img width="387" alt="Screenshot 2024-04-02 at 2 43 02 PM" src="https://github.com/react-native-video/react-native-video/assets/107127397/a11bbe87-fb6e-4390-8430-8a7fd3f0a8a3">

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are: 
1. Run `git clean -xdf && yarn xbasic install && cd examples/basic/ios && pod install && cd - && yarn xbasic ios`*
* At this point I noticed my `Podfile.lock` had some differences with git diff `COCOAPODS: 1.15.2`
2. Click channel up until reaches 'sintel with sideLoaded subtitles', watch until audio occurs.

Observe no captions/subtitles.

paulrinaldi:
@freeboub am I correct that this video in the sample should display textTracks? I am just presuming based off its title.

paulrinaldi:
Additional info:
iOS simulator used: iPhone SE (3rd Generation) iOS 17.4
MacOs Sonoma 14.4.1, Apple M1 Pro

paulrinaldi:
I can confirm that `this.srcList[this.state.srcListId]?.textTracks` is equal to `[{"language": "en", "title": "test", "type": "text/vtt", "uri": "https://bitdash-a.akamaihd.net/content/sintel/subtitles/subtitles_en.vtt"}]` on render
And I can confirm that
`this.state.selectedTextTrack` is equal to `undefined`

freeboub:
> @freeboub am I correct that this video in the sample should display textTracks? I am just presuming based off its title.

Yes I think it should and I also think this was working on 5.2 version

paulrinaldi:
Even when I set `selectedTextTrack={{type: 'index', value: 0}}`, it does not show them.

paulrinaldi:
Same result for `selectedTextTrack={{ type: SelectedTrackType.INDEX, value: 0 }}`

coofzilla:
[...]

PR #3660, opened 2024-04-06
Fix/ensure tracks are available in sample

freeboub:
## Summary
Ensure tracks are available correctly for the app.
onAudioTracks event was reporting some null value
+ minor sample enhancement

### Motivation
Review behavior from: https://github.com/react-native-video/react-native-video/issues/3642
This is only a partial fix, we still need to review extranally loaded text tracks

### Changes
mainly ensure onTextTracks return good value
remove unused function
improve sample 

## Test plan
can be tested on sample


KrzysztofMoch (approved):

paulrinaldi:
I am indeed seeing the subtitles now in the sample :D for "sintel with subtitles" but not for "sintel with sideLoaded subtitles"! Nice work y'all on latest master.

PR #3668, opened 2024-04-09
fix(android): fixed bug where video would not be visible after remount and change of drm source

YangJonghun:
## Summary
fixed bug where video would not be visible after remount and change of drm source

### Motivation
`onAttachedToWindow` called when component remount.
and initialization of player with `drm` prop have networking delay.
these situations cause the `initializePlayer` function to be called again while it is running.
overridden `onAttachedToWindow` code was written with PR #426.
in that time this code needed, but current code doesn't need this behavior. (it can be replaced by an action called inside `setSrc`)

### Changes
- do not initialize player when view attached to window
- remove overridden Android native lifecycles (`onDetachedFromWindow`, `onAttachedToWindow`)

## Test plan
- prepare real device and two dash source with valid drm config.
- change source and drm prop with user action(`changeSource` function changes source)
```tsx
<Button title={'change source'} onPress={changeSource} />
<Video
    // for enforce remount without complicated component hierarchy
    key={source.uri}
    source={source}
    drm={drm}
 />
```

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -322,12 +322,6 @@ public class ReactExoplayerView extends FrameLayout implements
         mainHandler = new Handler();
     }
 
-    @Override
-    protected void onAttachedToWindow() {
-        super.onAttachedToWindow();
-        initializePlayer();
-    }
-
     // LifecycleEventListener implementation
     @Override
     public void onHostResume() {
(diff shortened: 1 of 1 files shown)


freeboub:
This change shall replace following PR: https://github.com/react-native-video/react-native-video/pull/3300/files
For me the change make sens and is OK.
It would be interesting to have more tester for this PR ...

VMBindraban (approved):

freeboub (approved):

KrzysztofMoch (approved):

PR #3385, opened 2023-11-26
feat: implement enterPictureInPictureOnLeave prop for both platform(Android, iOS)

YangJonghun:
#### Update the changelog
feat(android): implement pictureInPicture prop and onPictureInPictureStatusChanged event

#### Describe the changes
- (Breaking Change) Remove `pictureInPicture` boolean prop (iOS)
- Add `enterPictureInPictureOnLeave` boolean prop (Android, iOS)
- Add `enterPictureInPicture` method (Android, iOS)
- Add `exitPictureInPicture` method (Android, iOS)
- Add `onPictureInPictureStatusChanged` event listener (Android)

- Resolve issue #2621

memo)
As I know Activity's `onUserLeaveHint` is acceptable lifecycle method for enter PIP. but it's impossible to detect that method within this library. so, I implement `enterPicturInPicture` within onHostPause.
If someone doesn't like this approach, please leave comment in this PR, I'll give you intent style code.

PR #3672, opened 2024-04-11
fix: remove iOS multiple pip assignments

KrzysztofMoch:
## Summary
Fix multiple assignments from pip in iOS

### Motivation
Improve performance and potencial bugs

### Changes
- updated iOS pip init function 

## Test plan
- [x] Tested Locally
- [x] CI passed

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -70,7 +70,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
             #if os(iOS)
                 if _pictureInPictureEnabled {
                     _playerViewController?.allowsPictureInPicturePlayback = true
-                    
+
                     if _pip == nil {
                         initPictureinPicture()
                     }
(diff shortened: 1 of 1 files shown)


KrzysztofMoch:
@YangJonghun I have fixed with multiple assignments, thanks for catching this one 🙌

About setting `allowsPictureInPicturePlayback` in `createPlayerViewController` - we need this to handle scenario where `_playerViewController` is null when `_pictureInPictureEnabled` is setting up - this is not an problem as in both scenarios we are setting this value to `_pictureInPictureEnabled`

YangJonghun:
@KrzysztofMoch 
This code also has a potential problem.
The `RCTPictureInPicture._pipController` is never reassigned after the `deinitPipController` function is called.

I fixed this in commit [ad06910](https://github.com/react-native-video/react-native-video/pull/3385/commits/ad06910dd013db193c8e3f0c8c12b80e69c83066) and you can implement something similar if you like.

KrzysztofMoch:
To be followed in #3385

PR #3682, opened 2024-04-16
Fix/remove unnecessary import

### LINKED ISSUE #3622: [BUG]: Duplicate interface definition for class 'RCTEventDispatcher'
### Version

v6 (Beta)

iOS

### Architecture

Old architecture

### What happened?

At random times when rebuilding the project with XCode, the build fails with the issue you see below. In theory, nothing should have changed because I ran Build once and then again 10 minutes later and then it fails.

The only way to fix the issue I've found is to `yarn remove react-native-video && pod install && yarn add react-native-video@6 && pod install`. This fixes it every time, but only temporarily until it pops up again a day later.

Screenshot 2024-03-27 at 9 38 25 AM

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are: unclear, perhaps build an app with react-native-video and then wait a day and build again. I am happy to debug if anyone has recommendations.

freeboub:
## Summary
remove unnecessary import: 

### Motivation
fix: https://github.com/react-native-video/react-native-video/issues/3622

### Changes
remove include

## Test plan
build on CI


KrzysztofMoch (approved):
So I was also looking into it but I was able to find why in was added - it seems to work so let's merge this

KrzysztofMoch:
if there will be any issue we will revert

PR #3683, opened 2024-04-16
fix(android): prevent changing resolution video track when video load

YangJonghun:
## Summary
fix(android): prevent changing resolution video track when video load

### Motivation
- for prevent resolution video track changed by react prop when video is initially loaded
- In the above case, the `isUsingContentResolution` and `selectTrackWhenReady` flags don't work

### Changes
- `setSelectedVideoTrack` doesn't call `setSelectedTrack` when `loadVideoStarted` is `true`, it will call within `videoLoaded` after video ready

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1833,7 +1833,7 @@ public class ReactExoplayerView extends FrameLayout implements
     public void setSelectedVideoTrack(String type, String value) {
         videoTrackType = type;
         videoTrackValue = value;
-        setSelectedTrack(C.TRACK_TYPE_VIDEO, videoTrackType, videoTrackValue);
+        if (!loadVideoStarted) setSelectedTrack(C.TRACK_TYPE_VIDEO, videoTrackType, videoTrackValue);
     }
 
     public void setSelectedAudioTrack(String type, String value) {


KrzysztofMoch (approved):
LGTM 👍

freeboub:
Do you use contentStartTime ?

YangJonghun:
@freeboub
no I don't

freeboub:
@YangJonghun Ok, thank you (sorry for the very quick question !)
I don't really understand why you need to do this change as setSelectedTrack will not do anything in case of no video track available.
Additionally, instead of putting doing the setSelected track in onLoad, I think it would make more sense to do it in the onVideoTrackAvailable. In some case, we can have the track list changing during the playback, in that case the selectedVideoTrack will not be reapplied.

Let say it looks good to me, but please can you detail what is the issue of selecting video tracks when the video is not loaded ?

YangJonghun:
@freeboub 
I'm sorry to late response. I'm on vacation, so it's hard to provide any sample code.
It was a problem that video start buffer and stucked.
here is how to reproduce. (I'm not sure to reproduce. it was production's problem)

1. play master mpd source (includes mutiple resolution) with specific resolution track(specify with prop).
3. source prop change (also it is master and includes same resolutions).
4. player doesn't work(on load never called) before call seek function.

I know we have to set resolutions after source is ready, but developers can't easily recognize or misuses even they know that(it is so hard to debug)

PR #3707, opened 2024-04-23
fix(ios): keep subtitle selection on foreground

Myst1024:
## Summary
- When playing an `m3u8` file with embedded subtitle tracks, the user's subtitle selections are wiped out whenever the player is foregrounded.  This PR prevents that scenario

### Motivation
- Users were losing subtitle settings when the app was minimized/the screen was turned off, then the app was foregrounded again.

### Changes
- Only update `selectedTextTrack` if there is a value being passed in `selectedTextTrackCriteria`

## Test plan
- On IOS, Play an `m3u8` file that includes embedded subtitle tracks such as the [sintel with subtitles](https://github.com/react-native-video/react-native-video/blob/e8ce4979b38a8ef6f0db9ae6049f4d1c230bd280/examples/basic/src/VideoPlayer.tsx#L133) example
- Choose one of the embedded subtitle tracks
- Minimize the app or turn off the screen
- Foreground the app or turn the screen back on
- The previously selected subtitles should continue to display when the video resumes

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -848,9 +848,12 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
             setMaxBitRate(_maxBitRate)
         }
 
+        if _selectedTextTrackCriteria != nil {
+            setSelectedTextTrack(_selectedTextTrackCriteria)
+        }
+
         setAudioOutput(_audioOutput)
         setSelectedAudioTrack(_selectedAudioTrackCriteria)
-        setSelectedTextTrack(_selectedTextTrackCriteria)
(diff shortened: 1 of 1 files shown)


KrzysztofMoch (approved):
LGTM 👍, but should we do same for audioTrack?

freeboub:
Your patch doesn't look problematic, but I am disturbed as I don't reproduce the issue ... do you have specific props to get the issue ? (maybe you use controls ?)

Myst1024:
> Your patch doesn't look problematic, but I am disturbed as I don't reproduce the issue ... do you have specific props to get the issue ? (maybe you use controls ?)

I have `controls={true}` and `playWhenInactive={true}` which I could see contributing to this.
Using version `6.0.0-rc.0`

Issue #3716, opened 2024-04-25
[BUG]: Ads audio still playing when exiting the video (iOS only)


tenet-angelstrizioli:
### Version

"react-native-video": "6.0.0-rc.0"

### What platforms are you having the problem on?

iOS

### Architecture

Old architecture

### What happened?

Using adTagUrl on iOS, load an ad and then go back or unmount the video component you can still hear the audio of the ad for 4-5 seconds.
This happen both on emulator and real device.

https://github.com/react-native-video/react-native-video/assets/140423003/29029402-77f1-4ab0-a962-5b5d4295c0c9

On android is working as expected. 

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are: 
1. Create a react app
2. Install react-native-video@6.0.0-rc.0
3. Create a video component and add adTagUrl 
4. Enter video and go back, you will hear for 4-5 seconds the audio of the ad loaded before.

Expected behaviour:
When exiting the video the ad should stop reproducing the audio.

Khuongnb:
They likely forgot to call the destroy method when the player is detached from the super view. While waiting for the fixes, you can use my [patch-package](https://www.npmjs.com/package/patch-package) file for a temporary solution.

Please note that in this patch file, I've also added 2 events for iOS, (AD_CUEPOINTS & AD_ERROR) to grab all ad positions, error when loading adTagUrl and additional metadata of the playing ad break to know which ad position is playing (use cases are included in the file), these event will not included in others release without patching so use at your own risk.

[react-native-video+6.0.0-rc.0.patch](https://github.com/react-native-video/react-native-video/files/15147699/react-native-video%2B6.0.0-rc.0.patch)

tenet-angelstrizioli:
This works like a charm, thank you a lot! 
I don't know if i should leave this ticket open so they can add this solution to the library logic or close it cause this solution its working

KrzysztofMoch:
@Khuongnb do you want to create PR 😄 or I should do it ?

freeboub:
Thank you for the patch, I confirm this is looking good, just need to add new events in the documentation !

Khuongnb:
Sorry for the late response. I created a PR for this issue, which only addresses the ad audio problem. The new ad events will need to be implemented on Android before they can be used to ensure functionality on both platforms, I will open another PR when it's available. Thanks everyone.

freeboub:
PR has been merged, will be available in 6.0.0-RC.1

[...]

PR #3722, opened 2024-04-30
fix(ios): destroy adsManager when player detach from super view (#3716)

Khuongnb:
## Summary
* In iOS/tvOS, when the player is detached from the superview, the adsManager is not destroyed and keeps playing ads until they finish.

### Motivation
* Issue #3716

### Changes
* Added `releaseAds` method to `RCTIMAAdsManager`.
* Called `releaseAds()` in `removeFromSuperview` of `RCTVideo`.

## Test Plan
* Remove player when ads are playing; the adsManager should stop producing sound immediately.

### DIFF
--- a/ios/Video/Features/RCTIMAAdsManager.swift
+++ b/ios/Video/Features/RCTIMAAdsManager.swift
@@ -49,9 +49,9 @@
             // Destroy AdsManager may be delayed for a few milliseconds
             // But what we want is it stopped producing sound immediately
             // Issue found on tvOS 17, or iOS if view detach & STARTED event happen at the same moment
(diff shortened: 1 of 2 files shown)


KrzysztofMoch:
build fails because of pods cache... I will build this locally

freeboub:
> build fails because of pods cache... I will build this locally 

I have a doubt that it builds without IMA, please double check that !

KrzysztofMoch (approved):
LGTM 🙌 
@freeboub I was able to build successfully app with & without ADS 
(not sure why CI builds fails tbh)

PR #3723, opened 2024-04-30
feat: add notification controls

KrzysztofMoch:
## Summary
Add `showNotificationControls` prop to Android & iOS

### Motivation
Allow to control video from notification controls

## Test plan
- [x] Tested locally
- [x] CI passes

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -175,6 +175,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private ExoPlayer player;
     private DefaultTrackSelector trackSelector;
     private boolean playerNeedsSource;
+    private MediaMetadata customMetadata;
 
     private ServiceConnection playbackServiceConnection;
     private PlaybackServiceBinder playbackServiceBinder;
@@ -860,7 +861,12 @@ public class ReactExoplayerView extends FrameLayout implements
         }
         config.setDisableDisconnectError(this.disableDisconnectError);
 
-        MediaItem.Builder mediaItemBuilder = new MediaItem.Builder().setUri(uri);
+        MediaItem.Builder mediaItemBuilder = new MediaItem.Builder()
+                .setUri(uri);
+
+        if (customMetadata != null) {
+            mediaItemBuilder.setMediaMetadata(customMetadata);
+        }
 
         if (adTagUrl != null) {
             mediaItemBuilder.setAdsConfiguration(
@@ -1608,7 +1614,7 @@ public class ReactExoplayerView extends FrameLayout implements
 
     // ReactExoplayerViewManager public api
 
-    public void setSrc(final Uri uri, final long startPositionMs, final long cropStartMs, final long cropEndMs, final String extension, Map<String, String> headers) {
+    public void setSrc(final Uri uri, final long startPositionMs, final long cropStartMs, final long cropEndMs, final String extension, Map<String, String> headers, MediaMetadata customMetadata) {
         if (uri != null) {
             boolean isSourceEqual = uri.equals(srcUri) && cropStartMs == this.cropStartMs && cropEndMs == this.cropEndMs;
             hasDrmFailed = false;
@@ -1621,6 +1627,21 @@ public class ReactExoplayerView extends FrameLayout implements
             this.mediaDataSourceFactory =
                     DataSourceUtil.getDefaultDataSourceFactory(this.themedReactContext, bandwidthMeter,
                             this.requestHeaders);
+
+            if (this.customMetadata != customMetadata && player != null) {
+                MediaItem currentMediaItem = player.getCurrentMediaItem();
+
+                if (currentMediaItem == null) {
+                    return;
+                }
+
+                MediaItem newMediaItem = currentMediaItem.buildUpon().setMediaMetadata(customMetadata).build();
+
+                // This will cause video blink/reload but won't louse progress
+                player.setMediaItem(newMediaItem, false);
+            }
+
+            this.customMetadata = customMetadata;
 
             if (!isSourceEqual) {
                 reloadSource();
@@ -1639,6 +1660,7 @@ public class ReactExoplayerView extends FrameLayout implements
             this.extension = null;
             this.requestHeaders = null;
             this.mediaDataSourceFactory = null;
+            this.customMetadata = null;
             clearResumePosition();
         }
     }
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -10,10 +10,15 @@ import static androidx.media3.common.C.TIME_END_OF_SOURCE;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.ActivityManager;
+import android.content.ComponentName;
 import android.content.Context;
+import android.content.Intent;
+import android.content.ServiceConnection;
 import android.media.AudioManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Handler;
+import android.os.IBinder;
 import android.os.Looper;
 import android.os.Message;
 import android.text.TextUtils;
@@ -92,6 +97,7 @@ import androidx.media3.exoplayer.upstream.DefaultBandwidthMeter;
 import androidx.media3.extractor.metadata.emsg.EventMessage;
 import androidx.media3.extractor.metadata.id3.Id3Frame;
 import androidx.media3.extractor.metadata.id3.TextInformationFrame;
+import androidx.media3.session.MediaSessionService;
 import androidx.media3.ui.LegacyPlayerControlView;
 
 import com.brentvatne.common.api.ResizeMode;
@@ -170,6 +176,9 @@ public class ReactExoplayerView extends FrameLayout implements
     private DefaultTrackSelector trackSelector;
     private boolean playerNeedsSource;
 
+    private ServiceConnection playbackServiceConnection;
+    private PlaybackServiceBinder playbackServiceBinder;
+
     private int resumeWindow;
     private long resumePosition;
     private boolean loadVideoStarted;
@@ -228,6 +237,8 @@ public class ReactExoplayerView extends FrameLayout implements
     private String[] drmLicenseHeader = null;
     private boolean controls;
     private Uri adTagUrl;
+
+    private boolean showNotificationControls = false;
     // \ End props
 
     // React
@@ -344,6 +355,12 @@ public class ReactExoplayerView extends FrameLayout implements
     @Override
     public void onHostDestroy() {
         cleanUpResources();
+    }
+
+    @Override
+    protected void onDetachedFromWindow() {
+        cleanUpResources();
+        super.onDetachedFromWindow();
     }
 
     public void cleanUpResources() {
@@ -651,6 +668,10 @@ public class ReactExoplayerView extends FrameLayout implements
         PlaybackParameters params = new PlaybackParameters(rate, 1f);
         player.setPlaybackParameters(params);
         changeAudioOutput(this.audioOutput);
+
+        if(showNotificationControls) {
+            setupPlaybackService();
+        }
     }
 
     private DrmSessionManager initializePlayerDrm(ReactExoplayerView self) {
@@ -734,6 +755,55 @@ public class ReactExoplayerView extends FrameLayout implements
         initializePlayerControl();
         setControls(controls);
         applyModifiers();
+    }
+
+    private void setupPlaybackService() {
+        if (showNotificationControls & player != null) {
+            playbackServiceConnection = new ServiceConnection() {
+                @Override
+                public void onServiceConnected(ComponentName name, IBinder service) {
+                    playbackServiceBinder = (PlaybackServiceBinder) service;
+
+                    try {
+                        playbackServiceBinder.getService().registerPlayer(player);
+                    } catch (Exception e) {
+                        DebugLog.e("ExoPlayer Playback", "Cloud not register ExoPlayer");
+                    }
+                }
+
+                @Override
+                public void onServiceDisconnected(ComponentName name) {
+                    playbackServiceBinder = null;
+                }
+
+                @Override
+                public void onNullBinding(ComponentName name) {
+                    DebugLog.e("ExoPlayer Playback", "Cloud not register ExoPlayer");
+                }
+            };
+
+            Intent intent = new Intent(themedReactContext, VideoPlaybackService.class);
+            intent.setAction(MediaSessionService.SERVICE_INTERFACE);
+
+            themedReactContext.startService(intent);
+
+            int flags;
+            if (Build.VERSION.SDK_INT >= 29) {
+                flags = Context.BIND_AUTO_CREATE | Context.BIND_INCLUDE_CAPABILITIES;
+            } else {
+                flags = Context.BIND_AUTO_CREATE;
+            }
+
+            themedReactContext.bindService(intent, playbackServiceConnection, flags);
+        }
+
+        if(!showNotificationControls && playbackServiceBinder != null) {
+            if(player != null) {
+                playbackServiceBinder.getService().unregisterPlayer(player);
+            }
+
+            themedReactContext.unbindService(playbackServiceConnection);
+        }
     }
 
     private DrmSessionManager buildDrmSessionManager(UUID uuid, String licenseUrl, String[] keyRequestPropertiesArray) throws UnsupportedDrmException {
@@ -923,12 +993,20 @@ public class ReactExoplayerView extends FrameLayout implements
             if (adsLoader != null) {
                 adsLoader.setPlayer(null);
             }
+
+            if(playbackServiceBinder != null) {
+                playbackServiceBinder.getService().unregisterPlayer(player);
+                themedReactContext.unbindService(playbackServiceConnection);
+            }
+
             updateResumePosition();
             player.release();
             player.removeListener(this);
             trackSelector = null;
+
             player = null;
         }
+
         if (adsLoader != null) {
             adsLoader.release();
         }
@@ -1942,6 +2020,20 @@ public class ReactExoplayerView extends FrameLayout implements
 
     public void setContentStartTime(int contentStartTime) {
         this.contentStartTime = contentStartTime;
+    }
+
+    public void setShowNotificationControls(boolean showNotificationControls) {
+        this.showNotificationControls = showNotificationControls;
+
+        if (playbackServiceConnection == null && showNotificationControls) {
+            setupPlaybackService();
+        }
+
+        if(!showNotificationControls && playbackServiceBinder != null) {
+            if (player != null) {
+                playbackServiceBinder.getService().unregisterPlayer(player);
+            }
+        }
     }
 
     public void setDisableBuffering(boolean disableBuffering) {
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1631,6 +1631,20 @@ public class ReactExoplayerView extends FrameLayout implements
     // ReactExoplayerViewManager public api
 
     public void setSrc(final Uri uri, final long startPositionMs, final long cropStartMs, final long cropEndMs, final String extension, Map<String, String> headers, MediaMetadata customMetadata) {
+
+        if (this.customMetadata != customMetadata && player != null) {
+            MediaItem currentMediaItem = player.getCurrentMediaItem();
+
+            if (currentMediaItem == null) {
+                return;
+            }
+
+            MediaItem newMediaItem = currentMediaItem.buildUpon().setMediaMetadata(customMetadata).build();
+
+            // This will cause video blink/reload but won't louse progress
+            player.setMediaItem(newMediaItem, false);
+        }
+
         if (uri != null) {
             boolean isSourceEqual = uri.equals(srcUri) && cropStartMs == this.cropStartMs && cropEndMs == this.cropEndMs;
             hasDrmFailed = false;
@@ -1643,20 +1657,6 @@ public class ReactExoplayerView extends FrameLayout implements
             this.mediaDataSourceFactory =
                     DataSourceUtil.getDefaultDataSourceFactory(this.themedReactContext, bandwidthMeter,
                             this.requestHeaders);
-
-            if (this.customMetadata != customMetadata && player != null) {
-                MediaItem currentMediaItem = player.getCurrentMediaItem();
-
-                if (currentMediaItem == null) {
-                    return;
-                }
-
-                MediaItem newMediaItem = currentMediaItem.buildUpon().setMediaMetadata(customMetadata).build();
-
-                // This will cause video blink/reload but won't louse progress
-                player.setMediaItem(newMediaItem, false);
-            }
-
             this.customMetadata = customMetadata;
 
             if (!isSourceEqual) {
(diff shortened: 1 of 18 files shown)


zoriya:
Speaking for android: why do you use custom Notifications with custom actions instead of using the standard `MediaSessionCompat` and `MediaSessionConnector`? Those two classes create system's media notifications and listen to hardware volume/media buttons (like headphones or remotes) and assistant on TV.

KrzysztofMoch:
I want in feature to allow set custom actions

zoriya:
You can use custom actions with the built-in media session classes: https://developer.android.com/codelabs/supporting-mediasession#4

I believe the built-in one has a huge value, especially to run react-native-video on android tv.

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -0,0 +1,152 @@
...
+            notificationManager.createNotificationChannel(
+                NotificationChannel(
+                    NOTIFICATION_CHANEL_ID,
+                    NOTIFICATION_CHANEL_ID,
+                    NotificationManager.IMPORTANCE_LOW
+                )
+            )
+        }
+
+        val contentTitle = session.player.currentMediaItem?.mediaMetadata?.title ?: "Video with no title"
```
zoriya: I think we should be able to customize this via a prop and not take the one from the video's metadata.

KrzysztofMoch:
👀 thanks for info, will check out this

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
KrzysztofMoch: Something like prop "metadata" in source ?

zoriya: I forgot about these props, sorry. I think all of them could be used then!

freeboub: I agree, it should be useful to add a metadata prop. 
in source you already have a metadata field: https://react-native-video.github.io/react-native-video/component/props#overriding-the-metadata-of-a-source
But I am not sure the source is the good place for metadata. Metadata can change while watching live content
BTW, hardcoded string should be forbidden ! what about none english speakers :)

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -347,6 +358,12 @@ public void onHostDestroy() {
         cleanUpResources();
     }
 
+    @Override
+    protected void onDetachedFromWindow() {
```
freeboub, line 364: why did you add this ? Is this really mandatory ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -734,6 +755,55 @@ private void finishPlayerInitialization() {
...
+        if (showNotificationControls & player != null) {
+            playbackServiceConnection = new ServiceConnection() {
+                @Override
+                public void onServiceConnected(ComponentName name, IBinder service) {
+                    playbackServiceBinder = (PlaybackServiceBinder) service;
+
+                    try {
+                        playbackServiceBinder.getService().registerPlayer(player);
+                    } catch (Exception e) {
+                        DebugLog.e("ExoPlayer Playback", "Cloud not register ExoPlayer");
```
freeboub: can you use a TAG in DebugLog like other DebugLog calls please ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -734,6 +755,55 @@ private void finishPlayerInitialization() {
...
+                    }
+                }
+
+                @Override
+                public void onServiceDisconnected(ComponentName name) {
+                    playbackServiceBinder = null;
```
freeboub: can you playbackServiceBinder.getService().unregisterPlayer(player); here ?
I am afraid the service keep holding a player Reference...
Not sure my comment make sens ...

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -734,6 +755,55 @@ private void finishPlayerInitialization() {
...
+                }
+
+                @Override
+                public void onNullBinding(ComponentName name) {
+                    DebugLog.e("ExoPlayer Playback", "Cloud not register ExoPlayer");
```
freeboub: same comment on log

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: I don't think so, but is there better way to call some code when view id unmounted ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -734,6 +755,55 @@ private void finishPlayerInitialization() {
...
+            if (Build.VERSION.SDK_INT >= 29) {
+                flags = Context.BIND_AUTO_CREATE | Context.BIND_INCLUDE_CAPABILITIES;
+            } else {
+                flags = Context.BIND_AUTO_CREATE;
+            }
+
+            themedReactContext.bindService(intent, playbackServiceConnection, flags);
+        }
+
+        if(!showNotificationControls && playbackServiceBinder != null) {
```
freeboub: That is strange to setupPlaybackService in a setup function ...
maybe you should split the function in 2 different functions ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1934,6 +2012,20 @@ public void setContentStartTime(int contentStartTime) {
...
+    public void setShowNotificationControls(boolean showNotificationControls) {
+        this.showNotificationControls = showNotificationControls;
+
+        if (playbackServiceConnection == null && showNotificationControls) {
+            setupPlaybackService();
+        }
+
+        if(!showNotificationControls && playbackServiceBinder != null) {
+            if (player != null) {
+                playbackServiceBinder.getService().unregisterPlayer(player);
```
freeboub: I think you should also clean playbackServiceBinder = null here ?
see my comment in setupPlaybackService(); split

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -0,0 +1,152 @@
...
+    // Player Registry
+
+    fun registerPlayer(player: ExoPlayer) {
+        if (mediaSessionsList.containsKey(player)) {
+            return
+        }
+
+        val mediaSession = MediaSession.Builder(this, player)
+            .setId("RNVideoPlaybackService_" + player.hashCode())
+            .setCallback(VideoPlaybackCallback(10000L))
```
freeboub: maybe extract this 10 sec to a constant and add a comment indicating this is hardcoded. 10sec is OK for me

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -0,0 +1,152 @@
...
+    @SuppressLint("PrivateResource")
+    private val seekForwardBtn = CommandButton.Builder()
+        .setDisplayName("forward")
+        .setSessionCommand(commandSeekForward)
+        .setIconResId(androidx.media3.ui.R.drawable.exo_notification_fastforward)
+        .build()
+
+    @SuppressLint("PrivateResource")
+    private val seekBackwardBtn = CommandButton.Builder()
+        .setDisplayName("forward")
```
freeboub:
```suggestion
        .setDisplayName("backward")
```

examples/basic/src/VideoPlayer.tsx
```diff
@@ -819,6 +819,7 @@ class VideoPlayer extends Component {
     return (
       <TouchableOpacity style={viewStyle}>
         <Video
+          showNotificationControls={true}
```
freeboub: can you add a button to enable / disable it please ?

ios/Video/NowPlayingInfoCenterManager.swift
```diff
@@ -0,0 +1,266 @@
...
+        let commandCenter = MPRemoteCommandCenter.shared()
+
+        invalideTargets(commandCenter)
+
+        playTarget = commandCenter.playCommand.addTarget { [weak self] _ in
+            guard let self, let player = self.currentPlayer else {
+                return .commandFailed
+            }
+
+            if player.rate == 0 {
```
freeboub, line 130: That is sad to test rate to play ...
I think there is a "global issue" mixing rate and pause/playback state ...

ios/Video/NowPlayingInfoCenterManager.swift
```diff
@@ -0,0 +1,266 @@
...
+            }
+
+            return .success
+        }
+
+        skipBackwardTarget = commandCenter.skipBackwardCommand.addTarget { [weak self] _ in
+            guard let self, let player = self.currentPlayer else {
+                return .commandFailed
+            }
+            let newTime = player.currentTime() - CMTime(seconds: 10, preferredTimescale: .max)
```
freeboub: same comment than for android and the 10 sec

ios/Video/RCTVideo.swift
```diff
@@ -1057,6 +1079,17 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         }
     }
 
+    @objc
+    func setShowNotificationControls(_ showNotificationControls: Bool) {
+        if showNotificationControls == true, let player = _player {
```
freeboub: I am sure you can do cleaner code by refactoring the 2 `let player = _player` !

freeboub (changes requested):
A lot of small comments, but nothing big to change.
I am just doubting about metadata source. I have never see on my project video with metadata. I really think we should fallback on field provided by the application.
You have 2 choices: 
- either use metadata from the source (quick low impact, but not dynamic during playback, and not yet implemented on android)
- either create a new metadata field in the props to allow supporting dynamic update. 

And thank you for the PR !

KrzysztofMoch:
@zoriya seems like `MediaSessionConnector` from exoplayer is deprecated in favour of `MediaSession` from media3
CleanShot 2024-05-01 at 12 35 44

I will check if with current implementation, tv pilots will work (I will test it via adb)

zoriya:
Yes, I tried to use MediaSession from media3 in the draft PR. Is there a reason we can't use it?

KrzysztofMoch:
We don't have exoplayer2 any more as we migrated to media3

KrzysztofMoch:
I have tested and with current implementation we support key events like 126, 85 so I guess with this TV should be supported ?

zoriya:
Yes, since everything was migrated to media3 I don't see why we can't simply use: `androidx.media3:media3-session:$media3_version`

KrzysztofMoch:
We are using it
https://github.com/react-native-video/react-native-video/blob/edbb3b60aab0085ccd047f69bd04e7898a0b62a1/android/build.gradle#L226-L229

And I am using `MediaSession` from `androidx.media3:media3-session` (you can look into imports in `VideoPlaybackService`) as `MediaSessionConnector` is no longer available in media3

zoriya:
Okay, I'm sorry I was referring to outdated docs. I got the right one now.

I don't understand why you create a notification manually. media3's session support notifications out of the box: https://developer.android.com/media/media3/session/background-playback#notification

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: OK, no problem

KrzysztofMoch:
@freeboub I don't see a way to dynamically update metadata... for now I am using `player.replaceMediaItem` but this resets video. I also have tried to update notification content when setting metadata but "content" is lost when you do any action (like pause, seek)

freeboub:
> @freeboub I don't see a way to dynamically update metadata... for now I am using `player.replaceMediaItem` but this resets video. I also have tried to update notification content when setting metadata but "content" is lost when you do any action (like pause, seek)

Ok, thank you for double checking !

KrzysztofMoch:
So I guess in this case will be better to move prop `metadata` to source ? or should I leave it outside

freeboub:
> So I guess in this case will be better to move prop `metadata` to source ? or should I leave it outside

🤔 It should be in source then... As you will need it at playback startup it fully make sense to put it in the source !

ios/Video/NowPlayingInfoCenterManager.swift
KrzysztofMoch: According to Apple documentation 
CleanShot 2024-05-06 at 10 59 27
So I think that checking `rate == 0` to check if player is paused in good practice

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
KrzysztofMoch: Sure, I removed this - this now should be set by `metadata` in `source` prop

KrzysztofMoch:
> I don't understand why you create a notification manually. media3's session support notifications out of the box

When I was using those out of the box, there could be only 1 "notification" even with there was more then 1 player - with custom notification each player can have separate "notification"

cc @zoriya

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -174,6 +177,32 @@ public void setSrc(final ReactExoplayerView videoView, @Nullable ReadableMap src
...
+            String description = ReactBridgeUtils.safeGetString(propMetadata, "description");
+            String artist = ReactBridgeUtils.safeGetString(propMetadata, "artist");
+            String imageUriString = ReactBridgeUtils.safeGetString(propMetadata, "imageUri");
+
+            Uri imageUri = null;
+
+            try {
+                imageUri = Uri.parse(imageUriString);
+            } catch (Exception e) {
+                DebugLog.e("ExoPlayer Warning", "Could not parse imageUri in metadata");
```
freeboub, line 194: @KrzysztofMoch you miss this TAG ! :)

zoriya (approved):
Not sure how often one would want multiples media notifications at the same times, media controls would get weird too in those cases, but why not. LGTM (even tho I don't know much). Thanks for taking a look at this!

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
KrzysztofMoch: We don't have TAG here ...

freeboub: So create one please 😢

freeboub: OK, let say it will be for later ... (much more to handle in fact)

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -174,6 +177,32 @@ public void setSrc(final ReactExoplayerView videoView, @Nullable ReadableMap src
...
+            }
+
+            customMetadata = new MediaMetadata.Builder()
+                    .setTitle(title)
+                    .setSubtitle(subtitle)
+                    .setDescription(description)
+                    .setArtist(artist)
+                    .setArtworkUri(imageUri)
+                    .build();
+        }
```
freeboub, line 204: I think there should be a else to reset customMetadata, without it when changing content you may keep previous metadata (if no new metadata available)

docs/pages/component/props.mdx
```diff
@@ -694,20 +694,23 @@ source={{ uri: 'https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8'
 
 #### Overriding the metadata of a source
 
-<PlatformsList types={['tvOS']} />
+<PlatformsList types={['Android, iOS, tvOS']} />
 
-Provide an optional `title`, `subtitle`, `customImageUri` and/or `description` properties for the video.
```
freeboub, line 699: be carfull on tvOS, customImageUri is still used ...

freeboub, line 699: For for the renaming, can you just ping the original author of this code to inform him of the breaking change please ?

ios/Video/RCTVideo.swift
```diff
@@ -540,26 +563,30 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
         }
 
-        if let description = _source?.description {
+        if let description = _source?.customMetadata?.description {
             mapping[.commonIdentifierDescription] = description
         }
 
-        if let customImageUri = _source?.customImageUri,
-           let imageData = RCTVideoUtils.createImageMetadataItem(imageUri: customImageUri) {
+        if let customImageUri = _source?.customMetadata?.imageUri,
```
freeboub:
```suggestion
        if let imageUri = _source?.customMetadata?.imageUri,
```

ios/Video/RCTVideo.swift
```diff
@@ -540,26 +563,30 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+           let imageData = await RCTVideoUtils.createImageMetadataItem(imageUri: customImageUri) {
```
freeboub:
```suggestion
           let imageData = await RCTVideoUtils.createImageMetadataItem(imageUri: imageUri) {
```

freeboub (changes requested):
1 small comment about metadata reset changing source (can be reproduced on android with the sample)
Tested on android, this is great job 👍 
I tried on ios simulator. It was not working at all with the sample (iPhone 15 Pro Max ios 17.4).

+ I will send you small change in the sample for better testing

KrzysztofMoch:
@freeboub just to let you know - all the time I was testing on physic device (iOS)

freeboub:
@KrzysztofMoch the PR: https://github.com/TheWidlarzGroup/react-native-video/pull/5/files

freeboub:
> @freeboub just to let you know - all the time I was testing on physic device (iOS)

can you test on simulator for my curiosity please ? I don't have iPhone anymore :/

KrzysztofMoch:
@Duell10111 ping for you as I made breaking change for your code for custom metadata. I moved propreties for user metadata from `source` to `source.metadata` to make it more readable

KrzysztofMoch:
@freeboub issue with "metadata reset changing source" should be fixed now (now I am disabling notification controls for player with "no media item" (mostly invalid url/source no view etc.)

freeboub (approved):

KrzysztofMoch:
I have tested iOS simulator and it does not support "NowPlayingInfo" notification so I added to docs info that fot iOS you need to run on real device

YangJonghun:
@KrzysztofMoch
#3723 is awesome! however, prop name seems narrow in scope compared to the actual functionality.
MediaSession is not only about controlling notifications and showing metadata about the video, but it's also about synchronizing processing for external controls (airplay, car, air pod, remote controller etc..).
But it's hard to tell from the prop name if it covers this functionality.

KrzysztofMoch:
@YangJonghun you right, I had not better idea then - we can add this information to docs maybe ?

freeboub:
> @YangJonghun you right, I had not better idea then - we can add this information to docs maybe ?

I agree it is a good idea!

zoriya:
Custom notifications require editing the `AndroidManifest.xml` (as stated in the readme), would it be possible to include those changes on the repository, so expo could automatically apply them? Should I make a new issue to track that?

freeboub:
@zoriya Your request make sens, but I am 100% sure if need, can you please open a ticket or discussion to discuss about it with @KrzysztofMoch ? (or we can continue here ... ).
The issue of integrating it in manifest (move it by default to: android/src/main/AndroidManifest.xml) is that it will make the feature mandatory. 
I don't really care about service declaration (It can be move without side effects). but I worry about new permissions:
```xml
<uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
<uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
```

Today, if user doesn't want the feature and don't add it his manifest, will the app still work ?

zoriya:
That's true, it will always request for permissions. I'm not sure if media notifications is a permission that require user consent or if it is automatically granted.

I think this should be opt-out more than opt-in tho, as users, we expect to be able to control media using media buttons.

YangJonghun:
@KrzysztofMoch @freeboub @zoriya 
I think we can make people choose changeable opt-in.

some of the Player's features require permissions, services, and activities.
- service & FOREGROUND_SERVICE(permission) : download, media session, google cast ...
- activity : picture in picture ...

Currently, If someone need these features, they have to write within their AndroidManifest.xml. because Android doesn't support programmatically generate Service or Activity
But, If we write each feature as native Android dependency, people can choose and install with native flag like below
```
    if (ExoplayerDependencies["useReactNativeVideoPIP"]) {
      // For PictureInPIcture support with react-native-video
      implementation "com.react-native-video.picture-in-picture:$rnv_version"
    }
    if (ExoplayerDependencies["useReactNativeVideoMediaSession"]) {
      // For MediaSession support with react-native-video
      implementation "com.react-native-video.media-session:$rnv_version"
    }
```
(As far as I know, `react-native-notifee`'s [app.notifee.core](https://github.com/invertase/notifee/blob/main/android/src/main/AndroidManifest.xml) takes that approach, which allows it to get around some of the limitations of RN library.)

fyi) if we choose this approach, #3385 can be rewrite with Activity.

freeboub:
@YangJonghun Thank you for the clue.
We can also imagine another package react-native-video-media-session which would be responsible of the feature.
BTW, I think the real question we need to answer is should we want to remove mediaSession from the app ?! 

 jetified-media3-session-1.3.1.aar: 506K 
And same question can come for UI:
jetified-media3-ui-1.1.1.aar: 419K

I am not 100% convinced yet of this necessity :/ 
Maybe moving the tags in the manifest can be enough for now ...

YangJonghun:
@freeboub 
I don't care about the size of the package, I want the library to support users to specify only the permissions for the features they actually use (because Play Store has strict permission checks for background behavior).

There are two ways to do this: opt-in and opt-out, and currently we choose opt-out, where users specify the permissions and services they need in their code to use certain features.

My hypothesis is that react-native-video supports opt-in to improve library user's DX. If it's a regular React-Native library, not a native dependency, you can't specify the service, activity, permission, etc. required for the feature separately.

I think opt-in and opt-out are both good!👍

vargajacint:
@YangJonghun You may not but other developers are care about package size

freeboub:
As someone said and remove his comment: "You may not but other developers are care about package size"
That is not the main topic for now :)
Regarding to optin vs optout, I think you choose the good strategy to avoid changing default behavior of the package !
Regarding to the 2 new permissions, I don't think they change anything for store submission even if the feature is not used ? What's why I wonder if we can add it to the AndroidManifest from the package.

YangJonghun:
Recently, Google Play Console has been requiring proof of app content actually using Foreground Service, so if someone not using it and only have the permission added, this could be a reason for rejection.
fyi) https://developer.android.com/about/versions/14/changes/fgs-types-required#google-play-enforcement

freeboub:
@YangJonghun Thank you for the information ! So I think we should keep implementation as is !

YangJonghun:
@KrzysztofMoch 
When the `playInBackground` prop is false, I was expecting the notification control to exit as well, but currently it doesn't behave that way. Is there any plans to change it?

KrzysztofMoch:
On iOS or Android ?

YangJonghun:
@KrzysztofMoch
I tested on Android

KrzysztofMoch:
> @KrzysztofMoch When the `playInBackground` prop is false, I was expecting the notification control to exit as well, but currently it doesn't behave that way. Is there any plans to change it?

I think it makes sense - will try to care of it next week

Issue #3704, opened 2024-04-23
[BUG]: Error while updating 'SelectedTextTrack' 


rahulnainwalttn:
### Version

v6 (Beta)

### What platforms are you having the problem on?

Android

### Architecture

Old architecture

### What happened?

My text tracks looks like below JSON. Initially, index 1 is selected. when I am selecting 2 or 3 indexes its working fine but as soon i am selecting index 0 app is getting crashed in **[6.0.0-rc.0](https://github.com/react-native-video/react-native-video/compare/v6.0.0-beta.8...v6.0.0-rc.0)**

Stream URL - https://bitmovin-a.akamaihd.net/content/sintel/hls/playlist.m3u8

"textTracks": [{"index": 0, "language": "de", "selected": false, "title": "Deutsch", "type": "text/vtt"}, {"index": 1, "language": "en", "selected": true, "title": "English", "type": "text/vtt"}, {"index": 2, "language": "es", "selected": false, "title": "Espanol", "type": "text/vtt"}, {"index": 3, "language": "fr", "selected": false, "title": "Français", "type": "text/vtt"}]

![Image_20240423_171324_526](https://github.com/react-native-video/react-native-video/assets/136161024/ae9f6c66-9168-4ebc-96a2-7e617f2036fd)

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are:

rahulnainwalttn:
Logs - 

1 - 
```
Error while updating prop selectedTextTrack
                                                                                                    java.lang.reflect.InvocationTargetException
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp(ViewManagersPropertyCache.java:93)
                                                                                                    	at com.facebook.react.uimanager.ViewManagerPropertyUpdater$FallbackViewManagerSetter.setProperty(ViewManagerPropertyUpdater.java:136)
                                                                                                    	at com.facebook.react.uimanager.ViewManagerPropertyUpdater.updateProps(ViewManagerPropertyUpdater.java:56)
                                                                                                    	at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:48)
                                                                                                    	at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:142)
[...]

PR #3728, opened 2024-05-02
Fix/text track selection by index ios

freeboub:
## Summary
Apply type change (done for codegen) on selected track by index on ios.

### Motivation
Make subtitles working ! see https://github.com/react-native-video/react-native-video/issues/3704

### Changes
- Fix selected track type (always string now)
- Update sample to be able to retest selection by index easilly

## Test plan
- can be tested with sample
- BTW, selection by index doesn't work on sideLoaded text tracks


KrzysztofMoch (approved):
LGTM

PR #3735, opened 2024-05-05
perf: ensure we do not provide callback to native if no callback provided from app

freeboub:
## Summary
As discussed with @KrzysztofMoch a possible optimization is to avoid giving callback functions (starting with 'on*') to native if application didn't provide it 

### Motivation
performances improvements.
Additionnally, we must keep the useCallback as paramters of the function are differents between native & applicative, for exemple:
```TS
    const _onTimedMetadata = useCallback(
      (e: NativeSyntheticEvent<OnTimedMetadataData>) => {
        onTimedMetadata?.(e.nativeEvent);
      },
      [onTimedMetadata],
    );
````
send e.nativeEvent to application

### Changes
Just add ternary to check to avoid giving the callback to native

## Test plan
Basic tests done on android/ios with the sample

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -542,38 +542,60 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           selectedAudioTrack={_selectedAudioTrack}
           selectedVideoTrack={_selectedVideoTrack}
           onGetLicense={useExternalGetLicense ? onGetLicense : undefined}
-          onVideoLoad={onVideoLoad as (e: NativeSyntheticEvent<object>) => void}
-          onVideoLoadStart={onVideoLoadStart}
-          onVideoError={onVideoError}
-          onVideoProgress={onVideoProgress}
-          onVideoSeek={onVideoSeek}
+          onVideoLoad={
+            onLoad
+              ? (onVideoLoad as (e: NativeSyntheticEvent<object>) => void)
+              : undefined
+          }
+          onVideoLoadStart={onLoadStart ? onVideoLoadStart : undefined}
+          onVideoError={onError ? onVideoError : undefined}
+          onVideoProgress={onProgress ? onVideoProgress : undefined}
+          onVideoSeek={onSeek ? onVideoSeek : undefined}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by KrzysztofMoch** [comment]
I also think we should do "early returns" - because we don't need to load tracks data if user don't revive it

~~I will send you patch for iOS later~~
I will make separate PR for this

File: src/Video.tsx [hunk 1]
```diff
@@ -542,38 +538,60 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
-          onVideoLoad={onVideoLoad as (e: NativeSyntheticEvent<object>) => void}
-          onVideoLoadStart={onVideoLoadStart}
-          onVideoError={onVideoError}
-          onVideoProgress={onVideoProgress}
-          onVideoSeek={onVideoSeek}
+          onVideoLoad={
+            onLoad
+              ? (onVideoLoad as (e: NativeSyntheticEvent<object>) => void)
+              : undefined
+          }
```
KrzysztofMoch: We want to also have this callback when `hasPoster` is `true`

src/Video.tsx
```diff
@@ -542,38 +538,60 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+          onVideoLoadStart={onLoadStart ? onVideoLoadStart : undefined}
```
KrzysztofMoch: Same - We want to also have this callback when `hasPoster` is true

KrzysztofMoch (changes requested):
We need to add "or" condition to places that "poster" logic depends on

like 
```jsx
onVideoLoad={
  onLoad || hasPoster
    ? (onVideoLoad as (e: NativeSyntheticEvent<object>) => void)
    : undefined
}
```

freeboub:
Good catch agin, thank you !
I integrate some of the swift fix you proposed to me also ! (but not all)

KrzysztofMoch (approved):
Nice let's merge this - I will make PR for "early returns" tomorrow

PR #3741, opened 2024-05-07
perf(ios): add early returns

KrzysztofMoch:
## Summary
Add early returns - this will allow to skip code that was called only for callbacks (that are not defied/used by user)

On android we can do it as there we are sending events and we don't know if callback is defined or not

## Test plan
- [x] Tested in example app
- [x] CI pass

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -1355,7 +1355,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
                 self._startPosition = -1
             }
 
-            if onVideoLoad != nil {
+            if onVideoLoad != nil, self._videoLoadStarted {
                 var duration = Float(CMTimeGetSeconds(_playerItem.asset.duration))
 
                 if duration.isNaN || duration == 0 {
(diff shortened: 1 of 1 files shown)


ios/Video/RCTVideo.swift
```diff
@@ -307,6 +307,8 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     // MARK: - Progress
 
     func sendProgressUpdate() {
+        guard onVideoProgress != nil else { return }
+
```
freeboub, line 314: I think you will introduce a regression linked to ads,
please see next lines: 

```swift
            #if USE_GOOGLE_IMA
                if !_didRequestAds && currentTimeSecs >= 0.0001 && _adTagUrl != nil {
                    _imaAdsManager.requestAds()
                    _didRequestAds = true
                }
            #endif
```

ios/Video/RCTVideo.swift
```diff
@@ -1463,7 +1472,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     }
 
     func handleVolumeChange(player: AVPlayer, change: NSKeyValueObservedChange<Float>) {
-        guard let _player else { return }
+        guard let _player, onVolumeChange != nil else { return }
```
freeboub, line 1473: not sure this is a real optmi, but OK

ios/Video/RCTVideo.swift
```diff
@@ -1475,7 +1484,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     func handleExternalPlaybackActiveChange(player _: AVPlayer, change _: NSKeyValueObservedChange<Bool>) {
         #if !os(visionOS)
-            guard let _player else { return }
+            guard let _player, onVideoExternalPlaybackChange != nil else { return }
```
freeboub, line 1485: also not usefull here ...

ios/Video/RCTVideo.swift
```diff
@@ -1331,73 +1334,79 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+                "time": NSNumber(value: self._pendingSeekTime),
+                "tolerance": NSNumber(value: 100),
+            ])
+            self._pendingSeek = false
         }
 
-        var width: Float = 0
-        var height: Float = 0
-        var orientation = "undefined"
+        if self._startPosition >= 0 {
```
freeboub: these setSeek where in a Task, Are your sure of this move ?

ios/Video/RCTVideo.swift
```diff
@@ -1331,73 +1334,79 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+                self._videoLoadStarted = false
             }
-
+        } else {
             self._videoLoadStarted = false
-            self._playerObserver.attachPlayerEventListeners()
-            self.applyModifiers()
         }
+
+        self._playerObserver.attachPlayerEventListeners()
```
freeboub: Is it really expected to move these events outside of the Task ?

freeboub (changes requested):

ios/Video/RCTVideo.swift
KrzysztofMoch: Good catch! I have made it to skip "onVideoProgress != nil" when using ads as we need this code to run in this situation
```swift
#if !USE_GOOGLE_IMA
  guard onVideoProgress != nil else { return }
#endif
```

ios/Video/RCTVideo.swift
KrzysztofMoch: I think that this can be faster as we don't access properties but not sure

ios/Video/RCTVideo.swift
KrzysztofMoch: Yup, I moved this whole code into Task

ios/Video/RCTVideo.swift
KrzysztofMoch: I moved whole code into Task

ios/Video/RCTVideo.swift
freeboub: @KrzysztofMoch In fact a better optimization would be to avoid registering the callback in fact !

ios/Video/RCTVideo.swift
```diff
@@ -1373,25 +1355,52 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+                if let videoTrack = tracks?.first {
+                    width = Float(videoTrack.naturalSize.width)
+                    height = Float(videoTrack.naturalSize.height)
+                } else if _playerItem.presentationSize.height != 0.0 {
+                    width = Float(_playerItem.presentationSize.width)
+                    height = Float(_playerItem.presentationSize.height)
+                }
+                orientation = width > height ? "landscape" : width == height ? "square" : "portrait"
+
+                if self._videoLoadStarted {
```
freeboub: don't you think we should merge 
`if self._videoLoadStarted {`
and 
`if onVideoLoad != nil {`

will will avoid getTracks and ...
?

freeboub (approved):

ios/Video/RCTVideo.swift
KrzysztofMoch: You are right!

ios/Video/RCTVideo.swift
KrzysztofMoch: Yes, I will do it in separate PR

PR #3746, opened 2024-05-08
fix(android): video flickering add playback start

### LINKED ISSUE #3740: [BUG]: Video first frame do flickering when player load video and start playing, Demo Attached
### Version

react-native-video: ^6.0.0-rc.0

react-native: 0.71.14

Video URL: HLS m3u8

Android

### Architecture

Old architecture

### What happened?

I have a TikTok-like app where, on the home screen, when we try to scroll the video, there is a flickering issue. While this problem was not present in the previous version (v5), it has appeared now. I have recorded a demo video demonstrating this issue, and it is attached

### Reproduction

repository link

### Reproduction

Steps to reproduce this bug are:  Swipable videos like TikTok, and swiping the video, a flicker will appear. like when we swipe to video, the first frame is small having black spaces, then suddenly remove black spaces, goes to full mode

freeboub:
## Summary
Avoid flickering when video playback starts on android

### Motivation
Fix: https://github.com/TheWidlarzGroup/react-native-video/issues/3740
This issue gives a very bad quality feedback...

### Changes
I fix few warning and did few clean up but the main change is to delay the shutterView removal by 15 ms

## Test plan
Can be tested in the sample


KrzysztofMoch (approved):
I don't like that we need to use such workarounds... 
LGTM  ✅

freeboub:
Yes, but the issue is quite complicated, 
We have a main view to display the video, and a view to adapt to the aspect ratio of the video.
At playback start we use the main view as the video size. 
One playback is started we know the expected aspect ratio of the video and then resize the view to adapt to expected resize mode.
We cannot do this resizing before knowing the real aspect ratio which will be played.
The best fix would be to provide target aspect ratio of the video in the source, then we would be able to adapt view size with target video size. The issue to implement that is mainly that when you start a playback you don't necessary know which will be the video aspect ratio, So this solution may not apply to all videos / backend implementation ...

The delay is only 15ms, so 1 additional frame for user perspective, I think this will be seemless, so it is acceptable.

KrzysztofMoch:
I want to make release, so I will merge this

PR #3747, opened 2024-05-08
Chore/improve issue template

freeboub:
## Summary
Improve issue template

### Motivation
Avoid bugs on outdated version


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -7,6 +7,8 @@ body:
   - type: markdown
     attributes:
       value: Thanks for taking the time to fill out this bug report!
+      Please do not report issue on 5.2.1 version, this version is not maintained anymore.
+      Only issues on version > V6 will be handled. Please also ensure your issue is reproduced with the last release!
 
   - type: textarea
```
KrzysztofMoch, line 13: It's less confusing to use smaller input for version
```suggestion
  - type: input
```

KrzysztofMoch (approved):

.github/ISSUE_TEMPLATE/bug-report.yml
freeboub: > It's less confusing to use smaller input for version
> ```suggestion
>   - type: input
> ```

I agree but I don't think this is a good way to proceed :) putting an input will make necessary to patch this file to include each new release! You put input in the past but it leads to '6.0.0-beta' input, as there were 10 beta, these tickets are confusing now. A free field allows you to put the exact version

KrzysztofMoch: I agree, but input is smaller textarea (also allow for free text)

PR #3750, opened 2024-05-10
fix(ios): call `onLoadStart` earlier

### LINKED ISSUE #3744: onBuffer & onLoadStart not working properly
### Version

react-native-video: 6.0.0-rc.0

iOS

### What happened?

In android everything working fine, But in IOS onBuffer not working & onLoadStart fired with onLoad

KrzysztofMoch:
## Summary
call callback `onLoadStart` earlier to make it work properly 

### Motivation
fix #3744

### Changes
- move callback `onLoadStart` earlier
- don't call `onVideoBuffer` in `handlePlaybackLikelyToKeepUp` if value is same as previous

## Test plan
- [x] Tested Locally
- [x] CI pass

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -375,6 +375,17 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
             throw NSError(domain: "", code: 0, userInfo: nil)
         }
 
+        // Perform on next run loop, otherwise onVideoLoadStart is nil
+        onVideoLoadStart?([
+            "src": [
+                "uri": _source?.uri ?? NSNull(),
+                "type": _source?.type ?? NSNull(),
+                "isNetwork": NSNumber(value: _source?.isNetwork ?? false),
+            ],
+            "drm": _drm?.json ?? NSNull(),
+            "target": reactTag,
+        ])
+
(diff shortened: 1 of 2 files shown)


ios/Video/RCTVideo.swift
```diff
@@ -1434,7 +1435,11 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     }
 
     // Continue playing (or not if paused) after being paused due to hitting an unbuffered zone.
-    func handlePlaybackLikelyToKeepUp(playerItem _: AVPlayerItem, change _: NSKeyValueObservedChange<Bool>) {
```
freeboub, line 1437: I am not sure this will be enough as there are 2 other calls in handleReadyForDisplay and handlePlaybackBufferKeyEmpty.
I think the buffering state shall handled with a local variable to know what is the oldState reported to the UI, and not the old handlePlaybackLikelyToKeepUp

freeboub:
I think this change may also fix: https://github.com/TheWidlarzGroup/react-native-video/issues/3675

ios/Video/RCTVideo.swift
KrzysztofMoch: Grate Idea! this is working perfect now

freeboub (approved):

PR #3751, opened 2024-05-10
Fix(android): avoid video resizing flickering

freeboub:
## Summary
resize video View when we receive the list of video tracks instead of resizing it when the video is displayed

### Motivation
Fix video flickering when starting playback:  https://github.com/TheWidlarzGroup/react-native-video/issues/3740 and https://github.com/TheWidlarzGroup/react-native-video/issues/3653

### Changes
resize video View when we receive the list of video tracks instead of resizing it when the video is displayed

## Test plan
can be tested with the sample app


YangJonghun:
It is still possible to reproduce flickering.
Changing the resolution via selectedVideoTrack with dash source with multiple heights can reproduce flicker.

freeboub:
> It is still possible to reproduce flickering. Changing the resolution via selectedVideoTrack with dash source with multiple heights can reproduce flicker.

Thank you for the feedback, I didn't focus on selectedVideoTrack, but it is my plan soon ! :)

KrzysztofMoch:
Lets merge, @freeboub I will leave you fixing selectedVideoTrack case

abbasmoosavi:
Yes, it was fixed in v6.0.0-rc.2.

PR #3754, opened 2024-05-10
chore(android): refactor side loaded text tracks

freeboub:
## Summary
split side loaded text tracks parsing and handling by the player implementation

### Motivation
Have a code which is more modular.

### Changes
Add an intermediate classes to handle externally loaded subtitles tracks.
⚠️ Before the change, if no language was provided a fallback "language " + index was used. I think this is a bad behavior from the player because there is no translation support and index can quiet random.

## Test plan
Tested with the sample


KrzysztofMoch (approved):
LGTM

PR #3759, opened 2024-05-10
fix(avoid): avoid early return in setSrc

freeboub:
## Summary
Fix issue found in sample
Use case:
- launch sample on android
- channel up until invalid url
- channel down
- Issue: video doesn't restart

### Changes
Continue handling setSrc even is player is not allocated

### Motivation
@KrzysztofMoch This is more a PR to discuss (even if it really fix the issue).
Don't you think this part of code should be called after reloadSource like in initializePlayerCode : https://github.com/TheWidlarzGroup/react-native-video/blob/f87a7938c6941c69915477fcf5f08c54a2635597/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java#L628


KrzysztofMoch (approved):
omg, this surly should be like this ... good catch!

KrzysztofMoch:
> Don't you think this part of code should be called after reloadSource like in initializePlayerCode :

I guess you are right, maybe we can move code that is setting metadata to function to don't duplicate code

KrzysztofMoch:
I will merge this as it fix problem, will look into moving this to function and call it after reloadSource

PR #3767, opened 2024-05-13
Chore/video tracks crash and clean

freeboub:
## Summary
Fix crash when app provide a string instead of index in selectedTrack management
+ 2 minor clean up

### Motivation
avoid crash and clean code

### Changes
handle exception in case of invalid type


freeboub:
This is a preliminary PR, a bigger one will come soon !

KrzysztofMoch (approved):

PR #3783, opened 2024-05-16
update IMA SDK

philipheinser:
We need the new version of the IMA SDK and I checked and it should be compatible.


freeboub (approved):

KrzysztofMoch (approved):
Double checked, should work fine

PR #3784, opened 2024-05-16
fix(android): playback doesn't work with 0 startPositionMs

YangJonghun:
## Summary
Some device(maybe low performance device) doesn't work with zero startPositionMs and DRM
(playback doesn't work before user trigger seek)
So I use resetPosition = true when startPositionMs is 0

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -745,11 +745,11 @@ public class ReactExoplayerView extends FrameLayout implements
         boolean haveResumePosition = resumeWindow != C.INDEX_UNSET;
         if (haveResumePosition) {
             player.seekTo(resumeWindow, resumePosition);
-        }
-        if (startPositionMs >= 0) {
+            player.setMediaSource(mediaSource, false);
+        } else if (startPositionMs > 0) {
             player.setMediaSource(mediaSource, startPositionMs);
         } else {
-            player.setMediaSource(mediaSource, !haveResumePosition);
+            player.setMediaSource(mediaSource, true);
         }
         player.prepare();
         playerNeedsSource = false;


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -745,11 +745,13 @@ private void initializePlayerSource(ReactExoplayerView self, DrmSessionManager d
...
-        }
-        if (startPositionMs >= 0) {
+            player.setMediaSource(mediaSource, false);
+        } else if (startPositionMs == 0) {
+            player.setMediaSource(mediaSource, true);
+        } else if (startPositionMs > 0) {
             player.setMediaSource(mediaSource, startPositionMs);
         } else {
-            player.setMediaSource(mediaSource, !haveResumePosition);
+            player.setMediaSource(mediaSource, false);
```
freeboub: correct me if I am wrong, but the last else case should be:
player.setMediaSource(mediaSource, true );
?

```suggestion
            player.setMediaSource(mediaSource, true);
```

YangJonghun: This is the part that has changed.
If `resetPosition` is false, it use `getCurrentPosition()` internally.
I felt this behavior was more appropriate, so I fixed it.

YangJonghun: If you think it's better to guarantee the same behavior as before, I'll fix it.

freeboub: I was thinking the aim of this fix is to add the case with ` if (startPositionMs == 0) {` ?
because here it is the normal use case ?!

YangJonghun: I think I'm wrong and it would be better to just fix the bug case. Thanks :)

freeboub (approved):

KrzysztofMoch (approved):

ZakirBangash:
@freeboub @YangJonghun  I have tiktok like app , where we scroll the videos, on android 9 S8 Samsung, some videos get stuck, now in my client phone two videos constantly stuck. while others are working, I just want to know whether this PR resolved this issue?

freeboub:
@ZakirBangash let's try on 6.2 , will be released soon

ZakirBangash:
@freeboub Yes Please 🙏, We are in production App Nooberly,  client reported this behaviour, Would be great help if we have this issue fixed.

PR #3786, opened 2024-05-16
fix(android): remaining ad view when zapping

freeboub:
## Summary
release adsLoader if unnecessary

### Motivation
Can be reproduced on sample: 
play content with ads, during ad playback, change content
-> an ad overlay is still displayed over the video

### Changes
Release the adsLoader if no adTagUrl is provided

## Test plan
can be tested in the sample


KrzysztofMoch (approved):
LGTM ✅

PR #3794, opened 2024-05-19
feat: add setVolume function to component's ref

seyedmostafahasani:
## Summary
Added setVolume function to component's ref.

### Motivation
Handle volume with ref.

### Changes
Added a new function to component's ref.

## Test plan

### DIFF
--- a/docs/pages/component/methods.mdx
+++ b/docs/pages/component/methods.mdx
@@ -98,7 +98,7 @@ tolerance is the max distance in milliseconds from the seconds position that's a
 
 `setVolume(value): Promise<void>`
 
-The function to adjust the audio level of the video to a desired level.
+This function will change the volume exactly like [volume](../props#volume) property. default value and range are the same then.
 
 ### Example Usage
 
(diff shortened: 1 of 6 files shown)


docs/pages/component/methods.mdx
```diff
@@ -92,6 +92,14 @@ By default iOS seeks within 100 milliseconds of the target position. If you need
 
 tolerance is the max distance in milliseconds from the seconds position that's allowed. Using a more exact tolerance can cause seeks to take longer. If you want to seek exactly, set tolerance to 0.
 
+### `setVolume`
+
+<PlatformsList types={['Android', 'iOS']} />
+
+`setVolume(value): Promise<void>`
+
```
freeboub, line 100: Can you add a link to volume prop here please ?
Something like:
```suggestion

This function will change the volume exactly like [volume](../props#volume) property. default value and range are the same then.
```
Notice I am not sure of the link format ... to be double check in documentation of documentation :)

freeboub (approved):

KrzysztofMoch (approved):
LGTM!

PR #3813, opened 2024-05-22
feat(android): change default user agent value

YangJonghun:
## Summary
feat(android): change default user agent value

### Motivation
change default user agent value

### Changes
current default user agent generated from [Util.getUserAgent](https://github.com/androidx/media/blob/1.1.1/libraries/common/src/main/java/androidx/media3/common/util/Util.java#L1774-L1790) method
react-native-video prefixes the user-agent with ReactNativeVideo, but that's not very helpful. I think it would be more helpful to prefix the package name instead.

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/DataSourceUtil.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/DataSourceUtil.java
@@ -38,7 +38,7 @@ public class DataSourceUtil {
 
     public static String getUserAgent(ReactContext context) {
         if (userAgent == null) {
-            userAgent = Util.getUserAgent(context, "ReactNativeVideo");
+            userAgent = Util.getUserAgent(context, context.getPackageName());
         }
         return userAgent;
     }
--- a/docs/pages/component/props.mdx
+++ b/docs/pages/component/props.mdx
@@ -290,7 +290,7 @@ If a preferred [fullscreenOrientation](#fullscreenorientation) is set, causes th
 
 ### `headers`
 
-<PlatformsList types={['Android']} />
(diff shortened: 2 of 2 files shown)


KrzysztofMoch (approved):
Make sense for me

freeboub:
It fully makes sense, but I would prefer a new prop to configure it.
I see 2 reasons
- Maybe some user already use this value to track analytics server side (we should not disturb them
- some user may also want to change different userAgent depending on mobile used. I had the case in the past, we were changing user agent depending on phone capabilities (hevc support and widevine level support)

freeboub (changes requested):
Already done :)

YangJonghun:
@freeboub 
> Maybe some user already use this value to track analytics server side (we should not disturb them
- I agree, but we already provide `source.headers` prop. and it can be rewrite user-agent (if someone use analytics on serverside, then he can change it). also, current default user agent doesn't make it easy for clients to tell which package the request is from.

> some user may also want to change different userAgent depending on mobile used. I had the case in the past, we were changing user agent depending on phone capabilities (hevc support and widevine level support)
- Currently default UA does not provide any relevant information: the change in this PR is the header of the source, not the header of the drm (which is provided to the lisence server). If someone need that information, it seems like a different header or CMCD or something would be appropriate.

KrzysztofMoch:
I agree with @YangJonghun - we already provide option to set it and setting it to app package name make more sense then current one
I will highlight this change in the release notes

@freeboub I will merge this one, hope you wont be mad 😄

freeboub:
I don't go mad for that :) I just warn of potential issues, And I agree with you this change can only be better :D

PR #3819, opened 2024-05-22
fix(ios): fix playback status with lifecycle

### LINKED ISSUE #3790: [Feature]: detect airplay(or mirroring) playback within iOS native side
### Description

detect airplay(or mirroring) playback within native side and need to check within native lifecycle (background, inactive, etc ...)

### Why it is needed ?

sometimes user need to play via airplay or eternal playback
but current native codes are ignore it.
and it makes bug when `playInBackground`, `playWhenInactive` props are false

### Possible implementation

_No response_

### Code sample

_No response_

YangJonghun:
## Summary
fix(ios): fix playback status with lifecycle

### Motivation
- Modify it to ignore playInBackground because people can use AppleTV, airplay, etc.

### Changes
- check external playback active state with iOS lifecycle
- check playback status via "_isPlaying" instead of "_paused"
  - "_paused" is not exactly synchronized with the player's current state. this change is important because the introduction of the "showNotificationControls" prop makes it possible for external controls to control the video.
  
- resolve #3790

## Memo
Since the "showNotificationControls" prop allows people to control the video with external controls, I think we need to clarify the behavior of props like playInBackground, playWhenInactive. Currently, we can only control the entry point of lifecycle, and it can be played at any time while not in the foreground.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -1498,7 +1498,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     func handleExternalPlaybackActiveChange(player _: AVPlayer, change _: NSKeyValueObservedChange<Bool>) {
         #if !os(visionOS)
             guard let _player else { return }
-            if UIApplication.shared.applicationState == .background {
+            if !_playInBackground && UIApplication.shared.applicationState == .background {
                 _playerLayer?.player = nil
                 _playerViewController?.player = nil
             }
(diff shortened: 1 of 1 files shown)


freeboub:
fully agree with your memo ...

freeboub (approved):

freeboub:
@YangJonghun will it also fix https://github.com/TheWidlarzGroup/react-native-video/issues/3770 ?

YangJonghun:
@freeboub
I don't think so, that issue is about audio control. this PR simply changes state of the player.

KrzysztofMoch (approved):

KrzysztofMoch:
@YangJonghun After this change when we set playInBackground to `true` when app going background, playback is pausing even if there is no external playback active

YangJonghun:
@KrzysztofMoch 
I'll check it out. Thanks for letting me know:)

YangJonghun:
@KrzysztofMoch 
The modified code works as intended, and the issue is reproduced in commits before this PR was merged.
FYI) Commit before this PR was merged - https://github.com/TheWidlarzGroup/react-native-video#d072aeb451f515b84b821367bfa3c12cfbeeb725
Probably other code is the cause.

PR #3824, opened 2024-05-23
feat: add getCurrentPosition to component's ref

seyedmostafahasani:
## Summary
This function retrieves and returns the precise current position of the video playback, measured in seconds.

### Motivation
The goal is to provide users with the ability to access the current position of the video playback. This allows for the creation of custom controls like forward and rewind buttons, enhancing user experience.

### Changes
A new function `getCurrentPosition` has been added to the component's ref.

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -124,6 +124,7 @@ import com.brentvatne.react.R;
 import com.brentvatne.receiver.AudioBecomingNoisyReceiver;
 import com.brentvatne.receiver.BecomingNoisyListener;
 import com.facebook.react.bridge.LifecycleEventListener;
+import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.UiThreadUtil;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.google.ads.interactivemedia.v3.api.AdError;
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -688,8 +688,8 @@ public class ReactExoplayerView extends FrameLayout implements
 
     public void getCurrentPosition(Promise promise) {
         if (player != null) {
-            long currentPosition = player.getCurrentPosition() / 1000;
-            promise.resolve((double) currentPosition);
+            double currentPosition = player.getCurrentPosition() / 1000;
+            promise.resolve(currentPosition);
         } else {
             promise.reject("PLAYER_NOT_AVAILABLE", "Player is not initialized.");
         }
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -685,6 +685,15 @@ public class ReactExoplayerView extends FrameLayout implements
         mainHandler.postDelayed(mainRunnable, 1);
     }
 
+    public void getCurrentPosition(Promise promise) {
+        if (player != null) {
+            long currentPosition = player.getCurrentPosition() / 1000;
+            promise.resolve((double) currentPosition);
+        } else {
+            promise.reject("PLAYER_NOT_AVAILABLE", "Player is not initialized.");
+        }
+    }
+
     private void initializePlayerCore(ReactExoplayerView self) {
         ExoTrackSelection.Factory videoTrackSelectionFactory = new AdaptiveTrackSelection.Factory();
         self.trackSelector = new DefaultTrackSelector(getContext(), videoTrackSelectionFactory);
--- a/android/src/main/java/com/brentvatne/react/VideoManagerModule.kt
+++ b/android/src/main/java/com/brentvatne/react/VideoManagerModule.kt
@@ -65,7 +65,7 @@ class VideoManagerModule(reactContext: ReactApplicationContext?) : ReactContextB
     @ReactMethod
     fun getCurrentPosition(reactTag: Int, promise: Promise) {
         performOnPlayerView(reactTag) {
-           it?.getCurrentPosition(promise)
+            it?.getCurrentPosition(promise)
         }
     }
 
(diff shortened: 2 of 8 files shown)


freeboub:
Looks good to me, Just one note, I would prefer having the time in msec instead of sec. @KrzysztofMoch @YangJonghun what do you think of this point ?

android/src/main/java/com/brentvatne/react/VideoManagerModule.kt
```diff
@@ -61,6 +62,13 @@ class VideoManagerModule(reactContext: ReactApplicationContext?) : ReactContextB
         }
     }
 
+    @ReactMethod
+    fun getCurrentPosition(reactTag: Int, promise: Promise) {
+        performOnPlayerView(reactTag) {
+           it?.getCurrentPosition(promise)
```
freeboub:
```suggestion
            it?.getCurrentPosition(promise)
```

YangJonghun:
As far as I know, `onProgress` callback exists for the same reason. I wonder for what purpose this method was implemented.

seyedmostafahasani:
@YangJonghun  you're correct! In situations where the user only needs to get the current position when interacting with custom controls, such as the rewind and forward buttons, it's not necessary to use a callback function that is called every progress update interval.

Instead, the user can use this approach that is only called when the user clicks the rewind or forward buttons. When a button is clicked, you can get the current position and then seek forward or backward based on the button that was clicked.

YangJonghun:
@seyedmostafahasani 
That makes sense!
However, I think it might be a bit difficult to use it for precise work due to the limitations of old architecture, since the current time value keeps changing on playback.
It will be great later when we're done with the new architecture :)
Thanks for your explanation.👍

seyedmostafahasani:
So now what is your opinion about the response of this function, second or millisecond?

YangJonghun:
@seyedmostafahasani 
Oh sorry to late response.
I like second(specifically, second as a float value).
I prefer the same format as the value returned by `onProgress`, and I think many developers will be familiar with the web's video.currentTime format

seyedmostafahasani:
I completely agree with you about second.

freeboub:
As C developer I prefer working with integer, but if you are ok for sec, then OK for me !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -685,6 +686,15 @@ private void initializePlayer() {
         mainHandler.postDelayed(mainRunnable, 1);
     }
 
+    public void getCurrentPosition(Promise promise) {
+        if (player != null) {
+            long currentPosition = player.getCurrentPosition() / 1000;
+            promise.resolve((double) currentPosition);
```
freeboub: Directly use double here, if you cast to long, your will loose msec information !

```suggestion
            double currentPosition = player.getCurrentPosition() / 1000;
            promise.resolve(currentPosition);
```

freeboub (changes requested):
See my comment for android

freeboub (approved):

docs/pages/component/methods.mdx
```diff
@@ -100,6 +100,14 @@ tolerance is the max distance in milliseconds from the seconds position that's a
 
 This function will change the volume exactly like [volume](./props#volume) property. default value and range are the same then.
 
+### `getCurrentPosition`
+
+<PlatformsList types={['Android', 'iOS']} />
+
+`getCurrentPosition(): Promise<number>`
+
+This function retrieves and returns the precise current position of the video playback, measured in seconds.
```
freeboub, line 109:
```suggestion
This function retrieves and returns the precise current position of the video playback, measured in seconds.
This function will throw an error if player is not initialized.
```

ios/Video/RCTVideo.swift
```diff
@@ -1620,6 +1620,12 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         }
     }
 
+    @objc
+    func getCurrentPlaybackTime(_ resolve: @escaping RCTPromiseResolveBlock) {
+        let currentTime = RCTVideoUtils.getCurrentTime(playerItem: _playerItem)
+        resolve(currentTime)
```
freeboub: it would be great to match android behavior: if playerItem is nil, reject the promise.

freeboub (changes requested):

freeboub (approved):

freeboub (approved):

YangJonghun:
@freeboub @KrzysztofMoch @seyedmostafahasani 
I think we need to deprecate this feature
ref command with return value not supported by Fabric
(`save` command also have same problem)

seyedmostafahasani:
@YangJonghun I think we can use them with Fabric, but an important point is that these are deprecated methods on the native or JS side. Utilizing deprecated methods on the native or JS side may result in these functions not working correctly. I recommend checking out `react-native-blurhash`, as it employs Fabric. This library uses functions that return values via promises, such as the `encode` function, which you can observe in these URLs.
https://github.com/mrousavy/react-native-blurhash/blob/master/src/index.tsx
https://github.com/mrousavy/react-native-blurhash/blob/master/android/src/main/java/com/mrousavy/blurhash/BlurhashModule.kt

YangJonghun:
@seyedmostafahasani 
BlurHash is NativeModule.
TurboModule and Fabric are different. TurboModule can be return values
What I meant was that it's not possible to use a View as a ref and return it.

Currently we use `save` or `getCurrentPosition` in react-native-video is to reference a reactTag in NativeModule(VideoManager).
We are using `findNodeHandle` for above behavior, but findNodeHandle will be deprecated, so it is not recommended to continue using this method. We need to migrate to command method to replace it.
However command does not support return or callback functions.

seyedmostafahasani:
@YangJonghun Oh yes, you're right. Thanks for the detailed explanation. I was referring to the deprecated method `findNodeHandle`. Could you please take a look at the `react-native-view-shot` library? I think this library could help us.

YangJonghun:
@seyedmostafahasani 
Thanks for give me reference
But unfortunately react-native-view-shot also using `findNodeHandle` now. ([link](https://github.com/gre/react-native-view-shot/blob/master/src/index.js#L128))🥲

I guess I'll have to keep it for now and find a tricky way around it. Because we don't know when RN team get rid of findNodeHandle.
fyi) https://github.com/reactwg/react-native-new-architecture/blob/main/docs/enable-libraries-prerequisites.md

seyedmostafahasani:
@YangJonghun anytime!
Did you see this [PR](https://github.com/gre/react-native-view-shot/pull/516)?

I agree with you, we can keep it for now. Also, if I find a better solution, I will share it with you.

PR #3830, opened 2024-05-23
feat(android): Change subtitleLayout from child to sibling of layout

YangJonghun:
## Summary
Change subtitleLayout from child to sibling of layout

### Motivation
I think the behavior of subtitle position changing by resize mode is a bug, but if not, please close it.

### Changes
- as-is

- to-be

## Test plan
change resize mode (CONTAIN / COVER)

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -86,10 +86,10 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
         adOverlayFrameLayout = new FrameLayout(context);
 
         layout.addView(shutterView, 1, layoutParams);
-        layout.addView(subtitleLayout, 2, layoutParams);
-        layout.addView(adOverlayFrameLayout, 3, layoutParams);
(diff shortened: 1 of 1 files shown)


freeboub:
@YangJonghun I think you will fix this issue: https://github.com/TheWidlarzGroup/react-native-video/issues/3662
Can you please confirm, thank you !

freeboub (approved):

YangJonghun:
@freeboub
Maybe..? I think what he really wanted was a custom style

freeboub:
> @freeboub Maybe..? I think what he really wanted was a custom style

Not sure, let's ask in the ticket !

freeboub:
@YangJonghun This patch cause following regression: https://github.com/TheWidlarzGroup/react-native-video/issues/4019
I am not sure on how to fix it ...
Do you think we should add a prop:

subtitleFollowVideoView
-> true : old video style -> The subtitle will always match video position
-> false: your video style -> The subtitle will always match main view position

Maybe a string prop is better for flexibility in the futur also :/ 

Because depending of app expectation we can choose between the 2 options ?

freeboub:
@YangJonghun 
Please see last merged PR. The proposed change was causing a regression, the original behavior has been restored, so in next release you will have to add a prop to configure it.

PR #3831, opened 2024-05-23
fix(android): allow notification tap to foreground app

paulrinaldi:
- Update the documentation
N/A

- Provide an example of how to test the change
https://github.com/TheWidlarzGroup/react-native-video/assets/107127397/a4f1e82d-47cb-439f-9998-665fb2f70714
set `showNotificationControls={true}` on video player

- Focus the PR on only one area
N/A

- Describe the changes
Passes the activity which spawns the notification service to the service so that the service can set the intent to foreground that service when clicked.

## Summary
Enable tapping on notification to foreground the original app.

### Motivation
Feature parity between iOS and Android. General expectation of tapping a notification.

### Changes
set the pending intent for setSessionActivity on mediaSession creation in Android.

## Test plan
N/A

### Other Notes
Address #3881
We might want to import android.app.Activity instead of using fully qualified type to maintain codestyle.
We might want to pass this in a different way... this is a first attempt that works.
Need to clean this up still

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
@@ -1,8 +1,10 @@
 package com.brentvatne.exoplayer
 
 import android.annotation.SuppressLint
+import android.app.Activity
 import android.app.NotificationChannel
 import android.app.NotificationManager
+import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.os.Binder
@@ -23,6 +25,7 @@ class PlaybackServiceBinder(val service: VideoPlaybackService) : Binder()
 class VideoPlaybackService : MediaSessionService() {
     private var mediaSessionsList = mutableMapOf<ExoPlayer, MediaSession>()
     private var binder = PlaybackServiceBinder(this)
+    private var sourceActivity: Class<Activity>? = null
 
     // Controls
     private val commandSeekForward = SessionCommand(COMMAND_SEEK_FORWARD, Bundle.EMPTY)
@@ -44,10 +47,11 @@ class VideoPlaybackService : MediaSessionService() {
 
     // Player Registry
 
-    fun registerPlayer(player: ExoPlayer) {
+    fun registerPlayer(player: ExoPlayer, from: Class<Activity>) {
         if (mediaSessionsList.containsKey(player)) {
             return
         }
+        sourceActivity = from
 
         val mediaSession = MediaSession.Builder(this, player)
             .setId("RNVideoPlaybackService_" + player.hashCode())
@@ -63,6 +67,7 @@ class VideoPlaybackService : MediaSessionService() {
         hidePlayerNotification(player)
         val session = mediaSessionsList.remove(player)
         session?.release()
+        sourceActivity = null
 
         if (mediaSessionsList.isEmpty()) {
             cleanup()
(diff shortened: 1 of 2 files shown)


paulrinaldi:
Addresses https://github.com/TheWidlarzGroup/react-native-video/issues/3811

paulrinaldi:
FYI @freeboub using `setSessionActivity` did not create the behavior of a tap causing a foreground of the app.

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
YangJonghun, line 1: nit;
I think
```kt
val mediaSession = MediaSession.Builder(this, player)
            .setSessionActivity(PendingIntent.getActivity(this, 0, Intent(this, MainActivity::class.java), PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE))
```
makes code simple

paulrinaldi:
Screenshot 2024-05-22 at 11 34 56 AM

paulrinaldi:
https://github.com/TheWidlarzGroup/react-native-video/assets/107127397/c9cfd7c3-c262-46c2-87d0-0a14ba979d5b

Cleaned up the code and revised the design.

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -44,10 +46,11 @@ class VideoPlaybackService : MediaSessionService() {
 
     // Player Registry
 
-    fun registerPlayer(player: ExoPlayer) {
+    fun registerPlayer(player: ExoPlayer, from: Class<android.app.Activity>) {
         if (mediaSessionsList.containsKey(player)) {
             return
         }
+        startedFrom = from
```
freeboub: Can you please set startedFrom = null in unregisterPlayer ?
maybe you can find a better variable name, something like sourceActivity

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -110,9 +113,13 @@ class VideoPlaybackService : MediaSessionService() {
             return
         }
 
+        val returnToStartedActivity = Intent(this, startedFrom).apply {
```
freeboub: this is also a strange val name: returnToStartedActivity I would expect an Activity.
what do you think about: backToPlayerIntent ?

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
freeboub: (not mandatory, just a proposal)

paulrinaldi: I had a tough time figuring out what to name these, thank you!

freeboub: @paulrinaldi no problem, it is the principle of code review. #IAmTheBadCop 👮‍♂️ ;)

paulrinaldi: Done.

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
paulrinaldi: Done

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
paulrinaldi: Done.

freeboub (approved):

KrzysztofMoch (approved):
LGTM

PR #3832, opened 2024-05-23
fix: android cache header

### LINKED ISSUE #3805: [BUG]: on some video the headers are ignored
### Version

6.0.0

Android

### Architecture

Old architecture

### What happened?

if i set header "Referer",but i think if i set a simple header in general the player ignore it. I check previous version and on beta 5 seem to be ok.
It' happen with mp4 video streaming and not with m3u

### Reproduction

_No response_

### Reproduction

it' hard to reproduce cause i've created an url resolver for server like doodstream

lovegaoshi:
## Summary

fix #3805 

### Motivation

fix #3805 

### Changes

simpleCache stays as a singleton, cacheDataSourceFactory is no longer a singleton and will always be recreated instead

## Test plan

in example app basic, referer and UA are added to github's bigbunny:

image


freeboub (approved):
code looks good to me, let's go.
Thank you for the quick update @lovegaoshi !

PR #3835, opened 2024-05-24
fix(android): refactor source, fix random DRM issue and crop start on local asset

### LINKED ISSUE #2693: Move startup props into a single configuration

freeboub:
## Summary
Move source prop in a dedicated class

### Motivation
Keep code clean and more split
In the same time I fix: 
Also fix a side issue when using a local file cropping props were not applied
Also fix random DRM issue by refactoring initializePlayerSource : https://github.com/TheWidlarzGroup/react-native-video/issues/3082
Prepare another refactor to fix: https://github.com/TheWidlarzGroup/react-native-video/issues/2693 and https://github.com/TheWidlarzGroup/react-native-video/issues/3665

### Changes
Move parsing of source in a dedicated class
Rework usage of the source
Rework MediaMetadata initialization (ping @KrzysztofMoch)
merge setSrc and setRawSrc
Add missing clearResumePosition(); in setSrc
improve logs

## Test plan
Can be tested with the sample


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1770,22 +1757,9 @@ public void setAdTagUrl(final Uri uri) {
...
-            if (!isSourceEqual) {
-                reloadSource();
-            }
-        }
-    }
-
     public void setTextTracks(SideLoadedTextTrackList textTracks) {
         this.textTracks = textTracks;
-        reloadSource();
+        reloadSource(); // FIXME Shall be moved inside source
```
freeboub, line 1776: This fixme is for: https://github.com/TheWidlarzGroup/react-native-video/issues/3665

android/src/main/java/com/brentvatne/common/api/Source.kt
```diff
@@ -0,0 +1,205 @@
...
+    fun isEqual(src: Source?): Boolean {
+        if (src == null) return false
+        return (
+            uri == src.uri &&
+                cropStartMs == src.cropStartMs &&
+                cropEndMs == src.cropEndMs &&
+                startPositionMs == src.startPositionMs &&
+                extension == src.extension
+            )
+    }
```
YangJonghun, line 57: nit;
How about override `Object.equals`? if you override it, you can use "==" within Kotlin

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
YangJonghun: nit;
I think we need to compare injected textTracks and this.textTracks before call reloadSource

android/src/main/java/com/brentvatne/common/api/Source.kt
freeboub: Good idea, I updated the code to override ==, thank you

YangJonghun: fyi) If you override `Object.equals`, `==` works in Kotlin, but not in Java.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Yes and not. Yes you proposal would give a better behavior, but for exemple when we switch from a content with external Subtitle to a content with subtitle, the process is the following:
- we receive setSource of the new url
   -> we will reloadsource
- we receive an empty tracks
   -> we will anyway reload source.

As Exoplayer don't support dynamic external tracks update, I plan to move it into the source class.

I identified all following props which should be moved in source
https://github.com/TheWidlarzGroup/react-native-video/issues/2693#issuecomment-1291154625

Can be done without breaking change I think if the props are wrapped in JS part, but it will need update os ios and android implementation...
Notice that ios implementation is already prepared for this move with `struct VideoSource`

android/src/main/java/com/brentvatne/common/api/Source.kt
freeboub: Thanks for the reminder, let's forget java 😂

freeboub: oups, you are right, I forgot we still use java 🙄 new update pushed !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1706,54 +1705,42 @@ public void onCues(CueGroup cueGroup) {
...
-            this.requestHeaders = headers;
+            this.source = source;
             this.mediaDataSourceFactory =
                     DataSourceUtil.getDefaultDataSourceFactory(this.themedReactContext, bandwidthMeter,
-                            this.requestHeaders);
-            this.customMetadata = customMetadata;
+                            source.getHeaders());
 
+            // refresh custom Metadata
+            customMetadata = ConfigurationUtils.buildCustomMetadata(source.getMetadata());
```
YangJonghun: `!Util.areEqual(this.customMetadata, customMetadata)`
this compare code is missing

freeboub: I don't think this is really mandatory, BTW, I add this check again and test the behavior. Looks good !
I can gain few ms in corner case

YangJonghun: fyi)
In my production environment
https://github.com/TheWidlarzGroup/react-native-video/issues/3763
Without this code, I was experiencing the above issue.🥲
I haven't found the root cause, but it appears to be a race condition caused by the re-render.

freeboub: @YangJonghun can It the DRM issue I fixed ? https://github.com/TheWidlarzGroup/react-native-video/issues/3082

YangJonghun: I understand that you're asking me to reproduce that issue, and It, hard to reproducing it because I don't know exactly how to reproduce it or what's causing it (please let me know if I'm misunderstanding you)

freeboub: No no, I just highlight you the DRM issue ;) 
I put back the test by the way !

YangJonghun: Thanks :)

YangJonghun (approved):

KrzysztofMoch (approved):
Look Good To Me 😎

YangJonghun:
@freeboub 
is there any reason about below changes??
```diff
- eventEmitter.error("Failed to initialize DRM Session Manager Framework!", new Exception("DRM Session Manager Framework failure!"), "3003");
```

PR #3846, opened 2024-05-28
Chore(android): refactor drm props

freeboub:
## Summary
Refactor DRM props into a dedicated class

### Motivation
Keep code clean and modularized

### Changes
Just move all DRM subProps to a dedicated class

## Test plan
Can be tested with sample


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -243,9 +244,8 @@ public class ReactExoplayerView extends FrameLayout implements
     private boolean playInBackground = false;
     private Map<String, String> requestHeaders;
     private boolean mReportBandwidth = false;
-    private UUID drmUUID = null;
-    private String drmLicenseUrl = null;
-    private String[] drmLicenseHeader = null;
+
```
```suggestion
```

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -127,27 +125,10 @@ public void onDropViewInstance(ReactExoplayerView view) {
...
-                        drmKeyRequestPropertiesList.add(key);
-                        drmKeyRequestPropertiesList.add(value);
-                    }
-                    videoView.setDrmLicenseHeader(drmKeyRequestPropertiesList.toArray(new String[0]));
-                }
-                videoView.setUseTextureView(false);
-            }
+        DRMProps drmProps = DRMProps.parse(drm);
+        if (drmProps != null) {
+            videoView.setDrm(drmProps);
```
YangJonghun: nit; If the drm prop is inserted as null, the old drm prop will still be there. wouldn't it be a problem? (I think it's a problem from the old behavior, not from this change)

YangJonghun: If purpose of this PR is to preserve the old behavior, it's fine.

KrzysztofMoch (approved):
LGTM

I agree with @YangJonghun that we should considerate if we need old behaviour

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
freeboub: No It was not the goal of this PR, So I remove the test.
It doesn't make sense to keep previous DRM config when user remove it !
Thank you for the catch !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1738,19 +1755,6 @@ public void setSrc(Source source) {
                     DataSourceUtil.getDefaultDataSourceFactory(this.themedReactContext, bandwidthMeter,
                             source.getHeaders());
 
-            // refresh custom Metadata
```
freeboub: @KrzysztofMoch I moved this part of code because it causes some crash in sample !

KrzysztofMoch: It's fine for me, if this fixes crash 👍

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -117,28 +116,9 @@ public void onDropViewInstance(ReactExoplayerView view) {
...
-                        drmKeyRequestPropertiesList.add(value);
-                    }
-                    videoView.setDrmLicenseHeader(drmKeyRequestPropertiesList.toArray(new String[0]));
-                }
-                videoView.setUseTextureView(false);
-            }
-        }
+        DRMProps drmProps = DRMProps.parse(drm);
+        videoView.setDrm(drmProps);
+        videoView.setUseTextureView(false);
```
YangJonghun, line 121: if drm props is null, we have to follow injected prop (useSecureView or useTextureView)

freeboub: Yes you are right, thank you. I am also wondering if it can make sense to use Surface View by default? What is your opinion?

YangJonghun: It's hard question. because it depends on what user using this library for
I prefer SurfaceView from a user perspective. In most situations, this is [beneficial](https://developer.android.com/media/media3/ui/playerview).
However, it can cause to make issue threads being created by people who don't read the documentation or don't know much about Android.

Decision is yours🫡

freeboub: I am working to remove this videoView.setUseTextureView(false);
It is a bit. The ugly solution is to add new variables recording the last calls to setUseTextureView and setUseSecureView and apply it in the else.

I would prefer implement it in another way:
- move DRM & viewType inside the source prop.
- > It will decrease the bridge usage, avoid restarting playback if we receive DRM prop after source and make cleaner code.

I will try to implement it, but I think we can merge this part now, and the next fix will come soon !

I propose to merge this PR as the setUseTextureView is already present in the code !

freeboub: @YangJonghun here is The patch I would to do for ensure we correctly reset the view state in the app:
https://github.com/TheWidlarzGroup/react-native-video/pull/3867
initially I didn't plan to it right now, but you make me accelerate :)
Still in draft as I need to do the same thing on ios, but that's the idea.
Let me know you thought please !
And ping @KrzysztofMoch

YangJonghun: I apologize for late response.
I left related comment in #3867 .

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -2259,16 +2259,11 @@ public void setBufferConfig(BufferConfig config) {
...
-        this.drmLicenseUrl = licenseUrl;
-    }
-
-    public void setDrmLicenseHeader(String[] header){
-        this.drmLicenseHeader = header;
+    public void setDrm(DRMProps drmProps) {
+        this.drmProps = drmProps;
+        if (drmProps != null && drmProps.getDrmType() != null) {
+            this.drmProps.setDrmUUID(Util.getDrmUuid(drmProps.getDrmType()));
+        }
```
YangJonghun, line 2266:
```java
        if (drmProps != null && drmProps.getDrmType() != null) {
            this.drmProps.setDrmUUID(Util.getDrmUuid(drmProps.getDrmType()));
        }
```
I think above code could be moved inside the DrmProps.parse fn.

YangJonghun (approved):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: I agree, but my problem with that is that I will have to import Utils from media3 in drmProp.kt. I really want to avoid including player specific code in the utils / data management classes ...

YangJonghun: I understand. but for that reason, it seems like parse function should be categorized in a different class. (for example, DRMUtils)
And I think removing dependencies from Utils seems hard to avoid because there's so much to prepare to implement as the SDK requires.🥲

If you worried deep dependencies, I recommend copying [media3's implementation](https://github.com/androidx/media/blob/1.1.0/libraries/common/src/main/java/androidx/media3/common/util/Util.java#L2067-L2083).

However, I'm not suggesting this because I think it's a serious issue, so you don't have to fix it.

PR #3850, opened 2024-05-28
refactor(ios): audio session manager

KrzysztofMoch:
## Summary
Create Audio Session Manager for better and more reliable audio session management 

### Motivation
In old implementation each component is changing audio session properties without carrying about other components - with audio session manager, components will respect other components "needs"

### Note
This is very risky change (at last for me) and need tests to see if everything is working as it should


ios/Video/RCTVideo.swift
```diff
@@ -241,6 +242,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
+        AudioSessionManager.shared.registerView(view: self)
```
freeboub, line 264: There is no unregisterView call, is it really expected ?

freeboub:
Change make sense for me (this is really cleaner I think).
As I don't have ios device, I cannot really test, consider I approuve PR (but please check my small comment!)
Thank you @KrzysztofMoch

YangJonghun:
First of all, thank you for your great work.
I would like to talk about two topics.

1. why is the AudioSession handling within NowPlayingInfo? I think this processing should be a simple structure that is handled when the Player is mounted and unmounted. At the same time, the current audioSession is also handled within setPictureInPicture, so it seems like it could affect the management of audioSessions that work as application singletons.

2. the work you're doing to tie the settings between videos together is great. However, since audioSession is a single instance throughout the application that determines the behavior of the app, I think that if this is necessary, it should be added to detect changes in the audioSession and reset it with a priority determined from the video.

freeboub:
@KrzysztofMoch is it fixing: https://github.com/TheWidlarzGroup/react-native-video/issues/3838 ?

KrzysztofMoch:
It can (or at last will show warning), background audio and `ignoreSilentSwitch="obey"` wont work together as there iw audio session category conflict

ios/Video/RCTVideo.swift
KrzysztofMoch: Oh, I will add unregisterView to deinit 👍
(but this I guess would be fine as it is because there is week object list so it would deinit anyway)

KrzysztofMoch:
>why is the AudioSession handling within NowPlayingInfo?

I am calling `updateAudioSessionCategory` function in places that "needs" some kind of playback category (eg. pictureInPicture needs `playback`). IIRC NowPlayingInfo also require `playback` category (not sure)

>At the same time, the current audioSession is also handled within setPictureInPicture

Ops, will make fix for it (good catch!)

>it should be added to detect changes in the audioSession and reset it with a priority determined from the video.

So we should add some prop to create priority for videos? It's kinda sucks as this will be iOS only need, but I guess that this would be useful 

@YangJonghun

KrzysztofMoch:
Hey @YangJonghun - Thanks for your feedback!

1. PiP and NowPlayingManager (IIRC) needs `playback` category, so I set that video using it and then update audio session with those values "in mind". Also I am calling function on shared instance so I don't understand issue

2. So we probably would need new iOS only prop for it? Something like `audioSessionPriority`?

YangJonghun:
@KrzysztofMoch 
> 1. PiP and NowPlayingManager (IIRC) needs playback category, so I set that video using it and then update audio session with those values "in mind". Also I am calling function on shared instance so I don't understand issue

I wrote that because it is my understanding that setting up an AudioSession overwrites all previous settings. If I'm wrong, please let me know.

> 2. So we probably would need new iOS only prop for it? Something like audioSessionPriority?

This is the same as `1.`, and I wrote it down because it seems like there's room for bugs due to interference from libraries other than react-native-video, but if I'm wrong about AudioSession, this is fine too.

YangJonghun (approved):

freeboub:
@KrzysztofMoch you can merge when you want !

KrzysztofMoch:
Okay, I will describe in docs how this will works to don't confuse peoples - I will do it this week

freeboub:
additionnal comment, you need to replace all usage of: `configureAudio` here the systems are mixed

frankusu:
Hi guys any updates on this ? trying to Play background music when video is muted. Stop background music when video is unmuted [#4156](https://github.com/TheWidlarzGroup/react-native-video/discussions/4156)
Do you guys know when this will get merged and if it will affect anything ?

ios/Video/AudioSessionManager.swift
```diff
@@ -0,0 +1,180 @@
...
+
+        if canAllowMixing {
+            let shouldEnableMixing = videoViews.allObjects.contains { view in
+                view._mixWithOthers == "mix"
+            }
+
+            let shouldEnableDucking = videoViews.allObjects.contains { view in
+                view._mixWithOthers == "duck"
+            }
+
```
frankusu, line 109: I know in the original code theres no case for switching back to inherit, would it be possible to add it in ? 
Because once you set it to 'mix' or 'duck' theres no way to set it back to 'inherit'

gerson:
Hey guys any updates on this? I'm having the same issue 😢

KrzysztofMoch:
To be followed here https://github.com/TheWidlarzGroup/react-native-video/pull/4466

Issue #3732, opened 2024-05-03
[BUG]: Several fullscreen-related issues


bfricka:
### Version

6.0.0-rc.0

### What platforms are you having the problem on?

iOS, Android

### Architecture

Old architecture

### What happened?

On both Android and iOS, entering fullscreen has issues. I'll try to lay them out here.

### Reproduction

**Issue 1: Entering Fullscreen Programmatically**

If you programmatically enter fullscreen on either Android and iOS, and the user dismisses fullscreen, you cannot again enter fullscreen mode programmatically.

Since the `onFullscreenPlayerDidDismiss` event is always called as expected when the user dismisses fullscreen, I hypothesized that some internal state / lifecycles were not being handled properly, and decided to add `videoRef.current?.dismissFullscreenPlayer()` in this event. This fixes the issue and allows the video to be opened in fullscreen again.

Clearly, if the user dismisses fullscreen, the video should handle its internal state and allow

**Issue 2: Cannot Enter Fullscreen on Android w/ Controls Disabled**

Android requires controls to be enabled, otherwise, it will not properly enter fullscreen. With controls enabled, entering fullscreen programmatically properly shows the video in a fullscreen with a black background. With controls disabled, the video just starts playing.

**Issue 3: Programmatically Entering Fullscreen on iOS with Controls Enabled Crashes the App**

If controls are enabled on iOS, entering fullscreen mode programmatically crashes the app. I'm doing this from a fullscreen modal, which has had issues on iOS in fullscreen mode in 5.x. However, that could be fixed by setting `transparent={false}` on the modal. Now, iOS always crashes when trying to programmatically enter fullscreen mode with controls enabled.

seyedmostafahasani:
@bfricka please update the library to the latest version and test it with `setFullScreen` method.

KrzysztofMoch:
No response or activity - let's close this one

PR #3855, opened 2024-05-29
chore: add setFullScreen to component's ref

seyedmostafahasani:
…eenPlayer & dismissFullscreenPlayer

## Summary
I added `setFullScreen` to handle fullScreen mode without controls.

### Motivation
The user can use it to handle fullscreen mode with custom controls, specifically refactoring the codebase and fix this [issue](https://github.com/TheWidlarzGroup/react-native-video/issues/3732)

### Changes
I added a function called `setFullScreen` and deprecated the functions `presentFullscreenPlayer` and `dismissFullscreenPlayer`.

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -465,7 +465,7 @@ public class ReactExoplayerView extends FrameLayout implements
         //Handling the fullScreenButton click event
         final ImageButton fullScreenButton = playerControlView.findViewById(R.id.exo_fullscreen);
         fullScreenButton.setOnClickListener(v -> setFullscreen(!isFullscreen));
-        updateFullScreenButtonVisbility();
+        updateFullScreenButtonVisibility();
         refreshProgressBarVisibility();
 
         // Invoking onPlaybackStateChanged and onPlayWhenReadyChanged events for Player
@@ -2184,18 +2184,14 @@ public class ReactExoplayerView extends FrameLayout implements
         return preventsDisplaySleepDuringVideoPlayback;
     }
 
-    private void updateFullScreenButtonVisbility() {
+    private void updateFullScreenButtonVisibility() {
         if (playerControlView != null) {
             final ImageButton fullScreenButton = playerControlView.findViewById(R.id.exo_fullscreen);
-            if (controls) {
-                //Handling the fullScreenButton click event
-                if (isFullscreen && fullScreenPlayerView != null && !fullScreenPlayerView.isShowing()) {
-                    fullScreenButton.setVisibility(GONE);
-                } else {
-                    fullScreenButton.setVisibility(VISIBLE);
-                }
-            } else {
+            //Handling the fullScreenButton click event
+            if (isFullscreen && fullScreenPlayerView != null && !fullScreenPlayerView.isShowing()) {
                 fullScreenButton.setVisibility(GONE);
+            } else {
+                fullScreenButton.setVisibility(VISIBLE);
             }
         }
     }
@@ -2219,7 +2215,7 @@ public class ReactExoplayerView extends FrameLayout implements
         WindowInsetsControllerCompat controller = new WindowInsetsControllerCompat(window, window.getDecorView());
         if (isFullscreen) {
             eventEmitter.fullscreenWillPresent();
-            if (controls && fullScreenPlayerView != null) {
+            if (fullScreenPlayerView != null) {
                 fullScreenPlayerView.show();
             }
             UiThreadUtil.runOnUiThread(() -> {
@@ -2230,9 +2226,10 @@ public class ReactExoplayerView extends FrameLayout implements
             });
         } else {
             eventEmitter.fullscreenWillDismiss();
-            if (controls && fullScreenPlayerView != null) {
+            if (fullScreenPlayerView != null) {
                 fullScreenPlayerView.dismiss();
                 reLayoutControls();
+                setControls(controls);
             }
             UiThreadUtil.runOnUiThread(() -> {
                 WindowCompat.setDecorFitsSystemWindows(window, true);
@@ -2241,7 +2238,7 @@ public class ReactExoplayerView extends FrameLayout implements
             });
         }
         // need to be done at the end to avoid hiding fullscreen control button when fullScreenPlayerView is shown
-        updateFullScreenButtonVisbility();
+        updateFullScreenButtonVisibility();
     }
 
     public void setUseTextureView(boolean useTextureView) {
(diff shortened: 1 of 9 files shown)


freeboub (changes requested):
We cannot remove existing APIs on 6.X branch, I don't understand why you need this change. Can you please clarify ?

seyedmostafahasani:
I think we can implement full-screen mode with just this function. We don't need two functions to handle it. Additionally, if we remove these functions, we will eliminate an unnecessary useState from the video component, which prevents re-rendering. Specifically, we will also fix another issue in the library, #3732. The main point is that we handle this logic with just one function. What is your opinion?

seyedmostafahasani:
@freeboub what do you think about my points?

freeboub:
@seyedmostafahasani We cannot change api on V6.X, if we integrate such api change, we will need to bump to V7 and all apps using the package will have to upgrade their integration. as the team have a lot of api changes in mind, we try to keep api interoperability as much as possible.

Here is a sample: 
https://github.com/TheWidlarzGroup/react-native-video/pull/3867/files#diff-e2131d0631cdc1ddf0ba36827a27b86adf8e5b8ed162a280a53c6219f4df5a5f

I just move few props from the root of video to the source object, to avoid upgrade issue, I comment in doc that the old api is deprecated, and point to the new way to proceed.

It will allow easy upgrade for apps, we keep in mind that old api (deprecated) can be removed on V7.
Can you please do the same thing please ?
I think merging the 2 apis make sense, but we need to keep compatibility of old api some time

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -2184,18 +2184,14 @@ public boolean getPreventsDisplaySleepDuringVideoPlayback() {
         return preventsDisplaySleepDuringVideoPlayback;
     }
 
-    private void updateFullScreenButtonVisbility() {
+    private void updateFullScreenButtonVisibility() {
         if (playerControlView != null) {
             final ImageButton fullScreenButton = playerControlView.findViewById(R.id.exo_fullscreen);
-            if (controls) {
-                //Handling the fullScreenButton click event
-                if (isFullscreen && fullScreenPlayerView != null && !fullScreenPlayerView.isShowing()) {
```
freeboub, line 2192: why did you remove this section ?
Please check git history, but I think it was fixing a problem !

freeboub (changes requested):

seyedmostafahasani:
> @seyedmostafahasani We cannot change api on V6.X, if we integrate such api change, we will need to bump to V7 and all apps using the package will have to upgrade their integration. as the team have a lot of api changes in mind, we try to keep api interoperability as much as possible.
> 
> Here is a sample: https://github.com/TheWidlarzGroup/react-native-video/pull/3867/files#diff-e2131d0631cdc1ddf0ba36827a27b86adf8e5b8ed162a280a53c6219f4df5a5f
> 
> I just move few props from the root of video to the source object, to avoid upgrade issue, I comment in doc that the old api is deprecated, and point to the new way to proceed.
> 
> It will allow easy upgrade for apps, we keep in mind that old api (deprecated) can be removed on V7. Can you please do the same thing please ? I think merging the 2 apis make sense, but we need to keep compatibility of old api some time

I completely agree with you about deprecating the old API and planning to remove it in V7. I will add deprecation notices for both `presentFullscreenPlayer` and `dismissFullscreenPlayer`. This approach ensures backward compatibility while allowing a smooth transition for users until V7, where we can fully remove the old API.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -2219,7 +2219,7 @@ public void setFullscreen(boolean fullscreen) {
         WindowInsetsControllerCompat controller = new WindowInsetsControllerCompat(window, window.getDecorView());
         if (isFullscreen) {
             eventEmitter.fullscreenWillPresent();
-            if (controls && fullScreenPlayerView != null) {
```
freeboub, line 2222: not sure why you did that, but in the absolute, it makes sense.
BTW, I think this "control" test is not here as if controls == false, fullScreenPlayerView shall also be null

docs/pages/component/methods.mdx
```diff
@@ -12,26 +12,44 @@ This page shows the list of available methods
 
 Take the player out of fullscreen mode.
 
-### `pause`
```
freeboub, line 15: I don't understand why you move pause, It is the first one in alphabetical order !

freeboub:
@seyedmostafahasani Thank you for the fix,
last remaining point, can you please remove deprecated functions in the sample please ? :)

following code shall be simplified:
 
```javascript
  toggleDecoration() {
    this.setState({decoration: !this.state.decoration});
    if (this.state.decoration) {
      this.video?.dismissFullscreenPlayer();
    } else {
      this.video?.presentFullscreenPlayer();
    }
  }
```

docs/pages/component/methods.mdx
```diff
@@ -32,6 +35,21 @@ On iOS, this displays the video in a fullscreen view controller with controls.
 
 On Android, this puts the navigation controls in fullscreen mode. It is not a complete fullscreen implementation, so you will still need to apply a style that makes the width and height match your screen dimensions to get a fullscreen video.
 
+> [!WARNING]
+> deprecated, use setfullScreen method instead
+
+### `setFullScreen`
```
freeboub: There is an ordering issue here !

freeboub (changes requested):

seyedmostafahasani:
@freeboub I removed the controls condition in the `updateFullScreenButtonVisibility` function. When the user sets `controls={false}`, the fullScreen button disappears in fullScreen mode, requiring the user to press the physical back button to exit fullScreen mode, which may cause confusion. What is your opinion?

examples/basic/src/VideoPlayer.tsx
```diff
@@ -431,9 +431,9 @@ class VideoPlayer extends Component {
   toggleDecoration() {
     this.setState({decoration: !this.state.decoration});
     if (this.state.decoration) {
```
freeboub: this.video?.setFullScreen(!this.state.decoration);
would be cleaner! The goal is to remove lines here :)

YangJonghun:
I don't use native controls and fullscreen in this library, but I'm not sure if this breaking change is really necessary in this library.
It's not a duplicate function, so it doesn't have much effect on user DX as long as user pass an argument. Also, It's easy to remove useState and keep the existing API, so why we make breaking change?

freeboub:
It is not a breaking change as previous api are still working! It is just a preparation for api clean up ! (Let say for V7)

YangJonghun:
@freeboub 
oh..then I misunderstood.
(If possible, it would be nice to have something like a new-arch only prop that works synchronously. `ref.fullscreen = true`)

freeboub:
Yes I agree, Fullscreen management is one of the next challenge I think...

KrzysztofMoch (approved):
LGTM

freeboub (approved):

PR #3872, opened 2024-06-03
fix: ensure progress is sent before `onEnd` callback

### LINKED ISSUE #3866: [BUG]: inconsistancy between currentTime and playableDuration in onProgress callback
### Version

6.1.2

iOS, Android

### Architecture

Old architecture

### What happened?

I have differencies between `currentTime` and `playableDuration` params from `onProgress`.
When video plays and stops at the end I expect that `currentTime` will be equal to `playableDuration`. But actual result is `currentTime` less than `playableDuration` for couple of milliseconds (dependent on video duration).
I also compare `currentTime` at the end of video with the video duration value returned from `onLoad` callback; the issue is the same.
It's become more critical for small videos less than 10 sec.

### Reproduction

repository link

### Reproduction

I provide simple code that shows what I do and under comments you can see what I receive in callbacks.
So, when video finishes playing I expect that `currentTime === playableDuration`
```typescript
  const handleVideoProgress = useCallback(
    ({ currentTime, playableDuration, seekableDuration }: OnProgressData) => {
      console.log(currentTime, playableDuration); // 596.2512817382812 596.473333333
    },
    [],
  );
  const handleLoad = useCallback(({ duration }: OnLoadData) => {
      console.log(duration); // 596.473333333
  }, []);
```

```jsx
        <Video
          source={{ uri: 'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4' }}
          ref={videoRef}
          style={styles.video}
          resizeMode={ResizeMode.CONTAIN}
          controls={false}
          onProgress={handleVideoProgress}
          onLoad={handleLoad}
        />
```

freeboub:
## Summary
Add explicitly an event to notify player position before onEnd 

### Motivation
fix: https://github.com/TheWidlarzGroup/react-native-video/issues/3866

### Changes
- Add explicitly onProgress just before sending onEnd
- Ensure progress position is less than duration
- remove unused notification on ios

## Test plan
can be tested with sample by adding log in onProgress


ios/Video/RCTVideo.swift
```diff
@@ -312,7 +312,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     // MARK: - Progress
 
-    func sendProgressUpdate() {
+    func sendProgressUpdate(fromEnd: Bool = false) {
```
KrzysztofMoch: For me `didEnd` is better name for this, WDYT ?
```suggestion
    func sendProgressUpdate(didEnd: Bool = false) {
```

KrzysztofMoch (approved):
LGTM, left one small comment

ios/Video/RCTVideo.swift
freeboub: Yes you are right ! done

PR #3879, opened 2024-06-06
fix(android): fix null pointer exception at playback start with item metadata

freeboub:
## Summary
Ensure we don't cause nullPointer exception (internally to exoplayer) when starting playback with notification enabled

### Motivation
Avoid popup error

### Changes
Extracting notification patch from this pull request:
https://github.com/TheWidlarzGroup/react-native-video/pull/3846/files#diff-a35ae7e6ae6efee3b22522a836bcbcdd9d59a214f087df454e4ccbb8cd5c0e76
Changes consist in initializing only once the mediaMetadata during source initialization 

## Test plan
Can be tested in sample, current issue can be reproduced by going on "no view" with next channel and going back to live channel with channel down

### DIFF
--- a/android/src/main/java/com/brentvatne/common/api/Source.kt
+++ b/android/src/main/java/com/brentvatne/common/api/Source.kt
@@ -14,6 +14,7 @@ import com.brentvatne.common.toolbox.ReactBridgeUtils.safeGetMap
 import com.brentvatne.common.toolbox.ReactBridgeUtils.safeGetString
 import com.facebook.react.bridge.ReadableMap
 import java.util.Locale
+import java.util.Objects
 
 /**
  * Class representing Source props for host.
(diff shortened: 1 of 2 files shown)


freeboub:
@KrzysztofMoch please let me know if it makes sens for you please ?

KrzysztofMoch (approved):
Looks good for me
Only thing is that we are missing `hashCode` method in `Source` class (android studio is giving me warning also) I think we can use what android studio is proposing
```kt
override fun hashCode(): Int {
    var result = uriString?.hashCode() ?: 0
     result = 31 * result + (uri?.hashCode() ?: 0)
     result = 31 * result + startPositionMs
     result = 31 * result + cropStartMs
     result = 31 * result + cropEndMs
     result = 31 * result + (extension?.hashCode() ?: 0)
     result = 31 * result + (metadata?.hashCode() ?: 0)
     result = 31 * result + headers.hashCode()
     return result
}
```
WDYT ?

PR #3881, opened 2024-06-06
fix(ios): Implicit use of 'self' in closure - use 'self.' to make capture semantics explicit (#3764)

### LINKED ISSUE #3764: [BUG]: Implicit use of 'self' in closure; use 'self.' to make capture semantics explicit
### Version

6.0.0-rc.1

iOS

### Architecture

### What happened?

A bug happened!

### Reproduction

repository link

### Reproduction

dibyendu:
Patch commit to fix the error: `Implicit use of 'self' in closure; use 'self.' to make capture semantics explicit`

### Fix
#3764

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -258,6 +258,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private long lastBufferDuration = -1;
     private long lastDuration = -1;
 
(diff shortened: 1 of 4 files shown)


freeboub (approved):
I approuve as I think there is no risk with that PR.
BTW, do you know what is the "root cause" ? If it is a xcode version issue, can you provide it and we should try to add a CI action to build with this version in order to avoid having again the issue ...
And thank you for the patch !

KrzysztofMoch (approved):
If it fix issue then it's fine for me, but it very strange that CI build, and for me and @freeboub it's working also

PR #3886, opened 2024-06-09
fix(android): show controls in notification on older androids

KrzysztofMoch:
## Summary
On older android media session is not automatically adding controls to notification (see bellow)
image

This PR address this issue 

### Motivation
Fix issue with missing controls in notification on older androids

### Changes
- refactor notification api to support older devices 

## Test plan
Tested on various emulators with different android version (android 9, 12, 13, 14)

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
@@ -2,6 +2,7 @@ package com.brentvatne.exoplayer
 
 import android.annotation.SuppressLint
 import android.app.Activity
+import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.PendingIntent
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -148,10 +224,63 @@ class VideoPlaybackService : MediaSessionService() {
...
+
+            handleCommand(commandFromString(actionCommand), session)
+        }
+        return super.onStartCommand(intent, flags, startId)
+    }
+
     companion object {
-        const val COMMAND_SEEK_FORWARD = "SEEK_FORWARD"
-        const val COMMAND_SEEK_BACKWARD = "SEEK_BACKWARD"
+        private const val SEEK_INTERVAL_MS = 10000L
```
freeboub, line 250: there no setter to change SEEK_INTERVAL_MS, I think this is creating a regression ...

freeboub (changes requested):
One small comment, else, it looks good to me

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
KrzysztofMoch: There is no place when we are changing this (and there wasn't earlier either) - now it's private bcs we dont pass it to `VideoPlaybackCallback` as we handle commands int `VideoPlaybackService`
(not sure if I understands it correctly)

freeboub: OK, sorry, I did a mistake, Confusion with: https://thewidlarzgroup.github.io/react-native-video/component/props#controlsstyles

freeboub (approved):

PR #3899, opened 2024-06-12
feat: add `isSeeking` to `onPlaybackStateChanged`

### LINKED ISSUE #3895: [BUG]: play/pause button state is not synchronised with notification controls
### Version

6.2.0

iOS, Android

### System Version

iOS 17.5.1 / android 10 / emulator android 13

Real device, Simulator

### Architecture

Old architecture

### What happened?

It looks like play/pause button state is not synchronised with notification controls.

----------------

Samsung Note 9, android 10

https://github.com/TheWidlarzGroup/react-native-video/assets/64643049/a8c5e0be-f4f3-4adf-be20-f52fe2496e51

----------------

Iphone 13 mini, iOS 17.5.1

https://github.com/TheWidlarzGroup/react-native-video/assets/64643049/dbec059e-a51c-4ff7-be5d-fbc59ac0b3ed

### Reproduction

https://github.com/TheWidlarzGroup/react-native-video/tree/master/examples/basic

### Reproduction

1. Enable notification controls
2. Press play/pause button in the notification controls
3. Check the play/pause button state in a video player

KrzysztofMoch:
## Summary
Add `isSeeking` to `onPlaybackStateChanged` callback

### Motivation
This is pretty useful for devs that are using custom controls, bcs they can check if `onPlaybackStateChanged` was made bcs of pause/resume or seeking

fixes #3895

### Changes
- add  `isSeeking` to `onPlaybackStateChanged`
- on android i have created new variable to track seeking (not sure if it's ok but it's working)

## Test plan
Tested on android & iOS simulators

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -221,7 +221,13 @@ public class ReactExoplayerView extends FrameLayout implements
     private Runnable mainRunnable;
     private boolean useCache = false;
     private ControlsConfig controlsConfig = new ControlsConfig();
-    private boolean wasSeeking = false;
+
+    /*
+    * When user is seeking first called is on onPositionDiscontinuity -> DISCONTINUITY_REASON_SEEK
+    * Then we set if to false when playback is back in onIsPlayingChanged -> true
+    */
+    private boolean isSeeking = false;
+    private long seekPosition = -1;
 
     // Props from React
     private Source source = new Source();
@@ -1621,6 +1627,11 @@ public class ReactExoplayerView extends FrameLayout implements
             return;
         }
 
+        if (isPaused && isSeeking && !buffering) {
+            eventEmitter.seek(player.getCurrentPosition(), seekPosition);
+            isSeeking = false;
+        }
+
         isBuffering = buffering;
         eventEmitter.onVideoBuffer.invoke(buffering);
     }
@@ -1628,8 +1639,8 @@ public class ReactExoplayerView extends FrameLayout implements
     @Override
     public void onPositionDiscontinuity(@NonNull Player.PositionInfo oldPosition, @NonNull Player.PositionInfo newPosition, @Player.DiscontinuityReason int reason) {
         if (reason == Player.DISCONTINUITY_REASON_SEEK) {
-            wasSeeking = true;
-            eventEmitter.onVideoSeek.invoke(player.getCurrentPosition(), newPosition.positionMs % 1000); // time are in seconds /°\
+            isSeeking = true;
+            seekPosition = newPosition.positionMs;
             if (isUsingContentResolution) {
                 // We need to update the selected track to make sure that it still matches user selection if track list has changed in this period
                 setSelectedTrack(C.TRACK_TYPE_VIDEO, videoTrackType, videoTrackValue);
@@ -1680,11 +1691,15 @@ public class ReactExoplayerView extends FrameLayout implements
 
     @Override
     public void onIsPlayingChanged(boolean isPlaying) {
-        if (isPlaying) {
-            wasSeeking = false;
+        if (isPlaying && isSeeking) {
+            eventEmitter.seek(player.getCurrentPosition(), seekPosition);
         }
 
-        eventEmitter.onVideoPlaybackStateChanged.invoke(isPlaying, wasSeeking);
+        eventEmitter.onVideoPlaybackStateChanged.invoke(isPlaying, isSeeking);
+        
+        if (isPlaying) {
+            isSeeking = false;
+        }
     }
 
     @Override
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -221,6 +221,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private Runnable mainRunnable;
     private boolean useCache = false;
     private ControlsConfig controlsConfig = new ControlsConfig();
+    private boolean wasSeeking = false;
 
     // Props from React
     private Source source = new Source();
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1613,6 +1614,7 @@ private void onBuffering(boolean buffering) {
     @Override
     public void onPositionDiscontinuity(@NonNull Player.PositionInfo oldPosition, @NonNull Player.PositionInfo newPosition, @Player.DiscontinuityReason int reason) {
         if (reason == Player.DISCONTINUITY_REASON_SEEK) {
+            wasSeeking = true;
```
freeboub: It should be false here ? If I am not wrong, onPositionDiscontinuity is called at the end of seek, that's why we send seek end event: "eventEmitter.seek" 
It would make more sens to set `wasSeeking = true;` in `public void seekTo(long positionMs)`.

Not mandatory for the PR, But I also wonder if this seek event shall be reported during initial seeek when you use source.startPosition

KrzysztofMoch: Correct me if I am wrong - seekTo is our method so controls & external playback management will not be caught in this method 

> Not mandatory for the PR, But I also wonder if this seek event shall be reported during initial seeek when you use 
source.startPosition

yes I think we should not report it in this case

freeboub: Not mandatory for the PR, But I also wonder if this seek event shall be reported during initial seeek when you use
source.startPosition

yes I think we should not report it in this case
--> OK, agreed

Correct me if I am wrong - seekTo is our method so controls & external playback management will not be caught in this method
--> No sure, but yes seekTo is the internal call, but there is also direct call to player.seekTo ...
for exemple if use seek from notifications, will the change work correctly ?

KrzysztofMoch: Okay - let me re-do this then

YangJonghun:
Personally, I'd like to be able to track the state of the player rather than individual variables.
Attach [expo-video docs link](https://docs.expo.dev/versions/latest/sdk/video/#videoplayerstatus)([code link](https://github.com/expo/expo/blob/sdk-51/packages/expo-video/android/src/main/java/expo/modules/video/enums/PlayerStatus.kt)) as a reference.

android/src/main/java/com/brentvatne/common/react/VideoEventEmitter.java
```diff
@@ -340,9 +340,10 @@ public void buffering(boolean isBuffering) {
         receiveEvent(EVENT_BUFFER, map);
     }
 
-    public void playbackStateChanged(boolean isPlaying) {
+    public void playbackStateChanged(boolean isPlaying, boolean isSeeking) {
         WritableMap map = Arguments.createMap();
         map.putBoolean(EVENT_PROP_IS_PLAYING, isPlaying);
+        map.putBoolean("isSeeking", isSeeking);
```
freeboub: please extract isSeeking to a constant

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1664,7 +1677,15 @@ public void onVolumeChanged(float volume) {
...
     @Override
     public void onIsPlayingChanged(boolean isPlaying) {
-        eventEmitter.playbackStateChanged(isPlaying);
+        if (isPlaying && isSeeking) {
+            eventEmitter.seek(player.getCurrentPosition(), seekPosition);
+        }
+
+        eventEmitter.playbackStateChanged(isPlaying, isSeeking);
+
+        if (isPlaying) {
```
freeboub, line 1700: this test should be before  eventEmitter.playbackStateChanged ?
becuase you are changing isSeeking, but not sending it to app ?! 🤔

KrzysztofMoch:
> Personally, I'd like to be able to track the state of the player rather than individual variables. Attach [expo-video docs link](https://docs.expo.dev/versions/latest/sdk/video/#videoplayerstatus)([code link](https://github.com/expo/expo/blob/sdk-51/packages/expo-video/android/src/main/java/expo/modules/video/enums/PlayerStatus.kt)) as a reference.

@YangJonghun Good idea! But let's merge this one - then I will do it in next PR (as PR with this change will deprecate current API (isPlaying, isSeeking)

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: Yes that's correct the flow is:
1. when seek "is detected" `onPositionDiscontinuity` is called and in it we are setting `isSeeking` to true
2. after seeking is done (and we want to triger `onSeek` on JS side) two thing can happen:
 A) If player was not paused when user seek -> `onIsPlayingChanged` will be called
 B) If player was paused when user seek -> we will catch that loading is finished in `onBuffering`
 
We are sending event with `isSeeking=true` bcs we want to inform that `onPlaybackStateChanged` was called bcs of seeking, and then we will set in back to `false`

This is needed to check if we should show spiner when user is seeking (custom controls) - otherwise we don't know if `onPlaybackStateChanged` was called bcs of "pause/resume" or "seeking"

So I believe this is correct

freeboub:
@KrzysztofMoch this one conflict

KrzysztofMoch:
@freeboub Done!

freeboub (approved):

zoriya:
@KrzysztofMoch I agree with @YangJonghun there, having the state would be better. I'd like to use `onPlaybackStateChanged` for external changes to play/pause (hardware button, audio focus...) to update the UI.

Currently video buffer or seeking triggers this event. Seeking can be ignored reading the `isSeeking` value introduced in this PR but buffering need special control since the `onBuffer` event is called AFTER the `onPlaybackStateChanged`

PR #3909, opened 2024-06-14
feat: add plugins management

freeboub:
## Summary
Create a plugin system to allow other package to retrieve the player handle.
These side package can mainly handle analytics events

### Motivation
Add an easy way to track analytics management.
will answer to discussion: https://github.com/TheWidlarzGroup/react-native-video/discussions/3564

### Changes
on android and IOS, 
Add ReactNativeVideoManager, singleton to access to the package entry point
Each view register themselves in ReactNativeVideoManager
Add apis to register plugins in ReactNativeVideoManager
When creating a new player plugin system will be called (see documentation)
A sample plugin has also been added to test

## Test plan
N/A


KrzysztofMoch (commented):
Code and "idea" looks good to me 👍
I am also thinking about changing naming from "analytics plugins" -> "plugins" bcs this will allow to do way more things - not just analytics (or at last I think that)

I have not tested it yet - but will try do let you know to end of the week

freeboub:
> Code and "idea" looks good to me 👍 I am also thinking about changing naming from "analytics plugins" -> "plugins" bcs this will allow to do way more things - not just analytics (or at last I think that)
> 
> I have not tested it yet - but will try do let you know to end of the week

I agree for the renaming, I will do it. I tried yesterday to implement a thumbnail preview feature with the plugin system, but the sample I found are not so simple (thumbnails in mjpeg format). I will try again in few days !

ios/Video/ReactNativeVideoManager.swift
```diff
@@ -0,0 +1,57 @@
...
+    private let expectedMaxVideoCount = 10
+
+    // create a private initializer
+    private init() {}
+
+    public static let shared: ReactNativeVideoManager = {
+        let instance = ReactNativeVideoManager()
+        // setup code
+        return instance
+    }()
```
KrzysztofMoch: We can simplify this
```suggestion
    public static let shared: ReactNativeVideoManager = ReactNativeVideoManager()
```
(Btw if you look bellow there are comments for "ReactExoplayerViewManager" on iOS 😉)

KrzysztofMoch:
@freeboub I have tested this PR (via example plugin - thanks for making it 😄) and it works on iOS and Android - Great Job!

ios/Video/ReactNativeVideoManager.swift
freeboub: Done thank you

freeboub:
> Code and "idea" looks good to me 👍 I am also thinking about changing naming from "analytics plugins" -> "plugins" bcs this will allow to do way more things - not just analytics (or at last I think that)
> 

I did the change! THanks

KrzysztofMoch (approved):

venky145:
@freeboub Thanks for implementation.
When we listen to RNVPlugin (onInstanceCreated) we get avaplayer instance but which is just allocation but not the player item. Better to broadcast after the player item set but not prior

In RCTVideo -> func setupPlayer -> 

```
if _player == nil {
            _player = AVPlayer()
            ReactNativeVideoManager.shared.onInstanceCreated(id: instanceId, player: _player as Any)

            _player!.replaceCurrentItem(with: playerItem)

            if _showNotificationControls {
                // We need to register player after we set current item and only for init
                NowPlayingInfoCenterManager.shared.registerPlayer(player: _player!)
            }
        } else {
            #if !os(tvOS) && !os(visionOS)
                if #available(iOS 16.0, *) {
}
}
```
                
                
             
hope we should move 
`**ReactNativeVideoManager.shared.onInstanceCreated(id: instanceId, player: _player as Any)**
`after setting player item
        
 Correct me If I am wrong

PR #3910, opened 2024-06-14
fix(typescript): type checks on selectedTextTrack, selectedAudioTrack, selectedVideoTrack

Polarisation:
## Summary

Fix type checks on `selectedTextTrack`, `selectedAudioTrack` and `selectedVideoTrack` properties which were added in #3704.

### Motivation

[selectedTextTrack](https://thewidlarzgroup.github.io/react-native-video/component/props#selectedtexttrack), [selectedAudioTrack](https://thewidlarzgroup.github.io/react-native-video/component/props#selectedaudiotrack) and [selectedVideoTrack](https://thewidlarzgroup.github.io/react-native-video/component/props#selectedvideotrack) can both be an object with a `type` property and an undefined `value`. This allows `react-native-video` to distinguish between disabled subtitles and subtitles following system settings.

However this behavior was broken by the addition of an incorrect type check in #3704 which does not allow the `value` property to be undefined.

### Changes

Adapted the type check to allow `value` property to be `undefined`, as per documentation and TypeScript type definitions.

## Test plan

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -194,8 +194,15 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const typeOfValueProp = typeof selectedTextTrack.value;
-      if (typeOfValueProp !== 'number' && typeOfValueProp !== 'string' && typeOfValueProp !== 'undefined') {
-        console.log('invalid type provided to selectedTextTrack.value: ', typeOfValueProp);
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
+          'invalid type provided to selectedTextTrack.value: ',
+          typeOfValueProp,
+        );
         return;
       }
       return {
@@ -209,8 +216,15 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const typeOfValueProp = typeof selectedAudioTrack.value;
-      if (typeOfValueProp !== 'number' && typeOfValueProp !== 'string' && typeOfValueProp !== 'undefined') {
-        console.log('invalid type provided to selectedAudioTrack.value: ', typeOfValueProp);
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
+          'invalid type provided to selectedAudioTrack.value: ',
+          typeOfValueProp,
+        );
         return;
       }
 
@@ -225,8 +239,15 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const typeOfValueProp = typeof selectedVideoTrack.value;
-      if (typeOfValueProp !== 'number' && typeOfValueProp !== 'string' && typeOfValueProp !== 'undefined') {
-        console.log('invalid type provided to selectedVideoTrack.value: ', typeOfValueProp);
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
+          'invalid type provided to selectedVideoTrack.value: ',
+          typeOfValueProp,
+        );
         return;
       }
       return {
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -194,8 +194,8 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const type = typeof selectedTextTrack.value;
-      if (type !== 'number' && type !== 'string') {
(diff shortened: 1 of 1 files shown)


freeboub (approved):

src/Video.tsx
```diff
@@ -194,8 +194,8 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const type = typeof selectedTextTrack.value;
-      if (type !== 'number' && type !== 'string') {
-        console.log('invalid type provided to selectedTextTrack');
+      if (type !== 'number' && type !== 'string' && type !== 'undefined') {
+        console.log('invalid type provided to selectedTextTrack.value');
```
freeboub:
```suggestion
        console.log('invalid type provided to selectedTextTrack.type: ', type);
```

src/Video.tsx
```diff
@@ -209,8 +209,8 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         return;
       }
       const type = typeof selectedAudioTrack.value;
-      if (type !== 'number' && type !== 'string') {
-        console.log('invalid type provided to selectedAudioTrack');
+      if (type !== 'number' && type !== 'string' && type !== 'undefined') {
+        console.log('invalid type provided to selectedAudioTrack.value');
```
freeboub:
```suggestion
        console.log('invalid type provided to selectedAudioTrack.value: ', type);
```

freeboub (changes requested):
thank you for the patch, I am just proposing you to improve the log. I let you integrate the change

freeboub (approved):

src/Video.tsx
Polarisation: It actually is the type of the `selectedTextTrack.value` property that this refers to. But it was confusing because there is also a `.type` property but the `type` variable refers to the type of the `.value` propery. So to avoid confusion in future I have renamed the variable to `typeOfValueProp`.

Polarisation:
@freeboub Thank you, I made those changes. I also applied the same fix to `selectedVideoTrack` as I noticed this had the same bug.

freeboub (approved):

freeboub:
> @freeboub Thank you, I made those changes. I also applied the same fix to `selectedVideoTrack` as I noticed this had the same bug.

You are right, thank you !

freeboub (approved):

src/Video.tsx
```diff
@@ -193,9 +193,16 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+      const typeOfValueProp = typeof selectedTextTrack.value;
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
+          'invalid type provided to selectedTextTrack.value: ',
+          typeOfValueProp,
+        );
```
KrzysztofMoch: Let's use `console.error` for those

freeboub: Maybe console.warn is enough? As you want!

src/Video.tsx
```diff
@@ -193,9 +193,16 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
-      const type = typeof selectedTextTrack.value;
-      if (type !== 'number' && type !== 'string') {
-        console.log('invalid type provided to selectedTextTrack');
+      const typeOfValueProp = typeof selectedTextTrack.value;
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
```
freeboub:
```suggestion
        console.warn(
```

src/Video.tsx
```diff
@@ -208,9 +215,16 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
-      const type = typeof selectedAudioTrack.value;
-      if (type !== 'number' && type !== 'string') {
-        console.log('invalid type provided to selectedAudioTrack');
+      const typeOfValueProp = typeof selectedAudioTrack.value;
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
```
freeboub:
```suggestion
        console.warn(
```

src/Video.tsx
```diff
@@ -224,9 +238,16 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
-      const type = typeof selectedVideoTrack.value;
-      if (type !== 'number' && type !== 'string') {
-        console.log('invalid type provided to selectedVideoTrack');
+      const typeOfValueProp = typeof selectedVideoTrack.value;
+      if (
+        typeOfValueProp !== 'number' &&
+        typeOfValueProp !== 'string' &&
+        typeOfValueProp !== 'undefined'
+      ) {
+        console.log(
```
freeboub:
```suggestion
        console.warn(
```

PR #3915, opened 2024-06-16
fix(JS): safety check on resolve uri

freeboub:
## Summary
Add safety check on resolveAssetUri function call

### Motivation
I found the issue once, but I think it can heppen during development phase. It will avoid a crash and give informations to app developer

### Changes
Refactor resolveAssetSource to check result and avoid undefined.uri call

## Test plan
N/A


KrzysztofMoch (approved):
This makes sense - I also saw this error/crash on my friend's application

PR #3922, opened 2024-06-18
feat(ios): detect livestreams in now playing notification 

paul-rinaldi:
…am playing

This addresses part of https://github.com/TheWidlarzGroup/react-native-video/issues/3902

Apple however does encourage going further, "Consider indicating the progress of currently playing live content. People appreciate knowing where they’ll land when they jump into in-progress live content. You can use a progress bar or other indicator to show people how much content remains." (source: [Apple Design Guidelines](https://developer.apple.com/design/human-interface-guidelines/live-viewing-apps)). _**However**_ , Apple does not indicate if this goes for the minimal view of a Now Playing view. So this may be sufficient for many developers and audiences.

- Provide an example of how to test the change
To test the change, 
1. generate an ios build of the example/basic app, 
2. load it on a physical ios device, 
3. tap `show notification controls` on the ui, 
4. Drag from the top to see the notification controls

- Describe the changes
## Summary
Adds "LIVE" decorator to livestream notification (now playing)

Playback (can't remember which it was in example):
Screenshot 2024-06-18 at 11 08 46 AM
Live (red bull live, another live sample):
Screenshot 2024-06-18 at 10 18 59 AM

### Motivation
Saw that livestreams in our app show just --_-- <progress bar> --_-- which is what some apps (AP News, CNN (their pause/play doesn't work correctly though)) do on iOS but a few we saw show a LIVE decorator. To ensure our audience understand what is being shown, we should add this code.
iOS Apps that show --_-- on each side of a livestream now playing control center view: Fox (doesn't allow background play), FM Radio, NBC

### Changes
Added a check to determine if live content is currently playing

## Test plan
None

### DIFF
--- a/ios/Video/NowPlayingInfoCenterManager.swift
+++ b/ios/Video/NowPlayingInfoCenterManager.swift
@@ -226,6 +226,12 @@ class NowPlayingInfoCenterManager {
             })
         }
 
+        if CMTIME_IS_INDEFINITE(currentItem.asset.duration) {
+            nowPlayingInfo[MPNowPlayingInfoPropertyIsLiveStream] = true
+        } else {
+            nowPlayingInfo[MPNowPlayingInfoPropertyIsLiveStream] = false
+        }
+
         nowPlayingInfo[MPMediaItemPropertyPlaybackDuration] = currentItem.duration.seconds
         nowPlayingInfo[MPNowPlayingInfoPropertyElapsedPlaybackTime] = currentItem.currentTime().seconds
         nowPlayingInfo[MPNowPlayingInfoPropertyPlaybackRate] = player.rate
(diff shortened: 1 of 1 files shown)


ios/Video/NowPlayingInfoCenterManager.swift
```diff
@@ -226,6 +226,12 @@ class NowPlayingInfoCenterManager {
             })
         }
 
+        if let urlAsset = currentItem.asset as? AVURLAsset, urlAsset.url.absoluteString.hasSuffix("m3u8") {
```
freeboub: You cannot consider all m3u8 as live, please a a dedicated flag in the metadata api to set this flag

freeboub (changes requested):

ios/Video/NowPlayingInfoCenterManager.swift
YangJonghun: I think we can check live stream with `CMTIME_IS_INDEFINITE(_playerItem.asset.duration)` value

KrzysztofMoch (approved):
besides comment from freeboub, It looks good to me

paul-rinaldi:
Using @YangJonghun 's suggestion worked, live videos show as live and normal show as normal ones in `example/basic`! Worked.

As a side note, I tried the metadata way and that was not working, though it was fairly complex due to the iOS types so that is where it could have not worked correctly. Although passing a flag via metadata seemed, it may not be possible.

freeboub (approved):

freeboub:
> Using @YangJonghun 's suggestion worked, live videos show as live and normal show as normal ones in `example/basic`! Worked.
> 
> As a side note, I tried the metadata way and that was not working, though it was fairly complex due to the iOS types so that is where it could have not worked correctly. Although passing a flag via metadata seemed, it may not be possible.

That works for me ! Thank you

paulrinaldi:
Corrected the styling.

freeboub (approved):

freeboub (approved):

PR #3924, opened 2024-06-18
chore(android): video player play/pause during incoming calling

seyedmostafahasani:
## Summary
Handle video player play/pause during incoming calling

### Motivation
close this [issue](https://github.com/TheWidlarzGroup/react-native-video/issues/3871)

### Changes
Add a new listener to handle different situations.

## Test plan
You can play the video during an incoming call the video will pause automatically, and after the call ends, the video will resume.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerPhoneStateListener.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerPhoneStateListener.java
@@ -4,10 +4,10 @@ import android.telephony.TelephonyManager;
 
 import androidx.media3.exoplayer.ExoPlayer;
 
-public class PhoneStateListener extends PhoneStateListener {
+public class ExoPlayerPhoneStateListener extends PhoneStateListener {
     private final ExoPlayer exoPlayer;
 
-    public PhoneStateListener(ExoPlayer exoPlayer) {
+    public ExoPlayerPhoneStateListener(ExoPlayer exoPlayer) {
         this.exoPlayer = exoPlayer;
     }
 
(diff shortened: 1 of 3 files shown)


YangJonghun:
External interruptions, including incoming calls, are typically controlled by listening for audio focus events.
fyi) https://developer.android.com/media/optimize/audio-focus

seyedmostafahasani:
@YangJonghun thanks for sharing your knowledge with me.
Which way do you prefer to use it?

freeboub:
I agree with @YangJonghun it wouldbe better to handle the change only with audio focus management if possible.
@seyedmostafahasani I am also wondering if the listener you added need some specific permissions, for my general knowledge ?

seyedmostafahasani:
Sure, I agree with you. Let me change the code.
@freeboub, we need specific permissions for protected information. In this case, we don't need any permissions. For example, if we want to listen the phone state, we must add the `READ_PHONE_STATE` permission.

YangJonghun:
@seyedmostafahasani 
Already onAudioFocusChanged prop exposed. If we need pausing, we can implement within onAudioFocusChanged.
Some requirements may simply want the sound to go away during a call, while others may want a pause.
However, this implementation enforces the pause behavior internally.

seyedmostafahasani:
@YangJonghun, The user can handle it with this prop, but as you said earlier, I want to implement it internally specifically for incoming calls. We have quite a few users who are trying to learn RN recently, so they can’t work with these props well.

freeboub:
I have a major issue with the requested permission here!
We need at least to put in the default manifest
But putting this permission by default in the manifest will request new requierment to push the app on the store, see: https://stackoverflow.com/questions/41234205/warnings-your-apk-is-using-permissions-that-require-a-privacy-policy-android-p 

To be able to merge this PR, we need:
- document the new permission requierment and indicates this is optional
- ensure that an without the permission the app still work correctly

according to @YangJonghun I think this phone call handling shall be done only with the onAudioFocusChanged callback...

And regarding to the expected behavior (pause / mute / stop playback) a new prop should be added to indicate the expected behavior.
for exemple:
`audioFocusBehavior={'pause'}`
Notice that pause may not a possible behavior for live content.

In a second step this `audioFocusBehavior` shall be integrated with ios to have a unified api !

PR #3932, opened 2024-06-22
feat(android): allow building exoplayer from source

freeboub:
## Summary
Add configuration and documentation to be able to build from media3 / Exoplayer source code

### Motivation
I had to do a small workaround in exoplayer source code to develop a new feature/poc

### Changes
- Add configuration to build from source
- document the feature
- implement it in the sample. It is not enabled by default, but can be activated with only 2 gradle script uncommenting 

## Test plan
Easy to test by change the sample,  @KrzysztofMoch do you think I need to enable 1 dedicated build from source in the github CI ? My issue with that is that sample will need to clone media3 source code, it will highly increase folder size for developement environment ...


freeboub:
@KrzysztofMoch This one is ready for review !

examples/basic/android/app/build.gradle
```diff
@@ -81,12 +81,20 @@ android {
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     namespace "com.videoplayer"
+
+    compileOptions {
+        // These options are necessary to be able to build fro source
+        coreLibraryDesugaringEnabled true
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
+    }
```
KrzysztofMoch, line 90: Shouldn't this be mentioned in docs ?

examples/basic/android/app/build.gradle
```diff
@@ -147,6 +155,8 @@ dependencies {
             because("kotlin-stdlib-jdk8 is now a part of kotlin-stdlib")
         }
     }
+    // coreLibraryDesugaring is mandatory to be able to build exoplayer from source
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.4'
```
KrzysztofMoch, line 159: Also this

KrzysztofMoch (approved):
Looks good to me, left one small comment

examples/basic/android/app/build.gradle
freeboub: Yes you are right, but this is already in this section:
### Desugaring
I pointed the google documentation instead of copying it !

examples/basic/android/app/build.gradle
freeboub: The same :) 

Yes you are right, but this is already in this section:
### Desugaring
I pointed the google documentation instead of copying it !

freeboub:
> Looks good to me, left one small comment

Great, thank you, so let's merge this one ! 👍

YangJonghun:
@freeboub 
Looks great to me! sorry for the late confirmation.
By the way, do you have any plans to bump up a media3 version?

freeboub:
For media3 version, I tried few month ago, I had some issue regarding to gradle plugin. But you can force the version in the gradle file with following value. The react native version used in sample was not compatible with media3... 

RNVideo_media3Version=1.1.1

Notice also that in my thumbnail poc, I use master branch of media3 so version 1.3.1 shall work (no blocking point on API change).

YangJonghun:
@freeboub 
I asked because it was a target version issue and I wanted to insert the add-on from 1.2.0 if the library raised the default.

freeboub:
@YangJonghun I am not sure to understand exactly what you want do ... Upgrading locally to 1.2.0 doesn't ?

YangJonghun:
@freeboub 
I'd like to contribute for add CMCD prop.
However, this function require 1.2.0 or later version of media3, which is difficult to do in the current default version.

PR #3933, opened 2024-06-23
feat: add expo plugins

KrzysztofMoch:
## Summary
Add support for expo prebuild via expo plugins

### Motivation
Allow user to set native configuration from `app.json` / `app.config.js` and allow to auto generate those configs

### Changes
- change ts build to `CommonJS` as expo plugins require to be build that way

## Test plan
- created empty expo app
- installed library
- imported plugin (see docs)
- all options should work


freeboub:
looks good to me (even if it is not building yet).
I think you should document this change as it makes expo integration easier !

android/build.gradle
```diff
@@ -45,7 +45,7 @@ def ExoplayerDependenciesList = [
 ]
 
 def ExoplayerDependencies = ExoplayerDependenciesList.collectEntries { property ->
-    [(property): safeExtGet(property)?.toBoolean() ?: false]
+    [(property): safeExtGet('RNVideo_' + property)?.toBoolean() ?: false]
```
freeboub: Isn't it breaking current behavior ? or is this fixing an issue ?

KrzysztofMoch: You are right this can be breaking - I want to move all our properties to follow this `RNVideo_{NAME}` - will change it to support backward compatibility and update docs in other PR

KrzysztofMoch:
@freeboub I have comment out plugin in example app bcs of dependency issue - so to test it you will need to create new expo app and install `react-native-video` from this PR (Ofc I have already done it and it was working very good for me 😅)

freeboub (approved):
OK, maybe you can enable it in the basic sample also ?

KrzysztofMoch:
To do it we should remove `android` & `ios` folder and use [expo prebuild](https://docs.expo.dev/workflow/prebuild/). Actually we can do it - I can split (in other PR) `basic` example to `expo` & `bare`

freeboub:
> To do it we should remove `android` & `ios` folder and use [expo prebuild](https://docs.expo.dev/workflow/prebuild/). Actually we can do it - I can split (in other PR) `basic` example to `expo` & `bare`

If it is mandatory to remove `android` & `ios` folder to use expo plugin, OK, it make sense.
I think we can keep the tvos sample to develop on bare react native. As you know it is painfull to maintain multiple samples

freeboub:
I think you can merge this one and do basic sample changes in another PR

PR #3940, opened 2024-06-24
chore(android): rework view type

freeboub:
## Summary
rework view type management 

### Motivation
Clean up api and avoid application mistakes
extract part of changes from: https://github.com/TheWidlarzGroup/react-native-video/pull/3867

### Changes
⚠️ Behavior change ! the default view type is SURFACE VIEW instead of TEXTURE VIEW
deprecate useTextureView and useSecureView
merge both props internally to a new prop: viewType

## Test plan
Can be tested with sample


src/Video.tsx
```diff
@@ -571,6 +575,34 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+        'cannot use DRM on texture view. please set useTextureView={false}',
+      );
+    }
+    if (useSecureView && useTextureView) {
+      console.warn(
+        'cannot use SecureView on texture view. please set useTextureView={false}',
+      );
+    }
+
+    const _viewType = shallForceViewType
```
freeboub: maybe we should move this logic inside a memo :/ 
here in case of bad config a warning will be displayed on every render ...

KrzysztofMoch: @freeboub I think you can wrap all this code in `useMemo` like this:
```jsx
    const _viewType = useMemo(() => {
      const hasValidDrmProp =
        drm !== undefined && Object.keys(drm).length !== 0;

      const shallForceViewType =
        hasValidDrmProp && (viewType === ViewType.TEXTURE || useTextureView);

      if (shallForceViewType) {
        console.warn(
          'cannot use DRM on texture view. please set useTextureView={false}',
        );
      }
      if (useSecureView && useTextureView) {
        console.warn(
          'cannot use SecureView on texture view. please set useTextureView={false}',
        );
      }

      const _viewType = shallForceViewType
        ? useSecureView
          ? ViewType.SURFACE_SECURE
          : ViewType.SURFACE // check if we should force the type to Surface due to DRM
        : viewType
        ? viewType // else use ViewType from source
        : useSecureView // else infer view type from useSecureView and useTextureView
        ? ViewType.SURFACE_SECURE
        : useTextureView
        ? ViewType.TEXTURE
        : ViewType.SURFACE;
    }, [drm, useTextureView, useSecureView, viewType]);
 ```

KrzysztofMoch (approved):
Looks good for me - only this memo thing and I am good to merge this

src/Video.tsx
freeboub: Done, thank you

freeboub:
Let's merge, I would like to close the DRM prop PR just after

PR #3949, opened 2024-06-27
chore(example/basic): refactor state variable

seyedmostafahasani:
## Summary
I've refactored the state variable.

### Motivation
Fixing some probable issues and readability of the code.

## Test plan


seyedmostafahasani:
@freeboub I've created this PR from [this PR](https://github.com/TheWidlarzGroup/react-native-video/pull/3934), so first, we must merge the parent PR. Then, we will merge this PR.

KrzysztofMoch:
@seyedmostafahasani can you do a rebase?

seyedmostafahasani:
@KrzysztofMoch it's done!

examples/basic/src/VideoPlayer.tsx
```diff
@@ -25,53 +25,95 @@ import Video, {
...
+      type: SelectedVideoTrackType.AUTO,
+    });
+  const [srcListId, setSrcListId] = useState(0);
+  const [repeat, setRepeat] = useState(false);
+  const [controls, setControls] = useState(false);
+  const [useCache, setUseCache] = useState(false);
+  const [poster, setPoster] = useState<string | undefined>(undefined);
+  const [showNotificationControls, setShowNotificationControls] =
+    useState(false);
+  const [isSeeking, setIsSeeking] = useState(false);
```
KrzysztofMoch, line 72: @seyedmostafahasani it's not necessary but how about to move it to `useReducer`? I think It would be more readable (btw sorry for long review time)

seyedmostafahasani: @KrzysztofMoch Actually, I don't particularly like `useReducer` 😉, but if you want, I can use it.

freeboub: The best solution is to create a hook in RNV to retrieve all these informations from the player !

freeboub:
@seyedmostafahasani Can you please re split the Overlay file ? else, it looks good to me

seyedmostafahasani:
@freeboub I had to revert the Overlay file because I needed to pass many props to the component. If we handle all states with a custom hook, as you suggested, we can split this file again.

examples/basic/src/VideoPlayer.tsx
seyedmostafahasani: @freeboub as @KrzysztofMoch said earlier, I will handle it with `useReducer`.

freeboub: I don't think this is a good solution ... 
You will introduce new dependency in the sample, I don't think this is the way to go as all app have similar issues. It would be better to provide a common solution for that !
Please don't do it in this PR

freeboub:
> @freeboub I had to revert the Overlay file because I needed to pass many props to the component. If we handle all states with a custom hook, as you suggested, we can split this file again.

please re split it even if it is not the best solution for now !

seyedmostafahasani:
@freeboub it's done.

freeboub (approved):

freeboub:
@seyedmostafahasani OK for me, thank you !

seyedmostafahasani:
@freeboub no problem 🙏

KrzysztofMoch (approved):

PR #3950, opened 2024-06-28
fix(android): added setAllowChunklessPreparation to HlsMediaSource.Factory stub

### LINKED ISSUE #3948: [BUG]: Android project no longer builds if useExoplayerHls = false
### Version

6.3.0

Android

### System Version

Any

Real device

### Architecture

Old architecture

### What happened?

The fix for #3766 means that any app that doesn't choose to include the Media3 HLS libraries will now fail to build.  This is due to the new requirement for setAllowChunklessPreparation to exist on HlsMediaSource.Factory.

A stub implementation exists in the project to allow code to compile without including the HLS libraries - I believe it just needs to be updated to add the stub method. 

### Reproduction

repository link

### Reproduction

Try to build the app with useExoplayerHls = false

EETVApps:
## Summary

Updated the HlsMediaSource.Factory stub to include setAllowChunklessPreparation, to allow the Android project to build when Media3 HLS libraries are excluded (useExoplayerHls = false)

### Motivation

Fix #3948

## Test plan

Successful build

### DIFF
--- a/android/src/main/java/androidx/media3/exoplayer/hls/HlsMediaSource.java
+++ b/android/src/main/java/androidx/media3/exoplayer/hls/HlsMediaSource.java
@@ -30,5 +30,9 @@ public class HlsMediaSource {
         public MediaSource createMediaSource(MediaItem mediaItem) {
             return null;
         }
+
+        public Factory setAllowChunklessPreparation(boolean allowChunklessPreparation) {
+            return this;
+        }
     }
 }


KrzysztofMoch (approved):

freeboub (approved):

KrzysztofMoch:
FYI: I have checked if there is any other missing function in stub and there isn't

PR #3953, opened 2024-06-29
refactor: migrate ref methods to command

YangJonghun:
## Summary
- migrate ref methods to command
- add deprecated warning into `getReactTag` function
- remove unused native methods (`presentFullscreenPlayer`, `dismissFullscreenPlayer`)
- wrong type change (drm.getLicense)
  - as-is: `() => void`
  - to-be: `() => string | Promise<string>`

### Motivation
- we need to remove `findNodeHandle` ([link](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/enable-libraries-prerequisites.md))
- Command doesn't allow return value or callback argument. however `save` and `getCurrentPosition` functions return a value. so we have to consider about whether to keep it or drop it.

## Test plan


src/types/ViewType.ts
```diff
@@ -2,10 +2,10 @@
  * Define Available view type for android
  * these values shall match android spec, see ViewType.kt
  */
-enum ResizeMode {
```
freeboub, line 5: 😢

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerViewManager.java
```diff
@@ -82,6 +79,10 @@ public class ReactExoplayerViewManager extends ViewGroupManager<ReactExoplayerVi
     private static final String PROP_DEBUG = "debug";
     private static final String PROP_CONTROLS_STYLES = "controlsStyles";
 
+    private static final String CMD_SET_PLAYER_PAUSE_STATE = "setPlayerPauseState";
+    private static final String CMD_SEEK = "seek";
+    private static final String CMD_SET_VOLUME = "setVolume";
+    private static final String CMD_SET_FULLSCREEN = "setFullscreen";
```
KrzysztofMoch: We have method `setFullScreen` in JS so this one is not working
```suggestion
    private static final String CMD_SET_FULLSCREEN = "setFullScreen";
```

src/specs/VideoNativeComponent.ts
```diff
@@ -373,45 +369,45 @@ export interface VideoNativeProps extends ViewProps {
...
   setLicenseResultError: (
+    viewRef: React.ElementRef<HostComponent<VideoNativeProps>>,
     error: string,
     licenseUrl: string,
-    reactTag: number,
-  ) => Promise<void>;
-  setVolume: (volume: number, reactTag: number) => Promise<void>;
-  getCurrentPosition: (reactTag: number) => Promise<number>;
-  setFullScreen: (fullScreen: boolean, reactTag: number) => Promise<void>;
-}
```
KrzysztofMoch: nit: 
let's move `React.ElementRef<HostComponent<VideoNativeProps>>` into two types
1. `NativeVideoComponentType` for `HostComponent<VideoNativeProps>`
2. `NativeVideoComponentRefType` for `React.ElementRef<NativeVideoComponentType>`
(or you can name it whatever you want)

Just to not repeat it

KrzysztofMoch (changes requested):

KrzysztofMoch:
I will check if [Interop Layer](https://github.com/reactwg/react-native-new-architecture/discussions/135) will work with those changes and I am fine to merge this (after @freeboub review)

src/specs/NativeVideoDecoderProperties.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface NativeVideoDecoderPropertiesSpec extends TurboModule {
+  getWidevineLevel: () => Promise<Int32>;
+  isCodecSupported: (
+    mimeType: string,
+    width: Int32,
+    height: Int32,
+  ) => Promise<'unsupported' | 'hardware' | 'software'>;
+  isHEVCSupported: () => Promise<'unsupported' | 'hardware' | 'software'>;
+}
```
KrzysztofMoch: For some funny reason interfaces that extends `TurboModule` must have be named `Spec` (otherwise codegen is throwing error)
```suggestion
interface Spec extends TurboModule {
  getWidevineLevel: () => Promise<Int32>;
  isCodecSupported: (
    mimeType: string,
    width: Int32,
    height: Int32,
  ) => Promise<'unsupported' | 'hardware' | 'software'>;
  isHEVCSupported: () => Promise<'unsupported' | 'hardware' | 'software'>;
}

export interface NativeVideoDecoderPropertiesSpec extends Spec {}

export default TurboModuleRegistry.getEnforcing<Spec>(
  'RNVDecoderPropertiesModule',
);
```

src/specs/NativeVideoManagerModule.ts
```diff
@@ -0,0 +1,19 @@
...
+} from 'react-native/Libraries/Types/CodegenTypes';
+
+export type VideoSaveData = {
+  uri: string;
+};
+
+export interface NativeVideoManagerModuleSpec extends TurboModule {
+  save: (reactTag: Int32, option: UnsafeObject) => Promise<VideoSaveData>;
+  getCurrentPosition: (reactTag: Int32) => Promise<Int32>;
+}
```
KrzysztofMoch: Same for this one
```suggestion
interface Spec extends TurboModule {
  save: (reactTag: Int32, option: UnsafeObject) => Promise<VideoSaveData>;
  getCurrentPosition: (reactTag: Int32) => Promise<Int32>;
}

export interface NativeVideoManagerModuleSpec extends Spec {}

export default TurboModuleRegistry.getEnforcing<Spec>('VideoManager');
```

src/specs/VideoNativeComponent.ts
```diff
@@ -373,45 +369,45 @@ export interface VideoNativeProps extends ViewProps {
...
-  setPlayerPauseState: (paused: boolean, reactTag: number) => Promise<void>;
+type NativeVideoComponentType = HostComponent<VideoNativeProps>;
+type NativeVideoComponentRefType = React.ElementRef<NativeVideoComponentType>;
+
+export interface NativeVideoCommands {
+  seek: (
+    viewRef: NativeVideoComponentRefType,
+    time: Float,
+    tolerance?: Float,
+  ) => void;
```
KrzysztofMoch: It seems like we need to half-ly revert prev commit - function that are command need to have first arg that is type `React.ElementRef<T>` (even if we created alias it does not work with codegen)
so we need to change `viewRef: NativeVideoComponentRefType` to `viewRef: React.ElementRef<NativeVideoComponentType>`

KrzysztofMoch (changes requested):

src/specs/NativeVideoManagerModule.ts
YangJonghun: fixed it! thanks for checking

KrzysztofMoch:
@YangJonghun It also seems that we need to do some renaming as `setVolume` as Command collides with some other `setVolume` (probably from `volume` prop) - I suggest to rename `setVolume` Command to `setVolumeCMD`. It will be only internal change as we wrap commands in out component

KrzysztofMoch:
I am also wandering if we should use `TurboModuleRegistry.getEnforcing` with old architecture "package" (Turbo modules has it own way to export modules in package - you can find it [here](https://github.com/reactwg/react-native-new-architecture/blob/main/docs/turbo-modules.md)). I have check it and we cannot move to it (new TurboModules package way) yet as we need to firstly move view to fabric. 

All those changes that I am required from you are from my process of trying to build on New Arch - And I think that we cannot break support for Interop Layer as developers can depend on it. So for now it would be better to back to `requireNativeComponent` as with those changes I am not able to run app in new arch anymore.
WDYT @YangJonghun @freeboub ?

YangJonghun:
@KrzysztofMoch 
Thanks for comment!
But AFAIK `TurboModuleRegistry.getEnforcing` works properly on interop layer probably because we didn't set congenConfig.
Also I confirmed that it runs on version 0.74 with new architecture.

YangJonghun:
> @YangJonghun It also seems that we need to do some renaming as setVolume as Command collides with some other setVolume (probably from volume prop) - I suggest to rename setVolume Command to setVolumeCMD. It will be only internal change as we wrap commands in out component

I agree, but I'm a little confused, Am I correct in asking for your suggestion like code below?
```diff
    const setVolume = useCallback((volume: number) => {
-      return nativeRef.current && Commands.setVolume(nativeRef.current, volume);
+     return nativeRef.current && Commands.setVolumeCMD(nativeRef.current, volume);
    }, []);
```

KrzysztofMoch:
> Also I confirmed that it runs on version 0.74 with new architecture.

👀 In our example ? I have tried to run app with react-native-video (built from this PR) but I wasn't able to do it in `basic` example nor "clean" RN app (in both I used `react-native@0.74.3`)

> Am I correct in asking for your suggestion like code below?

Yes this is what I meant

YangJonghun:
> 👀 In our example ? I have tried to run app with react-native-video (built from this PR) but I wasn't able to do it in basic example nor "clean" RN app (in both I used react-native@0.74.3)

No, I tested on my own bare react-native project (created via `npx react-native init`)
I tried with example! Thanks for let me know

KrzysztofMoch:
> No, I tested on my own bare react-native project (created via `npx react-native init`) I tried with example! Thanks to let me know

Strange I also tried to run it on new bare react-native project - it built but there was an error that `VideoManager` Module is missing "in TurboModule list" (as if it was not detected)

YangJonghun:
@KrzysztofMoch 
I confirmed error you mentioned. Thanks
I'll be restore `NativeModules`

YangJonghun:
@KrzysztofMoch 
If we need to support interop layer, I think we need to drop this PR.
New arch doesn't support partial migration.🥲
or create new stacked branch and merge it for prepare to support new arch.

KrzysztofMoch:
This is something that also thought. I think it would be better to create new arch branch - then step by step migrate it 
@freeboub WDYT?

android/src/main/java/com/brentvatne/react/VideoDecoderPropertiesModule.kt
```diff
@@ -67,6 +67,6 @@ class VideoDecoderPropertiesModule(reactContext: ReactApplicationContext?) : Rea
     companion object {
         private val WIDEVINE_UUID = UUID(-0x121074568629b532L, -0x5c37d8232ae2de13L)
         private const val SECURITY_LEVEL_PROPERTY = "securityLevel"
-        private const val REACT_CLASS = "VideoDecoderProperties"
```
freeboub, line 70: is it really necessary to change the module name ?

ios/Video/RCTDecoderPropertiesModule.m
```diff
@@ -0,0 +1,13 @@
+#import "RCTDecoderPropertiesModule.h"
+
+@implementation RCTDecoderPropertiesModule
+
+RCT_EXPORT_MODULE(RNVDecoderPropertiesModule);
+
+RCT_EXPORT_METHOD(getWidevineLevel : (RCTPromiseResolveBlock)resolve reject : (RCTPromiseRejectBlock)reject) {}
```
freeboub, line 7: I think you should al least resolve or reject the Promise... (maybe reject with an error "not implemented") , for these 3 methods

src/Video.tsx
```diff
@@ -615,15 +611,14 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           drm={_drm}
           style={StyleSheet.absoluteFill}
           resizeMode={resizeMode}
-          fullscreen={isFullscreen}
```
freeboub, line 618: This PR is big, can you confirm the fullscreen property is still available for apps ?

freeboub:
> This is something that also thought. I think it would be better to create new arch branch - then step by step migrate it @freeboub WDYT?

Yes It can make sense, but I am really afraid of divergencies... I don't think this PR is too much risky ... but it is still better  to merge steps by steps...

android/src/main/java/com/brentvatne/react/VideoDecoderPropertiesModule.kt
YangJonghun: No, it's not required, but I changed it to reduce the chance of duplicates with other library(or core).

ios/Video/RCTDecoderPropertiesModule.m
YangJonghun: These functions are meaningless but, acceptable.
I think we should throw an error to notify library contributors.
(we should wrapping decoder properties module on JS side)

src/Video.tsx
YangJonghun: previously, we use internal fullscreen state(for handle present,dismiss fullscreen command) but it was meaningless.
fullscreen prop will be inject with `{...rest}` and fullscreen commands are work with command.setFullscreen(value).
previous
Rather, the old code was causing different behavior than expected. when user changing fullscreen prop value.

YangJonghun:
@freeboub @KrzysztofMoch 
Moved some code fixes to PR #3980

PR #3958, opened 2024-06-30
Add react-native-web support

### LINKED ISSUE #2719: React native web will not compile with react-native-video
ERROR in ./node_modules/react-native-video/Video.js 55:7
Module parse failed: Unexpected token (55:7)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|   }
| 
>   seek = (time, tolerance = 100) => {
|     if (isNaN(time)) {throw new Error('Specified time is not a number');}
| 

package.json

"react-native-video": "^5.2.0",

zoriya:
## Summary

This adds web support for RNV. It does not include ads, drm, hls or any fancy player options not supported by an HTML5 player. I think HLS.js would be good to have included by default but if we do, it should be in a FU PR.

For now this PR is in a very draft state, I have not yet tested nor updated the documentation. I'll do it in the following hours/days (I'm currently working from my hospital bed, so do not know how much work I'll do soon).

I made this draft mostly to get preliminary feedback on how I approached this. 

Closes #2719

## Missing steps
 - [x] Test (see testing plan)
 - [x] Write documentation/readme

## Test plan

 - [x] integrate or create a web example
 - [ ] ~~use this in [Kyoo](https://github.com/zoriya/kyoo), I took the base implementation from there and the projects needs HLS and custom handling for subtitles for example.~~ Will do after this is merged and we try to add shaka/video.js or something like that.


shell.nix
```diff
@@ -0,0 +1,12 @@
+{pkgs ? import <nixpkgs> {}}:
```
zoriya, line 1: I can remove this file if you want, it helps having dependencies automatically downloaded when you use nix.

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+        }),
+        [seek, resume, pause],
+      ),
+      showNotificationControls,
+    );
+
+    return (
+      <video
+        ref={nativeRef}
+        src={source?.uri as string | undefined}
```
zoriya: Non URL sources are not supported, I guess there's no way arround it.

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+          });
+        }}
+        onDurationChange={() => {
+          if (!nativeRef.current) {
+            return;
+          }
+          onLoad?.({
+            currentTime: nativeRef.current.currentTime,
+            duration: nativeRef.current.duration,
+            videoTracks: [],
```
zoriya, line 289: HTML5 does not give us tracks info (nor a way to select the current video/audio.

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+          });
+        }}
+        onLoadedData={() => onReadyForDisplay?.()}
+        onError={() => {
+          if (!nativeRef.current?.error) {
+            return;
+          }
+          onError?.({
+            error: {
+              errorString: nativeRef.current.error.message ?? 'Unknown error',
```
zoriya, line 320: I used existing properties, should I create new ones for the web?

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+              code: nativeRef.current.error.code,
+            },
+          });
+        }}
+        onLoadedMetadata={() => {
+          if (source?.startPosition) {
+            seek(source.startPosition / 1000);
```
zoriya: the `/ 1000` is because everything is second based except for the `startPosition` which is in ms.

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+        onPlay={() => onPlaybackStateChanged?.({isPlaying: true})}
+        onPause={() => onPlaybackStateChanged?.({isPlaying: false})}
+        onVolumeChange={() => {
+          if (!nativeRef.current) {
+            return;
+          }
+          onVolumeChange?.({volume: nativeRef.current.volume});
+        }}
+        onEnded={onEnd}
+        style={{position: 'absolute', inset: 0, objectFit: 'contain'}}
```
zoriya: I can't use the StyleSheet api since this is a raw html component. Should I simply add a comment to disable the eslint warning here?

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+      />
+    );
+  },
+);
+
+const useMediaSession = (
+  metadata: VideoMetadata | undefined,
+  nativeRef: RefObject<HTMLVideoElement>,
+  actions: {
```
zoriya: At first I used those actions to trigger events when calling them but I'm thinking the `<video />` events would work too.

I'll remove this and simply use the ref to seek/play/pause.

src/Video.web.tsx
```diff
@@ -0,0 +1,336 @@
...
+      ],
+      [
+        'seekforward',
+        (evt: MediaSessionActionDetails) =>
+          seekRelative(evt.seekOffset ? evt.seekOffset : 10),
+      ],
+      [
+        'seekto',
+        (evt: MediaSessionActionDetails) => actions.seek(evt.seekTime!),
+      ],
```
zoriya: The MediaSessionAPI supports `previoustrack` and `nexttrack` but we don't have properties to expose that right now. Should I add them? (I know it's also possible in android, but I'm not confident in adding it there).

src/types/video-ref.ts
```diff
@@ -0,0 +1,21 @@
...
+  presentFullscreenPlayer: () => void;
+  dismissFullscreenPlayer: () => void;
+  restoreUserInterfaceForPictureInPictureStopCompleted: (
+    restore: boolean,
+  ) => void;
+  save: (options: object) => Promise<VideoSaveData>;
+  setVolume: (volume: number) => void;
+  getCurrentPosition: () => Promise<number>;
+  setFullScreen: (fullScreen: boolean) => void;
+  nativeHtmlRef?: RefObject<HTMLVideoElement>; // web only
```
zoriya: This would be needed for a user to add custom HTML5 libs (like hls.js, video.js or better subtitles handling).

zoriya:
I tested that everything works with the `basic` example app using one of the .mp4 videos available without headers (since both HLS/Dash and headers are not available with a HTML5 basic video player).

I'm thinking that RNV should have more features than the default browsers allow, and integrating HLS/Dash/Ads should not be left for the user.

On my project, I'm using HLS.js, srt-webvtt and jassub to have hls/srt/ass support. Maybe we could find an extensible and not too bloaty library to use as a video backend on the web. I'm not familiar with the state of those libs right now but I have heard of video.js and shaka-player.

What do you think about integrating one of those libraries @freeboub @KrzysztofMoch?

src/types/video-ref.ts
zoriya: I did not document this, if we go into integrating vifdeo.js/shaka/something this would not be needed anymore and then removed.

KrzysztofMoch:
> I'm thinking that RNV should have more features than the default browsers allow, and integrating HLS/Dash/Ads should not be left for the user.

@zoriya I agree with you - but let's do it in next (separate) PR

I will try to review this PR this week!

examples/basic/package.json
```diff
@@ -13,12 +13,15 @@
     "pod-install:newarch": "cd ios && RCT_NEW_ARCH_ENABLED=1 bundle exec pod install && cd .."
   },
   "dependencies": {
+    "@expo/metro-runtime": "~3.2.1",
     "@react-native-picker/picker": "2.7.5",
     "expo": "^51.0.14",
     "expo-asset": "^10.0.9",
     "expo-image": "^1.12.12",
     "react": "18.2.0",
+    "react-dom": "18.2.0",
```
freeboub: I think you should add these react-dom and react-native-web in peerDependancies or optionanal dependancy in the main pakage.json

freeboub: can you also add an option to build / launch web version in the scripts object ?

src/types/video-ref.ts
freeboub: @zoriya Is it a good idea to expose nativeHtmlRef to the UI ? do it bring some additionnal feature .
maybe you can rename it to nativeHtmlVideoRef it will allow to create new values for shaka / video.js, ...

src/specs/VideoNativeComponent.ts
```diff
@@ -266,12 +267,12 @@ type OnReceiveAdEventData = Readonly<{
 
 export type OnVideoErrorData = Readonly<{
   error: Readonly<{
-    errorString?: string; // android
+    errorString?: string; // android | web
```
freeboub, line 288: these changes will bring conflicts ... if you need to resolve it, you can skip these comments I think

shell.nix
freeboub: @KrzysztofMoch what do you think of this ? I agree to keep it ..

src/Video.web.tsx
freeboub: no, but may need a comment in the the doc (even it doesn't really make sense on web I think ... )

src/Video.web.tsx
freeboub: add note in doc :)

src/Video.web.tsx
freeboub: error management shall be reworked ...

src/Video.web.tsx
freeboub: Do you have the issue if you move the style in a const variable ?
I think it would be better

src/Video.web.tsx
freeboub: Good idea, but It would be better in a separated PR ! please keep it for later

shell.nix
KrzysztofMoch: It doesn't mind me either so we can leave it

KrzysztofMoch (commented):
I have tested this in our `basic` example (with expo web) and it's working very well! 
For now are working only `mp4` files but support for other let's leave for later. I will look into code now.
Overall grate job @zoriya 🔥

zoriya:
Thanks, I'll correct review comments & rebase during the night!

examples/basic/package.json
zoriya: Are you sure? The example does need rnw to run in web mode. It won't be pulled by another package.

\+ react-native-windows is declared as a dependency, not a peer/optional dependency.

src/Video.web.tsx
```diff
@@ -0,0 +1,333 @@
...
+      onSeek,
+      onVolumeChange,
+      onEnd,
+      onPlaybackStateChanged,
+    },
+    ref,
+  ) => {
+    const nativeRef = useRef<HTMLVideoElement>(null);
+    const errorHandler = useRef<typeof onError>(onError);
+    errorHandler.current = onError;
```
KrzysztofMoch: What is this for?

src/Video.web.tsx
```diff
@@ -0,0 +1,333 @@
...
+    useImperativeHandle(
+      ref,
+      () => ({
+        seek,
+        pause,
+        resume,
+        setVolume,
+        getCurrentPosition,
+        // making the video fullscreen does not work with some subtitles polyfils
+        // so I decided to not include it.
```
KrzysztofMoch: You mean that some subtitles polyfils are not working when video is in fullscreen ? Users can open fullscreen from UI so I think we can implement those and add warning in docs WDYT?

zoriya: Yes, some subtitles polyfill use a canvas or other HTML elements to draw on top of the video, and those do not work if we simply full-screen the video.

I was thinking of implementing the full-screen feature after adding the polyfill, since it's possible to fullscreen a container of both the video and the polyfill.

src/Video.web.tsx
zoriya: Good catch, it's not used anymore. Removed it

src/Video.web.tsx
KrzysztofMoch: I understand, but can we for now implement "built-in" fullscreen? We can later edit it to support subtitles polyfills

zoriya: Kay, i'll add that

KrzysztofMoch: Thanks!

zoriya: Done, I put only the video element on fullscreen tho so controls are not shown in fullscreen.

On my website, I put the whole body in fullscreen, but that might not be doable for some apps. Do you have an idea how to improve that?

KrzysztofMoch: I was thinking about using this https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen

zoriya: Yes, that's what I used. Controls can't be shown using this tho.

KrzysztofMoch: https://github.com/TheWidlarzGroup/react-native-video/assets/83216158/2f45c2b2-5dea-4dc6-85fd-5805d9ba0fd7

I have checked it and it's working

zoriya: Built-in (HTML5 defaults) controls will show but not custom ones

zoriya: I see three ways of fixing this:
- have a props for a ref of the item that needs to be in fullscreen (could be a ref to document.body)
- have controls as child of the Video element
- add an html id/class `player-controls` and we could find the closest parent with this class to fullscreen

The second one is the most "react-like" but it kinda breaks expectation of having a simple Video element + it's a breaking.

The other two feels more like escape hatch.

What does the fullscreen behavior does on android/ios? I must say I never used it and always used fullscreen on the whole route via the router.

KrzysztofMoch: Fullscreen on iOS/Android also doesn't allow custom controls so it's not a problem (maybe in future we will add it)

examples/basic/package.json
freeboub: sorry, it is exemple package.json ... OK so

freeboub:
@zoriya I still 2 documentation comments to handle. + conflicts to fix . And it is good on my side.
@KrzysztofMoch not sure about timeline to merge this PR ?!

src/Video.web.tsx
zoriya: Already done: https://github.com/TheWidlarzGroup/react-native-video/pull/3958/files#diff-ea60cfe144c95a06a3ecdc85bf16b43620ba5711bfe2d01ea18d69da24050445R223

src/Video.web.tsx
zoriya: Added: https://github.com/TheWidlarzGroup/react-native-video/pull/3958/files#diff-e2131d0631cdc1ddf0ba36827a27b86adf8e5b8ed162a280a53c6219f4df5a5fR643

src/types/video-ref.ts
zoriya: Not sure where to add documentation for this, maybe in the method page? Since it's the page that talk about the video ref?

zoriya: I added it on the method page

src/types/index.ts
```diff
@@ -7,4 +7,4 @@ export {default as ResizeMode} from './ResizeMode';
 export {default as TextTrackType} from './TextTrackType';
 export {default as ViewType} from './ViewType';
 export * from './video';
-export * from '../specs/VideoNativeComponent';
```
zoriya, line 10: I think this was only used for event types so I removed it and added a

```typescript
export type * from '../spec/VideoNativeComponent';
```

in the `events.ts` file. I assumed the default export was not used, if it's used I can restore it like before and add a file that export a dummy object for the web.

zoriya:
I rebased & added all new properties. Everything should be documented, I missed the `nativeVideoHtmlElement` and added it, other doc comments should be fixed.

Ramon-Balaguer:
@zoriya How is this feature?

zoriya:
what does this mean? The PR just needs to be reviewed again/approved if that was the question

mrsantran:
merge this PR please :)

arijitsarkar2k14:
Any plans to merge this PR soon?

zoriya:
They said in #4114 that they have limited time this week. Give them some time.

\+ this pr only adds basic web supports, once this is merged i plan on adding hls & subtitles support properly. I probably won't add ads or drm support tho.

ananjaemin:
Is this going to be merged?

colonelpanic8:
@zoriya any sense of the status here. I'd be happy to help with hls support!

zoriya:
the pr is ready, it needs to be reviewed (and updated to master). I'll rebase to master as soon as it gets some traction from TWG side (don't want to keep rebasing for no reason).

jieey1140:
@zoriya  I hope this work gets merged as soon as possible!

ananjaemin:
@KrzysztofMoch @freeboub Please review it

src/Video.web.tsx
```diff
@@ -0,0 +1,419 @@
...
+        if (isNaN(time)) {
+          throw new Error('Specified time is not a number');
+        }
+        if (!nativeRef.current) {
+          console.warn('Video Component is not mounted');
+          return;
+        }
+        time = Math.max(0, Math.min(time, nativeRef.current.duration));
+        nativeRef.current.currentTime = time;
+        onSeek?.({seekTime: time, currentTime: nativeRef.current.currentTime});
```
LunatiqueCoder, line 70:
```suggestion
        time = Math.max(0, Math.min(time, nativeRef.current.duration));
        
        if (isFinite(time) {
           nativeRef.current.currentTime = time;
           onSeek?.({seekTime: time, currentTime: nativeRef.current.currentTime});
        }
```

<br>

### Fixes error:

<br>
Google Chrome 2024-11-03 08 57 44

LunatiqueCoder:
# 🎉🍾🥳 

We tried this branch in our demo web app that we're preparing at English Discoveries, along with [react-native-media-console](https://github.com/LunatiqueCoder/react-native-media-console) and it works like a charm! 

Here's a screenshot:

Google Chrome 2024-11-03 08 53 25

LunatiqueCoder (approved):

src/Video.web.tsx
zoriya: isn't this on you? I'd expect it to fail if you pass NaN or Infinity

LunatiqueCoder: hmm yeah actually -.- though it doesn't crash on the mobile side

src/Video.web.tsx
```diff
@@ -0,0 +1,419 @@
...
+          })
+        }
+        onPause={() =>
+          onPlaybackStateChanged?.({
+            isPlaying: false,
+            isSeeking: isSeeking.current,
+          })
+        }
+        onSeeking={() => (isSeeking.current = true)}
+        onSeeked={() => (isSeeking.current = false)}
```
KrzysztofMoch: Should we call `onSeek` callback here ?

KrzysztofMoch (changes requested):
This implementation it's fine for me 👍
@zoriya if you could resolve conflicts and add missing props/methods (eg. `setSource`), then @moskalakamil will review docs part and we will be ready to merge. Please note that we recently reworked examples so we need add web support to `examples/expo`

src/Video.web.tsx
zoriya: good catch

zoriya:
I did the preliminary rebase, I'll re-add an example, add new properties/event & fix potential bugs later today.

src/Video.web.tsx
```diff
@@ -0,0 +1,449 @@
...
+
+    // Stock this in a ref to not invalidate memoization when those changes.
+    const fsPrefs = useRef({
+      fullscreenAutorotate,
+      fullscreenOrientation,
+    });
+    fsPrefs.current = {
+      fullscreenOrientation,
+      fullscreenAutorotate,
+    };
```
moskalakamil, line 125: Why do we need these lines? We are setting default values in useRef

src/Video.web.tsx
```diff
@@ -0,0 +1,449 @@
...
+                await document.exitFullscreen();
+              }
+              screen.orientation.unlock();
+            }
+          } catch (e) {
+            // Changing fullscreen status without a button click is not allowed so it throws.
+            // Some browsers also used to throw when locking screen orientation was not supported.
+            console.error('Could not toggle fullscreen/screen lock status', e);
+          }
+        };
```
moskalakamil: Do we need to create this new function under the setFullscreen fn? Actually, I think we could pass this code directly there

src/Video.web.tsx
```diff
@@ -0,0 +1,449 @@
...
+
+    useEffect(() => {
+      if (paused) {
+        pause();
+      } else {
+        resume();
+      }
+    }, [paused, pause, resume]);
+    useEffect(() => {
+      if (volume === undefined) {
```
moskalakamil: Maybe !isNaN(volume) would be better

zoriya:
What's the expected behavior when using `setSource` and the `source` prop is defined/changes?
Should using `setSource` disable the source prop? Should 'setSource` only work if `source` is undefined?

src/Video.web.tsx
zoriya: the command just above it explains it, we want to access those values in the setFullcreen callback, but we don't want to invalidate the callback when either `fullscreenOrientation` or `fullcreenAutorotate` change.

We need to assign to the ref at every render because we want to keep the ref in sync with the props, if the props change the next call to setFullcreen should use the updated props.

src/Video.web.tsx
zoriya: that's true, i got used to doing that for `useEffect` x)

src/Video.web.tsx
zoriya: I added a isNaN check but we need te keep the undefined check

zoriya:
Except the `setSource` method that doesn't handle metadata nor interactions with the `source` prop (see my previous comment), this should be ready.

Once this get merged, I'll make follow-up PR for hls & subtitles while I transition kyoo to use this for the web.

src/Video.web.tsx
moskalakamil: and could we move it to the useEffect?

src/Video.web.tsx
moskalakamil: thats right

src/Video.web.tsx
zoriya: there is no need to useEffect here. The useRef doesn't trigger a re-render.

KrzysztofMoch:
> What's the expected behavior when using setSource and the source prop is defined/changes?
> Should using setSource disable the source prop? Should 'setSourceonly work ifsource` is undefined?

So this is good question 🤔 on mobile we do re-init on source change. `setSource` is overriding source prop but if source prop would be updated it will override source from `setSource`

So on web I guess with should do something like this ?
```jsx
const [source, setSource] = React.useState(props.source)

useEffect(() => {
  setSource(props.source)
}, [props.source])

// and add method to ref
```
WDYT @zoriya ?

zoriya:
This is assuming that `source` is stable (aka wrapped in a `useMemo`). Without this, each re-render (be it another prop change or a non-relevant `setState`) would reset the `setSource`.

I am fine with this behavior, if we decide to go with this. I'll also add it in the documentation.

KrzysztofMoch:
Yeah you're right! but do we have any alternative? (idk if there is, just thinking loudly)

KrzysztofMoch:
Maybe we could use [`React.memo`](https://react.dev/reference/react/memo) with custom `arePropsEqual` function?

zoriya:
No, I don't think there's a good alternative available. 
We could hold a ref to the old source and deep compare the objects, but that felt so wrong that I decided to ask what was the expected behavior.

Something like:
```js
const [src, setSource] = useState(source)
const oldSource = useMemo(source);
useEffect(() => {
	if (deepEqual(oldSource.current, source)) 
		return;
	setSource(source);
	oldSource.current = source;
}, [source]);

// add setSource to ref
```

With `deepEqual` being a function from something like [here](https://stackoverflow.com/questions/25456013/javascript-deepequal-comparison).

KrzysztofMoch:
Yeah this looks kinda evil 😅 But I think this will work best, also I would rename `oldSource` to `currentSource` (and here you are using `useMemo` but I guess you wanted to use `useRef`) - we can go with it

zoriya:
oh yeah i meant `useRef`. I'll push that today then x)

KrzysztofMoch (approved):
Thank you very much!
@moskalakamil can you look at this and test it? If you will be fine you can merge

moskalakamil:
Okay, I will check it later today. Thank you so much, @zoriya. Amazing work!

moskalakamil:
@zoriya We need to update constants/general.ts in the bare example as well, because the Expo example is overridden in the `src` folder every time we install node modules (I mean this lines: 
```
metadata: {
      title: 'Custom Title',
      subtitle: 'Custom Subtitle',
      artist: 'Custom Artist',
      description: 'Custom Description',
      imageUri:
        'https://pbs.twimg.com/profile_images/1498641868397191170/6qW2XkuI_400x400.png',
}
```

moskalakamil:
LGTM! Amazing work @zoriya 🔥

freeboub:
@zoriya I misse to thank you for this PR, so Thank you very much !

colonelpanic8:
@zoriya I vaguely remember that you mentioned you were planning to work on hls support.

I actually have a version of this working, but its kind of sloppy, but I'm wondering if you had made any inroads with that yourself. Might be helpful for us to coordinate.

zoriya:
I wanted to wait for #4255 to get merged before creating a new branch to prevent huge conflicts. I do already have hls support on kyoo which was the implementation I based this PR on ([here](https://github.com/zoriya/Kyoo/blob/master/front/packages/ui/src/player/video.web.tsx)).

KrzysztofMoch:
I need to fix expo example to merge it (but no idea how to do it for now)

colonelpanic8:
@zoriya okay interesting. doesn't look like that applies super cleanly on what is there though. What was the reasoning behind using hls.js instead of videojs or shaka? Seems like supporting dash if its possible might be a good thing.

zoriya:
I personally didn't need more than hls.js, for rnv we would indeed need to consider alternatives before.

PR #3961, opened 2024-07-01
feat(fabric): updated ios podspec for implementing new architecture

yungblud:
## Summary
- updated react-native-video.podspec as pre-setup (small task) for implemeting New Architecture on react-native-video ios

### Motivation
- Trying to implement new architecture to react-native-video module on iOS

### Changes
- updated react-native-video.podspec

## Test plan
- `pod install` (in the examples apps including FabricExample)
- and run build from xcode (basic example & FabricExample)

### DIFF
--- a/examples/FabricExample/Gemfile.lock
+++ b/examples/FabricExample/Gemfile.lock
@@ -5,11 +5,12 @@ GEM
       base64
       nkf
       rexml
-    activesupport (7.0.4.3)
+    activesupport (6.1.7.8)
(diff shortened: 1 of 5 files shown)


freeboub (approved):

KrzysztofMoch (approved):
LGTM
Btw it would be nice if could stay with swift - I know that new architecture is not supporting it but we cloud try to bridge it via tiny ObjC layer. It would be easier for us to maintain and PR for iOS would be easier to review and merge

PR #3963, opened 2024-07-02
fix: build issue on the latest react-native version

seyedmostafahasani:
## Test plan


seyedmostafahasani:
@YangJonghun I updated the `ReactVideoPackage.kt` file. I tested it with the latest react-native version, and everything works correctly. Thank you again for testing with the latest version.

android/src/main/java/com/brentvatne/react/ReactVideoPackage.kt
```diff
@@ -9,7 +9,9 @@ import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.uimanager.ViewManager
 
-class ReactVideoPackage @JvmOverloads constructor(private var config: ReactExoplayerConfig? = null) : ReactPackage {
+class ReactVideoPackage(private var config: ReactExoplayerConfig?) : ReactPackage {
```
YangJonghun: nit;
I don't think we need to define config as `var`🤔

KrzysztofMoch (approved):

freeboub (approved):

PR #3969, opened 2024-07-03
chore(android): migrate FullScreenPlayerView to Kotlin

seyedmostafahasani:
## Summary
Migrate FullScreenPlayerView to Kotlin

### Motivation

## Test plan


android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -0,0 +1,124 @@
...
+        companion object {
+            private const val UPDATE_KEEP_SCREEN_ON_FLAG_MS = 200L
+        }
+    }
+
+    init {
+        setContentView(containerView, generateDefaultLayoutParams())
+    }
+
+    override fun onBackPressed() {
```
YangJonghun, line 63: nit;
It's not point of this PR, onBackPressed was deprecated🥲 so, it would be nice to change it.

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -0,0 +1,124 @@
...
+        parent = null
+    }
+
+    private fun updateFullscreenButton(iconResourceId: Int, contentDescription: String) {
+        playerControlView?.let {
+            val imageButton: ImageButton = it.findViewById(com.brentvatne.react.R.id.exo_fullscreen)
+            imageButton.setImageResource(iconResourceId)
+            imageButton.contentDescription = contentDescription
+        }
+    }
```
YangJonghun, line 114: nit;
How about below codes?
```kotlin
    private fun updateFullscreenButton(playerControlView: LegacyPlayerControlView, isFullscreen: Boolean) {
            val imageButton = playerControlView.findViewById<ImageButton?>(com.brentvatne.react.R.id.exo_fullscreen)
            imageButton?.let {
                val imgResource = if (isFullscreen) androidx.media3.ui.R.drawable.exo_icon_fullscreen_exit else androidx.media3.ui.R.drawable.exo_icon_fullscreen_enter
                val desc = if (isFullscreen) {
                    context.getString(androidx.media3.ui.R.string.exo_controls_fullscreen_exit_description)
                } else {
                    context.getString(androidx.media3.ui.R.string.exo_controls_fullscreen_enter_description)
                }
                imageButton.setImageResource(imgResource)
                imageButton.contentDescription = desc
            }
    }
```

seyedmostafahasani: I agree with you.

seyedmostafahasani:
@YangJonghun thank you for reviewing the code. I have applied your comments. Please check them out.

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -0,0 +1,136 @@
...
+        super.onAttachedToWindow()
+        if (reactExoplayerView.preventsDisplaySleepDuringVideoPlayback) {
+            mKeepScreenOnHandler.post(mKeepScreenOnUpdater)
+        }
+        onBackPressedDispatcher.addCallback(onBackPressedCallback)
+    }
+
+    override fun onDetachedFromWindow() {
+        super.onDetachedFromWindow()
+        onBackPressedCallback.handleOnBackPressed()
```
YangJonghun: Is this function call really necessary?
(Everything else looks good to me)

seyedmostafahasani: @YangJonghun yes, it's necessary because when the user presses the back button, we need to handle this logic. We can move it to the `onStop` function. What do you think?

YangJonghun: I don't think this is a good implementation.
OnBackPressedDispatcher should behave according to lifecycle of Activity or Fragment.
That's why BackPress Callback currently also works on fullscreen icon clicks.
To implement this properly, it looks like you'll either have to use DialogFragment or use internal Android API. Since this implementation was not purpose of this PR, So it seems like a good idea to revert it.

seyedmostafahasani: So, should I either use the deprecated function `onBackPressed` or call this logic in the `onStop` function?

YangJonghun: For now, it seems like a better idea to use deprecated onBackPressed, so I suggest it.

YangJonghun (approved):
LGTM Thanks

seyedmostafahasani:
Thank you for reviewing!

KrzysztofMoch (approved):

freeboub (approved):

PR #3972, opened 2024-07-04
feat: add ability to define `poster` props as Image type and render poster as custom component

### LINKED ISSUE #3013: Ability to set header or custom component for poster
# Feature Request

## Why it is needed

Constantly the same posters can be boring in dynamic applications, so a structure like on source for posters or structures where we can show our custom own components as posters will be nice and simple to use. Long story short :D
It would be great if we can switch to a structure where we can pass header just like on source

### Code sample
```

<Video
          uri: podCasturl,//here is the,
          headers: {
            access_token: accessToken, // token for reach our api
          },
        }}
        poster={{
          uri: podCasturl,//here is the,
          headers: {
            access_token: accessToken, // token for reach our api
          },
        }}
        posterResizeMode="center"
      />
```
or 

```

const renderCustomPoster = () => {
    return (<Image />) // here custom images how developer needs
}
<Video
          uri: podCasturl,//here is the,
          headers: {
            access_token: accessToken, // token for reach our api
          },
        }}
        poster={renderCustomPoster()}
        posterResizeMode="center"
      />
```

moskalakamil:
## Summary

I added the ability for the user to fully manage the poster, including: managing all props, rendering their own poster, and fixed a bug with displaying the poster

Closes #3013

### Motivation

[Issue](https://github.com/TheWidlarzGroup/react-native-video/issues/3013)

### Changes
- Added `posterProps` to allow user manage all of the poster props (incl. source, resizeMode and all others)
- Added `renderPoster` props, to allow user render poster as a component
- Update docs
- Fixed bug -> When the video wasn't fully loaded but its image was visible, both the video and the poster were simultaneously visible.
Screenshot 2024-07-04 at 10 55 44

## Test plan

Run basic example ;)


docs/pages/component/props.mdx
```diff
@@ -431,26 +431,39 @@ Determine whether the media should continue playing when notifications or the Co
...
+Value: Props for the `Image` component. The poster is visible when the source attribute is provided.
 
-### `posterResizeMode`
+```javascript
+<Video>
+  posterProps={{
+    source: { uri: "https://baconmockup.com/300/200/" },
+    resizeMode: "cover",
+    // ...
+  }}
```
KrzysztofMoch: let's re-name it to back to `poster` to don't introduce breaking change - also we need to handle "old way" 
so type should be `string | ImageProps` and if it's string it should be passed to poster source if not then pass as props

src/types/video.ts
```diff
@@ -240,12 +246,12 @@ export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
   pictureInPicture?: boolean; // iOS
   playInBackground?: boolean;
   playWhenInactive?: boolean; // iOS
-  poster?: string;
-  posterResizeMode?: EnumValues<PosterResizeModeType>;
```
KrzysztofMoch, line 247: Let's mark this as deprecated but let's keep support it for a while (so if this is passed it should work)

KrzysztofMoch (changes requested):
Overall looks good but let's make it so we don't introduce breaking changes

moskalakamil:
@KrzysztofMoch Thanks for the review. It should look ok now :))

docs/pages/component/props.mdx
```diff
@@ -437,10 +437,23 @@ Determine whether the media should continue playing when notifications or the Co
 
 An image to display while the video is loading
 
-Value: string with a URL for the poster, e.g. "https://baconmockup.com/300/200/"
+Value: Props for the `Image` component. The poster is visible when the source attribute is provided.
+> [!WARNING]
+> Value: string with a URL for the poster, use `poster` as object instead
```
KrzysztofMoch:
```suggestion
> Value: string with a URL for the poster is deprecated, use `poster` as object instead
```

KrzysztofMoch (approved):

docs/pages/component/props.mdx
```diff
@@ -437,10 +437,23 @@ Determine whether the media should continue playing when notifications or the Co
 
 An image to display while the video is loading
 
-Value: string with a URL for the poster, e.g. "https://baconmockup.com/300/200/"
+Value: Props for the `Image` component. The poster is visible when the source attribute is provided.
+> [!WARNING]
```
freeboub: Can you please move the Warning under the title to keep doc uniform in doc

docs/pages/component/props.mdx
```diff
@@ -452,6 +465,22 @@ Determines how to resize the poster image when the frame doesn't match the raw v
 - **"repeat"** - Repeat the image to cover the frame of the view. The image will keep its size and aspect ratio. (iOS only)
 - **"stretch"** - Scale width and height independently, This may change the aspect ratio of the src.
 
+### `renderPoster`
```
freeboub: Don't you think it should be renderLoader or renderLoaderComponent.
I think the goal is to use animations in this component !
I think we should break the naming here to avoid confusions ...

examples/basic/src/VideoPlayer.tsx
```diff
@@ -246,7 +246,12 @@ const VideoPlayer: FC<Props> = ({}) => {
               cacheSizeMB: state.useCache ? 200 : 0,
             }}
             preventsDisplaySleepDuringVideoPlayback={true}
-            poster={state.poster}
+            poster={{
```
freeboub: I think we should use the new prop/function in the sample ! it would be better

docs/pages/component/props.mdx
moskalakamil: Sure ;))

examples/basic/src/VideoPlayer.tsx
moskalakamil: You mean `renderPoster` (or renderLoader which will be the final one)?

src/Video.tsx
```diff
@@ -606,14 +610,56 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         : ViewType.SURFACE;
     }, [drm, useSecureView, useTextureView, viewType]);
 
+    const posterStyle = useMemo<StyleProp<ImageStyle>>(() => {
```
freeboub: is it possible to move the content of this memo inside _renderPoster ?
As it is used only in _renderPoster I think it can avoid unnecessary computations

src/Video.tsx
```diff
@@ -606,14 +610,56 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
       <View style={style}>
         <NativeVideoComponent
           ref={nativeRef}
           {...rest}
           src={src}
           drm={_drm}
-          style={StyleSheet.absoluteFill}
+          style={{
+            ...StyleSheet.absoluteFillObject,
+            ...(showPoster ? {display: 'none'} : {}),
```
freeboub: Is this  mandatory ? here on each render, you will regenerate the style, then   <NativeVideoComponent will be rerendered too much ...

src/types/video.ts
```diff
@@ -33,6 +41,13 @@ export type ReactVideoSource = Readonly<
   }
 >;
 
+export type ReactVideoPosterSource = ImageURISource | ImageRequireSource;
```
freeboub, line 45: poster must be a url, require doesn't work, Are you sure of this change ?

freeboub (changes requested):
few comments, but this is a good feature ! Thank you !

moskalakamil:
> few comments, but this is a good feature ! Thank you !

Thanks for review! I'll refactor the code

examples/basic/src/VideoPlayer.tsx
freeboub: > You mean `renderPoster` (or renderLoader which will be the final one)?

RenderLoader ! :)

src/types/video.ts
```diff
@@ -240,12 +255,13 @@ export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
   pictureInPicture?: boolean; // iOS
   playInBackground?: boolean;
   playWhenInactive?: boolean; // iOS
-  poster?: string;
-  posterResizeMode?: EnumValues<PosterResizeModeType>;
+  poster?: string | ReactVideoPoster; // string is deprecated
```
freeboub, line 261: can you also mark deprescated with /** @deprecated */ like in this commit: https://github.com/TheWidlarzGroup/react-native-video/pull/3867/commits/731d1e77b18a388c12823eab711b117a6fce5e1e

src/Video.tsx
moskalakamil: @freeboub This line fixing the issue mentioned in the pr description, maybe i can give it inside useMemo, unless u have another idea?

src/types/video.ts
moskalakamil: There is something wrong in our example, that's why `require` doesn't work. We have similar issue, when we add  video souce as require (of course only in our example). I linked my changes to local project and it works well

seyedmostafahasani:
@moskalakamil please update the type of `poster` according to your changes in the example app.

moskalakamil:
> @moskalakamil please update the type of `poster` according to your changes in the example app.

Hi @seyedmostafahasani! I'm not sure what do you mean, can u explain more? I added new type for `poster` props in example app, and i'll change to `renderLoader` as freeboub said

src/Video.tsx
```diff
@@ -606,14 +613,102 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+          );
+          return;
+        }
+
+        return {
+          memoComp: (
+            <View style={StyleSheet.absoluteFill}>
+              <Loader />
+            </View>
+          ),
```
moskalakamil: This code enable to use memoized component (React.memo(Comp)) as renderLoader={MemoComp} (earlier possible only renderLoader={< MemoComp />})

src/types/video.ts
moskalakamil: at `poster` it's impossible to handle deprecated only for one type. I changed it for `posterResizeMode`

moskalakamil:
I adde also warnings when someone is using `renderLoader` in wrong way

docs/pages/component/props.mdx
```diff
@@ -452,6 +465,22 @@ Determines how to resize the poster image when the frame doesn't match the raw v
 - **"repeat"** - Repeat the image to cover the frame of the view. The image will keep its size and aspect ratio. (iOS only)
 - **"stretch"** - Scale width and height independently, This may change the aspect ratio of the src.
 
+### `renderLoader`
```
freeboub: please keep props in alphabetical order

seyedmostafahasani:
> > @moskalakamil please update the type of `poster` according to your changes in the example app.
> 
> Hi @seyedmostafahasani! I'm not sure what do you mean, can u explain more? I added new type for `poster` props in example app, and i'll change to `renderLoader` as freeboub said

Hey, I was suggestion to update type of `poster` variable, but I noticed you replaced it with a new variable.

src/Video.tsx
KrzysztofMoch: This looks quite impressive but let's just change type of renderLoader to `React.ReactNode` and it should be fine I guess without this code 😅

docs/pages/component/props.mdx
```diff
@@ -486,6 +499,22 @@ Speed at which the media should play.
 - **1.0** - Play at normal speed (default)
 - **Other values** - Slow down or speed up playback
 
+### `renderLoader`
+
+<PlatformsList types={['All']} />
+
+Allows you to create custom components to display while the video is loading. If both `renderLoader` and `poster` are provided, `poster` will be ignored.
```
freeboub:
```suggestion
Allows you to create custom components to display while the video is loading. If `renderLoader` is provided, `poster` and `posterResizeMode` will be ignored.
```

Minor, for more clarity, please double `posterResizeMode` is the good naming 😅

src/Video.tsx
```diff
@@ -606,14 +611,75 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+
     return (
       <View style={style}>
         <NativeVideoComponent
           ref={nativeRef}
           {...rest}
           src={src}
           drm={_drm}
-          style={StyleSheet.absoluteFill}
+          style={_style}
```
freeboub, line 682: But here you put display='none' on the video view ? It is weird ?

src/Video.tsx
```diff
@@ -688,9 +754,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           }
           viewType={_viewType}
         />
-        {hasPoster && showPoster ? (
-          <Image style={posterStyle} source={{uri: poster}} />
-        ) : null}
+        {hasPoster && showPoster ? _renderPoster() : null}
```
freeboub: I think ideally we should replace this line by `_renderPoster?.()` you need to move has poster && showPoster inside the useCallback !

src/Video.tsx
freeboub: Then _style would not be needed anymore !

freeboub (changes requested):

src/Video.tsx
moskalakamil: I added `display: none` to prevent an issue that occurs when someone uses the poster. Sometimes, there is a situation where the video isn't fully loaded (it doesn't trigger `onReadyForDisplay`), causing a brief moment where both the first video slide and the poster are visible simultaneously, which shouldn't happen. An alternative approach could be to hide the poster in the `onVideoLoad` function instead of `_onReadyForDisplay`. However, this has its drawback as well, because the poster is visible when the video isn't ready for viewing yet (although it is loaded).
Let me please know @freeboub which option is better for us

freeboub: OK, I understand (and I already fight with that issue).
please fix conflicts from this PR, I will test it myself.
Thank you

freeboub:
code changes looks good to me

moskalakamil:
@freeboub @KrzysztofMoch conflicts resolved 👍

freeboub (approved):

KrzysztofMoch (approved):
LGTM ✅ but there are conflicts again 😅

moskalakamil:
We need to be faster 😅 I resolved them @KrzysztofMoch

KrzysztofMoch:
Waiting for green CI and I will merge

moskalakamil:
Thanks 🚀

PR #3973, opened 2024-07-04
fix(ios): don't pause playback when entering background

KrzysztofMoch:
## Summary
Fix condition in order to don't pause playback when entering background

### Motivation
Fix entering background behaviour

## Test plan
- run app with Video, that has prop `playInBackground`
- go to home screen
- playback should be still playing

### DIFF
--- a/ios/Video/Features/RCTPictureInPicture.swift
+++ b/ios/Video/Features/RCTPictureInPicture.swift
@@ -6,12 +6,12 @@ import React
 
 #if os(iOS)
     class RCTPictureInPicture: NSObject, AVPictureInPictureControllerDelegate {
+        public private(set) var _pipController: AVPictureInPictureController?
         private var _onPictureInPictureEnter: (() -> Void)?
         private var _onPictureInPictureExit: (() -> Void)?
         private var _onRestoreUserInterfaceForPictureInPictureStop: (() -> Void)?
         private var _restoreUserInterfaceForPIPStopCompletionHandler: ((Bool) -> Void)?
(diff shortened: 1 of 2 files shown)


freeboub (approved):

YangJonghun:
I check it's fixed!
However, this code may force PIP to terminate on background with below code
```
        playInBackground={true}
        pictureInPicture={true}
```
I think we need to add conditional statement to check if the PIP is activated.
Also, if PIP is shut down on background, we need to check playInBackground state and handle it

KrzysztofMoch:
@YangJonghun something like this? cc92124514d63cfed51640b33ecdbbb94c2455fa

YangJonghun:
@KrzysztofMoch 
Yes, that's what I mean. But we need to validate that the app is in the background when PIP is turned off. If it is foreground, we have a problem.
Also, I think `_pipController.isPictureInPictureActive` would be more accurate than `_isActive`, but that code seems fine for now.
Thanks:)

KrzysztofMoch:
@YangJonghun Thanks for checking - I applied your suggestion to use `_pipController.isPictureInPictureActive`

YangJonghun:
@KrzysztofMoch Looks good to me your rewritten condition statement.
But please also add playInBackground handling for when the PIP is turned off that I mentioned.

KrzysztofMoch:
@YangJonghun Done ✅

YangJonghun (approved):

YangJonghun:
@KrzysztofMoch 
Thanks :)

PR #3978, opened 2024-07-06
feat!(android): exit fullscreen mode after finishing video playback

seyedmostafahasani:
## Summary
We don't need to exit fullscreen mode after finishing video playback because the user might be using repeat mode. This way, they won't need to interact with their phone after the video finishes.

### Motivation
Fix this issue [3703](https://github.com/TheWidlarzGroup/react-native-video/issues/3703).

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1278,9 +1278,6 @@ public class ReactExoplayerView extends FrameLayout implements
     }
 
     private void onStopPlayback() {
-        if (isFullscreen) {
-            setFullscreen(false);
-        }
         audioManager.abandonAudioFocus(audioFocusChangeListener);
     }
 


KrzysztofMoch (commented):
I have few questions:
- Do we have same behaviour on iOS ?
- What is default behaviour on android ?
- This can be breaking behaviour (I guess, but not sure) so maybe we should exit when `repeat=false`

seyedmostafahasani:
@KrzysztofMoch
About the first question: On iOS, after finishing the video it doesn't exit fullscreen mode.
For the second question: The default behavior on Android depends on the player. I checked with different players. One of them exits fullscreen mode after finishing, while another doesn't.
Third question: I think we can apply this change on Android because we should have the same behavior on both OS (iOS and Android).
Imagine a user wants to see a list of movies. The user start watching the first movie in fullscreen mode. When the first video finishes, another video starts immediately, but the player exits fullscreen mode. As a result, the user has to activate fullscreen mode again.

Finally, I think it's better to not exit fullscreen mode after finishing video playback.
wdyt?

freeboub:
Good explanation which justify the change. Agreed with @KrzysztofMoch , this behavior change is a breaking behavior it should be warned in release note !

freeboub (approved):

KrzysztofMoch:
I would like to change this to 
- if repeat is `false` we exit fullscreen after video finish
- if repeat is `true` we don't exit fullscreen 

WDYT ? @seyedmostafahasani

seyedmostafahasani:
@KrzysztofMoch I agree with you. 

What do you think about the point below?

Imagine a user wants to see a list of movies. The user start watching the first movie in fullscreen mode. When the first video finishes, another video starts immediately, but the player exits fullscreen mode. As a result, the user has to activate fullscreen mode again.

KrzysztofMoch (approved):
Okay fair point

seyedmostafahasani:
@KrzysztofMoch I think we don't need to exit full-screen mode automatically. I guess the user can exit it if they want to.

PR #3985, opened 2024-07-09
chore(android): migrate AspectRatioFrameLayout to Kotlin

seyedmostafahasani:
## Summary
migrate AspectRatioFrameLayout to Kotlin

### Motivation

## Test plan


KrzysztofMoch (approved):

android/src/main/java/com/brentvatne/exoplayer/AspectRatioFrameLayout.kt
YangJonghun, line 1: How about replacing functions that start with get/set with kotlin getter setters?
Also, simple getters are auto-generated in Kotlin, so I don't think you need to implement them.

freeboub: Yep, I think getResizeMode & getAspectRatio are useless.
For the setter, I agree it can simplify the code

freeboub (approved):

android/src/main/java/com/brentvatne/exoplayer/AspectRatioFrameLayout.kt
```diff
@@ -21,50 +21,27 @@ class AspectRatioFrameLayout(context: Context) : FrameLayout(context) {
...
-     *
-     * @param widthHeightRatio The width to height ratio.
-     */
-    fun setAspectRatio(widthHeightRatio: Float) {
-        if (videoAspectRatio != widthHeightRatio) {
-            videoAspectRatio = widthHeightRatio
-            requestLayout()
+    var videoAspectRatio: Float = 0f
+        set(value) {
+            if (value != videoAspectRatio) {
```
YangJonghun: nit;
I recommend using `field` instead of `videoAspectRatio` here because having a custom getter can lead to side effects, But it is not currently an issue.

YangJonghun (approved):

android/src/main/java/com/brentvatne/exoplayer/AspectRatioFrameLayout.kt
seyedmostafahasani: @YangJonghun  I agree with you about the side effects. I have applied your suggestion.

YangJonghun (approved):

PR #3989, opened 2024-07-10
fix(example/basic): select resizeMode

seyedmostafahasani:
## Summary
Handle resizeMode

### Motivation
Fix a bug with `resizeMode` in the basic app for current testing.

## Test plan

### DIFF
--- a/examples/basic/src/components/Overlay.tsx
+++ b/examples/basic/src/components/Overlay.tsx
@@ -14,6 +14,7 @@ import MultiValueControl, {
 } from '../MultiValueControl.tsx';
 import {
   AudioTrack,
+  EnumValues,
   ResizeMode,
   SelectedTrack,
   SelectedTrackType,
@@ -55,8 +56,8 @@ type Props = {
   setRate: (value: number) => void;
   volume: number;
   setVolume: (value: number) => void;
-  resizeMode: ResizeMode;
-  setResizeMode: (value: ResizeMode) => void;
+  resizeMode: EnumValues<ResizeMode>;
+  setResizeMode: (value: EnumValues<ResizeMode>) => void;
   isLoading: boolean;
   srcListId: number;
   useCache: boolean;
@@ -216,8 +217,17 @@ const _Overlay = forwardRef<VideoRef, Props>((props, ref) => {
   };
 
   const onResizeModeSelected = (value: MultiValueControlPropType) => {
-    if (typeof value === 'object') {
-      setResizeMode(value);
+    if (typeof value === 'string') {
+      switch (value) {
+        case ResizeMode.CONTAIN:
+          setResizeMode(ResizeMode.CONTAIN);
+          break;
+        case ResizeMode.COVER:
+          setResizeMode(ResizeMode.COVER);
+          break;
+        default:
+          setResizeMode(ResizeMode.STRETCH);
+      }
     }
   };
 
--- a/examples/basic/src/components/Overlay.tsx
+++ b/examples/basic/src/components/Overlay.tsx
@@ -218,16 +218,7 @@ const _Overlay = forwardRef<VideoRef, Props>((props, ref) => {
 
   const onResizeModeSelected = (value: MultiValueControlPropType) => {
     if (typeof value === 'string') {
-      switch (value) {
-        case ResizeMode.CONTAIN:
-          setResizeMode(ResizeMode.CONTAIN);
-          break;
-        case ResizeMode.COVER:
-          setResizeMode(ResizeMode.COVER);
-          break;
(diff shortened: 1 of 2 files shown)


examples/basic/src/components/Overlay.tsx
```diff
@@ -216,8 +217,17 @@ const _Overlay = forwardRef<VideoRef, Props>((props, ref) => {
...
+        case ResizeMode.CONTAIN:
+          setResizeMode(ResizeMode.CONTAIN);
+          break;
+        case ResizeMode.COVER:
+          setResizeMode(ResizeMode.COVER);
+          break;
+        default:
+          setResizeMode(ResizeMode.STRETCH);
+      }
     }
```
moskalakamil: Is it possible to make setResizeMode(value)? I think switch can be overcoding

seyedmostafahasani: The setResizeMode function accepts `EnumValue<ResizeMode>`, so we cannot pass a string to this function. Instead, we can cast the string to `EnumValue<ResizeMode>`.
`setResizeMode(value as EnumValues<ResizeMode>);`

moskalakamil: maybe we can make some changes to be able to pass the `EnumValue<ResizeMode>` value type to this function?
```tsx
const onResizeModeSelected = (value: EnumValue<ResizeMode>) => {
```

seyedmostafahasani: If you look at the code, you'll see that we have a component called `MultiValueControl` that is used by rate, volume, and resizeMode. Therefore, we cannot change the type of this function.

moskalakamil: Sure! I get it

moskalakamil: I found a way to do this, when replacing `MultiValueControl.tsx` with this code:
```tsx
import React from 'react';
import {
  StyleSheet,
  Text,
  TextStyle,
  TouchableOpacity,
  View,
} from 'react-native';
import {ResizeMode} from 'react-native-video';

/*
 * MultiValueControl displays a list clickable text view
 */

interface MultiValueControlType<T> {
  // a list a string or number to be displayed
  values: Array<T>;
  // The selected value in values
  selected?: T;
  // callback to press onPress
  onPress: (arg: T) => void;
}

const MultiValueControl = <T extends number | string | ResizeMode>({
  values,
  selected,
  onPress,
}: MultiValueControlType<T>) => {
  const selectedStyle: TextStyle = StyleSheet.flatten([
    styles.option,
    {fontWeight: 'bold'},
  ]);

  const unselectedStyle: TextStyle = StyleSheet.flatten([
    styles.option,
    {fontWeight: 'normal'},
  ]);

  return (
      {values.map(value => {
        const _style = value === selected ? selectedStyle : unselectedStyle;
        return (
          <TouchableOpacity
            key={value}
            onPress={() => {
              onPress?.(value);
            }}>
            <Text style={_style}>{value}</Text>
          </TouchableOpacity>
        );
      })}
    </View>
  );
};

const styles = StyleSheet.create({
  option: {
    alignSelf: 'center',
    fontSize: 11,
    color: 'white',
    paddingLeft: 2,
    paddingRight: 2,
    lineHeight: 12,
  },
  container: {
    flex: 1,
    flexDirection: 'row',
    alignItems: 'center',
    justifyContent: 'center',
  },
});

export default MultiValueControl;
```
And in `Overlay.tsx` change the following:

```tsx
  const onResizeModeSelected = (value: EnumValues<ResizeMode>) => {
  // ...
  const onVolumeSelected = (value: number) => {
  // ...
  const onRateSelected = (value: number) => {
```
We are able to have less code in our app

seyedmostafahasani: Thank you for the suggestion, but I think it isn't related to this PR. We can address it in another PR.

KrzysztofMoch (approved):

examples/basic/src/components/Overlay.tsx
```diff
@@ -216,8 +217,17 @@ const _Overlay = forwardRef<VideoRef, Props>((props, ref) => {
   };
 
   const onResizeModeSelected = (value: MultiValueControlPropType) => {
-    if (typeof value === 'object') {
-      setResizeMode(value);
+    if (typeof value === 'string') {
+      switch (value) {
+        case ResizeMode.CONTAIN:
```
freeboub: Following previous comment, I think you can use 'as' to simplify this function, can you have a look please?

moskalakamil (approved):

freeboub (approved):

PR #3990, opened 2024-07-10
chore(example/basic): refactor code to follow best react-native practices

moskalakamil:
## Summary
Update example code

### Motivation
Update rnv basic example code to follow best rn practices
## Test plan


examples/basic/src/VideoPlayer.tsx
```diff
@@ -201,15 +210,10 @@ const VideoPlayer: FC<Props> = ({}) => {
...
-    console.log('channel up');
-    goToChannel((srcListId + 1) % srcList.length);
-  }, [srcListId]);
-
-  const channelDown = useCallback(() => {
-    console.log('channel down');
-    goToChannel((srcListId + srcList.length - 1) % srcList.length);
-  }, [srcListId]);
+  const onFullScreenExit = () => {
+    // iOS pauses video on exit from full screen
```
freeboub, line 225: Not sure why you add this specific case ?!
A log would be enough

examples/basic/src/VideoPlayer.tsx
```diff
@@ -201,15 +210,10 @@ const VideoPlayer: FC<Props> = ({}) => {
     });
   };
 
-  const channelUp = useCallback(() => {
```
freeboub, line 215: Do you really need to move these functions?

examples/basic/src/VideoPlayer.tsx
```diff
@@ -47,7 +47,6 @@ const VideoPlayer: FC<Props> = ({}) => {
   const [_, setVideoSize] = useState({videoWidth: 0, videoHeight: 0});
   const [paused, setPaused] = useState(false);
   const [fullscreen, setFullscreen] = useState(true);
-  const [decoration, setDecoration] = useState(true);
```
freeboub, line 61: I don't understand why this is removed ?

KrzysztofMoch (approved):

freeboub (changes requested):

examples/basic/src/VideoPlayer.tsx
moskalakamil: When we exited fullscreen mode on iOS, the video was paused, and the state didn't update - to resume the video, we had to click the button twice instead of once

examples/basic/src/VideoPlayer.tsx
moskalakamil: in react we should use functions at top of the component

examples/basic/src/VideoPlayer.tsx
moskalakamil: We didn't use this state because we enter fullscreen mode with a ref, so the state isn't used

examples/basic/src/VideoPlayer.tsx
freeboub: OK, but goToChannel is now declared after channelUp / channelDown which looks strange to me ...

examples/basic/src/VideoPlayer.tsx
freeboub: Ok understood, but here you are hiding a bug in the package ?

freeboub:
And you have conflict now ...

examples/basic/src/VideoPlayer.tsx
moskalakamil: It's not a bug in package -> just in example app where we have "paused" or "playing" custom control and it wasn't updated after exiting fullscreen on ios

examples/basic/src/VideoPlayer.tsx
moskalakamil: updated :))

moskalakamil:
@freeboub resolved conflicts 👍

freeboub (approved):

freeboub:
OK for me, I just fix the conflict, should be OK to merge

PR #3991, opened 2024-07-11
fix(ios): remove pip check for other platforms

KrzysztofMoch:
## Summary
Remove Picture In Picture check for other platforms then iOS

### Motivation
Fix builds on AppleTV and visionOS

### Changes
- Remove Picture In Picture check for other platforms then iOS

## Test plan
- build app on AppleTV

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -172,7 +172,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     }
 
     func isPipActive() -> Bool {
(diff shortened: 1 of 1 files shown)


ios/Video/RCTVideo.swift
```diff
@@ -171,6 +171,14 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
         return _pictureInPictureEnabled
     }
 
+    func isPipActive() -> Bool {
+        #if os(ios)
```
KrzysztofMoch:
```suggestion
        #if os(iOS)
```

PR #3993, opened 2024-07-11
refactor(android): migrate AudioOutput to Kotlin

seyedmostafahasani:
## Summary
Migrate AudioOutput to Kotlin

### Motivation

## Test plan


freeboub (approved):

android/src/main/java/com/brentvatne/exoplayer/AudioOutput.kt
```diff
@@ -0,0 +1,25 @@
+package com.brentvatne.exoplayer
+
+import android.annotation.SuppressLint
+import androidx.media3.common.C
+
+@SuppressLint("InlinedApi")
+enum class AudioOutput(private val mName: String, @C.StreamType val streamType: Int) {
```
KrzysztofMoch: I think we can rename `mName` to just `name`
```kt
enum class AudioOutput(private val mName: String, @C.StreamType val streamType: Int) {
```

seyedmostafahasani: We cannot choose it because `name` is already a member of the `Enum` class. If you agree, we can rename it to `outputName`.

KrzysztofMoch: Sure, we can do rename

seyedmostafahasani: Do you agree with renaming `d` to `entry`?

KrzysztofMoch (approved):

PR #3997, opened 2024-07-11
refactor(android): migrate AudioBecomingNoisyReceiver to Kotlin

seyedmostafahasani:
## Summary
Migrate AudioBecomingNoisyReceiver to Kotlin

### Motivation

## Test plan


android/src/main/java/com/brentvatne/receiver/AudioBecomingNoisyReceiver.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.content.IntentFilter
+import android.media.AudioManager
+import androidx.core.content.ContextCompat
+
+class AudioBecomingNoisyReceiver(private val context: Context) : BroadcastReceiver() {
+    private var listener: BecomingNoisyListener = BecomingNoisyListener.NO_OP
+
+    init {
+        context.applicationContext
+    }
```
KrzysztofMoch: I think we want to do it that way (but please double check it)
```suggestion
class AudioBecomingNoisyReceiver(private val context: Context) : BroadcastReceiver() {
    private var listener: BecomingNoisyListener = BecomingNoisyListener.NO_OP
    private var context = context.applicationContext
```

seyedmostafahasani: It was a good point.

freeboub (approved):

KrzysztofMoch (approved):

PR #4012, opened 2024-07-16
fix(android): index of the selected track

seyedmostafahasani:
## Test plan


KrzysztofMoch (approved):

freeboub:
@seyedmostafahasani can you please describe your change please? Is it solving issue with specific content?

seyedmostafahasani:
@freeboub in the sample app, we show a list of videos that a user can switch between. Before this change, the selected tracker was always the first item because trackIndex always passed `0`. With this change, the correct selected track is always passed. Previously, if you ran the sample app, you could not see the subtitle for track `BigBugBunny sideLoaded subtitles`(last track in the sample app). However, now you can see it because the selected track is correct.

moskalakamil (approved):

PR #4018, opened 2024-07-17
feat(android): add error handling for Kotlin version mismatch

### LINKED ISSUE #4009: [BUG]: Android Kotlin version mismatch
### Version

^6.4.1

Android

### System Version

React Native

Real device, Simulator

### Architecture

Old architecture

### What happened?

> Task :react-native-video:compileReleaseKotlin FAILED
e: Incompatible classes were found in dependencies. Remove them from the classpath or use '-Xskip-metadata-version-check' to suppress errors
e: /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/META-INF/kotlin-stdlib-jdk7.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/META-INF/kotlin-stdlib-jdk8.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/META-INF/kotlin-stdlib.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /Users/builder/.gradle/caches/transforms-3/46c2d7941a38979bf7d7f7ec3d090515/transformed/jetified-kotlin-stdlib-common-1.9.10.jar!/META-INF/kotlin-stdlib-common.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /Users/builder/.gradle/caches/transforms-3/87bfc0762e08f5de7559bcbc54306d9d/transformed/jetified-okio-jvm-3.6.0.jar!/META-INF/okio.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /Users/builder/clone/node_modules/react-native-video/android/src/main/java/com/brentvatne/common/api/BufferConfig.kt: (40, 14): Class 'kotlin.jvm.JvmStatic' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
The class is loaded from /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/kotlin/jvm/JvmStatic.class
e: /Users/builder/clone/node_modules/react-native-video/android/src/main/java/com/brentvatne/common/api/BufferConfig.kt: (43, 22): Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
The class is loaded from /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/kotlin/Unit.class
e: /Users/builder/clone/node_modules/react-native-video/android/src/main/java/com/brentvatne/common/api/BufferConfig.kt: (44, 22): Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
The class is loaded from /Users/builder/.gradle/caches/transforms-3/113c75e423850f09a3a96afbb04140d6/transformed/jetified-kotlin-stdlib-1.9.10.jar!/kotlin/Unit.class
...
...
...

* What went wrong:
Execution failed for task ':react-native-video:compileReleaseKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction

### Reproduction

Just trying to compile for release the application

seyedmostafahasani:
## Summary
Create a custom task to check the Kotlin metadata version.

### Motivation
Show clarify error message for Kotlin version mismatch.
And fix: https://github.com/TheWidlarzGroup/react-native-video/issues/4009

## Test plan
I changed the Kotlin version in the sample app to `1.7.0`, then ran the app and encountered a new error: `Kotlin version mismatch: Project is using Kotlin version '1.7.0' but it must be at least '1.8.0'. Please update the Kotlin version`. After reverting the Kotlin version to `1.8.0` or higher, the app worked correctly.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,45 @@
 import com.android.Version
+import groovy.json.JsonSlurper
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+
+def getReactNativeVersion() {
+    def packageJsonFile = file("$rootDir/../package.json")
+    if (packageJsonFile.exists()) {
+        def packageJson = new JsonSlurper().parseText(packageJsonFile.text)
+        return packageJson.dependencies["react-native"] ?: packageJson.devDependencies["react-native"]
+    }
+    return null
+}
+
+def reactNativeVersion = getReactNativeVersion()
+
+def isReactNativeVersionBetween = { version, minVersion, maxVersionExclusive ->
+    def versionParts = { v -> v.replaceAll(/[^0-9.]/, '').tokenize('.')*.toInteger() }
+    def (v, minV, maxV) = [version, minVersion, maxVersionExclusive].collect { versionParts(it) }
+
+    def isAtLeastMin = { v1, v2 ->
+        for (int i = 0; i < Math.max(v1.size(), v2.size()); i++) {
+            int val1 = i < v1.size() ? v1[i] : 0
+            int val2 = i < v2.size() ? v2[i] : 0
+            if (val1 < val2) return false
+            if (val1 > val2) return true
+        }
+        return true
+    }
+
+    def isBelowMax = { v1, v2 ->
+        for (int i = 0; i < Math.max(v1.size(), v2.size()); i++) {
+            int val1 = i < v1.size() ? v1[i] : 0
+            int val2 = i < v2.size() ? v2[i] : 0
+            if (val1 > val2) return false
+            if (val1 < val2) return true
+        }
+        return false
+    }
+    return isAtLeastMin(v, minV) && isBelowMax(v, maxV)
+}
 
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNVideo_kotlinVersion']
@@ -210,6 +248,8 @@ def media3_version = safeExtGet('media3Version')
 def kotlin_version = safeExtGet('kotlinVersion')
 def androidxCore_version = safeExtGet('androidxCoreVersion')
 def androidxActivity_version = safeExtGet('androidxActivityVersion')
+def minReactNativeVersion = safeExtGet("minReactNativeVersion")
+def maxReactNativeVersionExclusive = safeExtGet("maxReactNativeVersionExclusive")
 
 dependencies {
     // For < 0.71, this will be from the local maven repo
@@ -303,5 +343,8 @@ dependencies {
         // Common functionality used across multiple media libraries
         implementation "androidx.media3:media3-common:$media3_version"
     }
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+
+    if (isReactNativeVersionBetween(reactNativeVersion, minReactNativeVersion, maxReactNativeVersionExclusive)) {
+        implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,45 +1,7 @@
 import com.android.Version
-import groovy.json.JsonSlurper
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
-
-def getReactNativeVersion() {
-    def packageJsonFile = file("$rootDir/../package.json")
-    if (packageJsonFile.exists()) {
-        def packageJson = new JsonSlurper().parseText(packageJsonFile.text)
-        return packageJson.dependencies["react-native"] ?: packageJson.devDependencies["react-native"]
-    }
-    return null
-}
-
-def reactNativeVersion = getReactNativeVersion()
-
-def isReactNativeVersionBetween = { version, minVersion, maxVersionExclusive ->
-    def versionParts = { v -> v.replaceAll(/[^0-9.]/, '').tokenize('.')*.toInteger() }
-    def (v, minV, maxV) = [version, minVersion, maxVersionExclusive].collect { versionParts(it) }
-
-    def isAtLeastMin = { v1, v2 ->
-        for (int i = 0; i < Math.max(v1.size(), v2.size()); i++) {
-            int val1 = i < v1.size() ? v1[i] : 0
-            int val2 = i < v2.size() ? v2[i] : 0
-            if (val1 < val2) return false
-            if (val1 > val2) return true
-        }
-        return true
-    }
-
-    def isBelowMax = { v1, v2 ->
-        for (int i = 0; i < Math.max(v1.size(), v2.size()); i++) {
-            int val1 = i < v1.size() ? v1[i] : 0
-            int val2 = i < v2.size() ? v2[i] : 0
-            if (val1 > val2) return false
-            if (val1 < val2) return true
-        }
-        return false
-    }
-    return isAtLeastMin(v, minV) && isBelowMax(v, maxV)
-}
 
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNVideo_kotlinVersion']
@@ -245,11 +207,8 @@ repositories {
 }
 
 def media3_version = safeExtGet('media3Version')
-def kotlin_version = safeExtGet('kotlinVersion')
 def androidxCore_version = safeExtGet('androidxCoreVersion')
 def androidxActivity_version = safeExtGet('androidxActivityVersion')
-def minReactNativeVersion = safeExtGet("minReactNativeVersion")
-def maxReactNativeVersionExclusive = safeExtGet("maxReactNativeVersionExclusive")
 
 dependencies {
     // For < 0.71, this will be from the local maven repo
@@ -342,9 +301,5 @@ dependencies {
 
         // Common functionality used across multiple media libraries
         implementation "androidx.media3:media3-common:$media3_version"
-    }
-
-    if (isReactNativeVersionBetween(reactNativeVersion, minReactNativeVersion, maxReactNativeVersionExclusive)) {
-        implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -30,10 +30,10 @@ buildscript {
     }
 
     ext {
-        if (!isVersionAtLeast(kotlinVersion, requiredKotlinVersion)) {
-            throw new GradleException("Kotlin version mismatch: Project is using Kotlin version $kotlinVersion, but it must be at least $requiredKotlinVersion. Please update the Kotlin version.")
+        if (!isVersionAtLeast(kotlin_version, requiredKotlinVersion)) {
+            throw new GradleException("Kotlin version mismatch: Project is using Kotlin version $kotlin_version, but it must be at least $requiredKotlinVersion. Please update the Kotlin version.")
         } else {
-            println("Kotlin version is correct: $kotlinVersion")
+            println("Kotlin version is correct: $kotlin_version")
         }
     }
 }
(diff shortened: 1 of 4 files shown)


android/gradle.properties
```diff
@@ -13,3 +13,4 @@ RNVideo_useExoplayerHls=true
 RNVideo_androidxCoreVersion=1.9.0
 RNVideo_androidxActivityVersion=1.7.0
 RNVideo_buildFromMedia3Source=false
+RNVideo_requiredKotlinVersion=1.9.0
```
YangJonghun: I hope use 1.8.0. Default version of Kotlin within RN templat is 1.8.0.
(from react-native@0.73.*)

KrzysztofMoch: I agree - I have tested and 1.8.0 is working
cc @seyedmostafahasani

android/build.gradle
```diff
@@ -15,7 +15,39 @@ buildscript {
     }
 }
 
-// This looks funny but it's necessary to keep backwards compatibility (: 
+task checkKotlinVersion {
+    doLast {
+        def kotlinVersion = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNVideo_kotlinVersion']
+        def requiredKotlinVersion = project.properties['RNVideo_requiredKotlinVersion']
```
KrzysztofMoch: We don't need `RNVideo_requiredKotlinVersion` - for `def requiredKotlinVersion` we can get `project.properties['RNVideo_kotlinVersion']` as this version is "min kotlin version required for library"

android/gradle.properties
seyedmostafahasani: @YangJonghun @KrzysztofMoch it's applied.

android/build.gradle
```diff
@@ -15,7 +15,39 @@ buildscript {
...
+            throw new GradleException("Kotlin version mismatch: Project is using Kotlin version $kotlinVersion, but it must be at least $requiredKotlinVersion. Please update the Kotlin version.")
+        } else {
+            println("Kotlin version is correct: $kotlinVersion")
+        }
+    }
+}
+
+tasks.matching { it.name.startsWith('pre') || it.name == 'compileDebugKotlin' }.all {
+    dependsOn checkKotlinVersion
+}
```
KrzysztofMoch: I don't think that we need to do it that we - we can convert `checkKotlinVersion` task to function and call this function eg. in buildscript or somewhere else

KrzysztofMoch (changes requested):
Left some comments - we also need remember to update documentation (also we should add info that we require `compileSdkVersion` & `targetSdkVersion` to be set to `34`

freeboub:
Hello, I don't understand which version of kotlin is used in the sample. I tried to look for it and it seems to be 1.7.0, please correct me if I am wrong ?

android/build.gradle
seyedmostafahasani: I think we need it because we want to check it out with kotlin version of react native app. as you said eailier, min kotlin version must be `1.8.0` on the app.

KrzysztofMoch:
@freeboub `1.7.0` is version "set" by `react-native-video` - sample is using `1.9.22` (as it's forced by react-native)
Thats why we did not have those issue with kotlin bcs our sample is using hight kotlin version

android/build.gradle
KrzysztofMoch: Yes you are right but `RNVideo_kotlinVersion` in `react-native-video/android/gradle.properties` is already min required version - we don't need another variable for it. 
`rootProject` is "project" of app
`project` is "project" of library
so `project.properties['RNVideo_kotlinVersion']` will always be our min kotlin version

seyedmostafahasani: I agree with you so we must update `RNVideo_kotlinVersion` to `1.8.0`. 
wdyt?

KrzysztofMoch: Yes, that's what I meant - then we can remove `RNVideo_requiredKotlinVersion `

freeboub:
Just for memory:
release react-native 0.73.0 : Kotlin to 1.8.0 and JDK Toolchain to 11
release react-native 0.72.0 : Kotlin to 1.7.22 for Gradle

android/build.gradle
seyedmostafahasani: It's refactored.

freeboub:
I had the issue while checking another bug. To workaround it I just remove : 
`implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version" ` from build.gradle.
Won't it be better to use the kotlin version used by react native and add the dependency only on old react native version which don't use it ?

seyedmostafahasani:
@freeboub, if removing the `kotlin-stdlib` dependency fixes the issue, we can handle it conditionally based on the React Native version. For example, we can add the dependency only for React Native versions below a certain threshold:

`if (reactNativeVersion < '0.65.0') {
    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
}`

However, I don't think this dependency is the root cause of the problem. We should investigate further to identify the exact issue.

WDYT?

KrzysztofMoch:
> I had the issue while checking another bug. To workaround it I just remove : 
> 
> `implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version" ` from build.gradle.
> 
> Won't it be better to use the kotlin version used by react native and add the dependency only on old react native version which don't use it ?
> 
> 

We should remove it I guess - I have checked template (react-native-create-library) and there is no line like this. You can also look at other libraries

seyedmostafahasani:
@KrzysztofMoch you're right. I've checked other libraries like `view-shot` and `vision-camera`, and I didn't see it. So, if you agree, we can remove it.

freeboub:
We all agree to remove it then :) I wonder if we should put a note in doc for old react native version (0.69 ?) to add it manually in app build.gradle ?

KrzysztofMoch:
I think this is good idea - we can put it even in installation section

seyedmostafahasani:
We can handle it in `build.gradle`. If the React Native version is `0.69.0` or `below`, it will be added. For `higher` versions, it doesn't apply.
@freeboub  @KrzysztofMoch  WDYT?

freeboub:
Ok for me but please check in which react native version kotlin was introduced. If it is before 0.69, no need to add it as it is noticed that 0.69 is the lowest supported version

seyedmostafahasani:
@freeboub I checked and noticed that starting from react native version 0.73.0, Kotlin was introduced. So, if the react native version of the app is between 0.69 and 0.72, we can add it.
https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks#kotlin-template-on-android

freeboub:
> @freeboub I checked and noticed that starting from react native version 0.73.0, Kotlin was introduced. So, if the react native version of the app is between 0.69 and 0.72, we can add it.
> https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks#kotlin-template-on-android

Yes, let's go with that !

seyedmostafahasani:
@freeboub @KrzysztofMoch @YangJonghun it has been updated according to the final suggestion.

android/build.gradle
```diff
@@ -1,10 +1,63 @@
 import com.android.Version
+import groovy.json.JsonSlurper
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
+def getReactNativeVersion() {
+    def packageJsonFile = file("$rootDir/../package.json")
```
freeboub: Package.json is not necessarily in ../ is there any other way to retrieve the version ? I am also wondering if some frameworks can hide react-native in sub package 🤨

android/build.gradle
```diff
@@ -280,5 +343,8 @@ dependencies {
         // Common functionality used across multiple media libraries
         implementation "androidx.media3:media3-common:$media3_version"
     }
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-}
+
+    if (isReactNativeVersionBetween(reactNativeVersion, minReactNativeVersion, maxReactNativeVersionExclusive)) {
```
freeboub: Here you are blocking build for version < minVersion . I don't think this is really expected...

android/build.gradle
```diff
@@ -1,10 +1,63 @@
...
+        return false
+    }
+    return isAtLeastMin(v, minV) && isBelowMax(v, maxV)
+}
+
 buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['RNVideo_kotlinVersion']
+    def requiredKotlinVersion = project.properties['RNVideo_kotlinVersion']
+
+    def isVersionAtLeast = { version, requiredVersion ->
```
freeboub, line 10: Looks like duplicate function with isAtLeastMin

android/build.gradle
```diff
@@ -1,10 +1,63 @@
...
+        def packageJson = new JsonSlurper().parseText(packageJsonFile.text)
+        return packageJson.dependencies["react-native"] ?: packageJson.devDependencies["react-native"]
+    }
+    return null
+}
+
+def reactNativeVersion = getReactNativeVersion()
+
+def isReactNativeVersionBetween = { version, minVersion, maxVersionExclusive ->
+    def versionParts = { v -> v.replaceAll(/[^0-9.]/, '').tokenize('.')*.toInteger() }
```
freeboub: Will it work if user choose a RC version ? In a more general way, is there any gradle package to handle versioning check ?

freeboub (changes requested):

freeboub:
I just had a review , but in a more general way, I think we shouldn't integrate kotlin on the react native video side, but just let this dependency on the app side.
Your PR brings a lot of complexity which will be painful to maintain... 
Don't worry the PR looks fine, thank you again for handling this issue!!! I just doubting of the best way to hanle it !

seyedmostafahasani:
@freeboub I couldn't agree more with you. I was fixing this issue last night, and I thought about it. I think it's better to handle it on the app side. As you said earlier, maintaining these changes will be difficult and complex in the future. So, we can update the documentation installation.
Thanks for responding to my PR!

seyedmostafahasani:
Let's see @KrzysztofMoch's opinion about it, then we can apply it to this PR.

android/build.gradle
```diff
@@ -280,5 +343,8 @@ dependencies {
         // Common functionality used across multiple media libraries
         implementation "androidx.media3:media3-common:$media3_version"
     }
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-}
+
+    if (isReactNativeVersionBetween(reactNativeVersion, minReactNativeVersion, maxReactNativeVersionExclusive)) {
+        implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
KrzysztofMoch: I think we should remove it - but add to docs that we require kotlin 1.8.0+

KrzysztofMoch (approved):

KrzysztofMoch:
@freeboub I would like to release `6.4.3` after this PR merge - can you look into it ?

PR #4031, opened 2024-07-24
fix(android): viewType is ignored when set to ViewType.TEXTURE

### LINKED ISSUE #4030: [BUG]: viewType Texture is ignored
### Version

6.4.3

Android

### System Version

Android 14

Real device

### Architecture

Old architecture

### What happened?

I have to use Texture View because I have 2 videos that are overlapping

I wanted to set `viewType={ViewType.TEXTURE}` but it didn't work (whereas deprecated `useTextureView` worked properly.

It's due to the value of `ViewType.TEXTURE` which is `0` (so viewed as false)

I will make a PR to fix it

### Reproduction

repository link

### Reproduction

Create a 2 videos that are overlapping with `viewType={ViewType.TEXTURE}`, you will see an overlap

mlecoq:
…XTURE

## Summary
The goal is to fix the behavior of `viewType={ViewType.TEXTURE}` which does not work (the value is ignored)

### Motivation
To fix #4030 

### Changes
I have changed a condition in order to have `ViewType.TEXTURE` (0) taken into account

## Test plan
It's a minor changement, I have tested it locally by setting `viewType` as `ViewType.TEXTURE` and it worked properly

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -584,28 +584,32 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
       const shallForceViewType =
         hasValidDrmProp && (viewType === ViewType.TEXTURE || useTextureView);
 
-      if (shallForceViewType) {
-        console.warn(
-          'cannot use DRM on texture view. please set useTextureView={false}',
-        );
-      }
       if (useSecureView && useTextureView) {
         console.warn(
           'cannot use SecureView on texture view. please set useTextureView={false}',
         );
       }
 
-      return shallForceViewType
-        ? useSecureView
-          ? ViewType.SURFACE_SECURE
-          : ViewType.SURFACE // check if we should force the type to Surface due to DRM
-        : viewType
-        ? viewType // else use ViewType from source
-        : useSecureView // else infer view type from useSecureView and useTextureView
-        ? ViewType.SURFACE_SECURE
-        : useTextureView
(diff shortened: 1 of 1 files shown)


seyedmostafahasani:
@mlecoq thanks for fixing this issue. I think it's better to remove #4030 from your commit, as you mentioned it in the motivation of your summary. Also, I think we can refactor it for simplicity and readability, as shown below.
```
if (shallForceViewType) {
    if (useSecureView) {
        return ViewType.SURFACE_SECURE;
    }
    return ViewType.SURFACE;
}

if (viewType !== null && viewType !== undefined) {
    return viewType;
}

if (useSecureView) {
    return ViewType.SURFACE_SECURE;
}

if (useTextureView) {
    return ViewType.TEXTURE;
}

return ViewType.SURFACE;
```
WDYT? @freeboub @KrzysztofMoch

freeboub (approved):

KrzysztofMoch (approved):
Fine for me, but If it is not problem could you apply @seyedmostafahasani suggestion ?

mlecoq:
@KrzysztofMoch @seyedmostafahasani done !

moskalakamil (approved):

jspizziri:
FWIW, I've tested this locally and it seems to be working from my albeit limited perspective.

PR #4033, opened 2024-07-24
fix(ios): metadata update race

blazlew:
## Summary

This fixes a bug where metadata sometimes did not update correctly in the notification center when the video source was changed. I believe this was caused by a race condition between 2 methods updating it.
As I don't know much about this codebase nor it contain any tests, I'm happy to know your feedback on it.
I'm already testing it in my project and it seems to work just fine.

### Motivation

Fix for: https://github.com/TheWidlarzGroup/react-native-video/issues/4032

### Changes

- unified methods that update metadata in the notification center
- removed misleading comment

## Test plan

1. play a video with metadata (title, artwork)
1. move the app into background observing metadata in the notification center
1. swap video source to different one until issue happens (just a few swaps are enough)

### DIFF
--- a/ios/Video/NowPlayingInfoCenterManager.swift
+++ b/ios/Video/NowPlayingInfoCenterManager.swift
@@ -107,13 +107,11 @@ class NowPlayingInfoCenterManager {
         registerCommandTargets()
 
         updateMetadata()
-
-        // one second observer
         playbackObserver = player.addPeriodicTimeObserver(
             forInterval: CMTime(value: 1, timescale: 4),
             queue: .global(),
             using: { [weak self] _ in
-                self?.updatePlaybackInfo()
+                self?.updateMetadata()
             }
         )
     }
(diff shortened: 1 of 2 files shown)


KrzysztofMoch:
Huh that strange `updateMetadata` is function that we call "manually" to update data like title, artwork etc.
`updatePlaybackInfo` is function that is called every second after setting `currentPlayer` to update "dynamic" data like duration, current position and other dynamic values
Also `updateMetadata` will override `updatePlaybackInfo` when is called but not in other way

Just to be sure, the metadata that is not updating is artwork, title ?

blazyliszek:
@KrzysztofMoch yeah, I got the idea behind this.
From my understanding `updatePlaybackInfo` is called 4x times more often, that is every 0.25 sec.
My theory is while those 2 function ran in parallel the `updateMetadata` tended to finish first, so it updated correctly, but a few milliseconds later `updatePlaybackInfo` merged it using previous outdated content.
So instead of fighting this race, I decided to cancel it.
First thing I noticed while working on the fix is that when I accidentally slowed down the `updateMetadata` by putting some logs in there, the issue was gone.
> Just to be sure, the metadata that is not updating is artwork, title ?

Yes, that's what I noticed.

I was a bit of sceptical, If I won't make the artwork pull or metadata override too often this way, but this seems to be optimised at a lower level.

KrzysztofMoch:
It can be caused by those lines 
https://github.com/TheWidlarzGroup/react-native-video/blob/aed331e54f370102bc51eb9b7fa2676f5435fc3b/ios/Video/NowPlayingInfoCenterManager.swift#L207-L217

If there will be condition race then this will ignore existing data - what if we do merge here (like in `updatePlaybackInfo`)

If this will not work then I fine with your idea but please add those "metadata lines" that you removed in `updatePlaybackInfo` to `updateMetadata`

blazyliszek:
Thanks, yeah I didn't think about properties other than the ones we update here. I guess `nowPlayingInfo` may later carry more info that we don't want to lose.
In the mean time, I'll try out the merge approach and update this PR. I believe this should also work well.

blazyliszek:
@KrzysztofMoch updated so it's now merging instead overriding, in case nowPlayingInfo have more properties in the future

KrzysztofMoch (approved):
Thanks! Let's merge 🙌

PR #4042, opened 2024-07-31
fix(ios): updated getLicense call to work with new syntax, and fixed spelling error (#4014)

### LINKED ISSUE #4041: [BUG]: getLicense() override doesn't work after refactor
### Version

6.4.3

iOS

### System Version

Any

Real device

### Architecture

Old architecture

### What happened?

We use the `getLicense()` override on the `drm` parameter to manipulate the licence request/response.

Recent refactor work has moved the top level `drm` parameter to inside the `source` parameter and deprecated the `drm` parameter.

When updating our code, we found that content wouldn't play.

On further investigation, we've found that the override isn't found if moved to the new location, and also the refactor of `onGetLicense()` in Video.tsx has removed the call to `setLicenseResultCmd` meaning a successful licence response is never processed - it's just sent to `setLicenseResultErrorCmd`.  There was also a spelling mistake in the refactor, causing the call to `setLicenseResultCmd` to always fail.

### Reproduction

repository link

### Reproduction

Attempt to play DRM protected content on iOS devices using the non-deprecated syntax.

EETVApps:
## Summary

Corrected the typo (lisence -> license), and updated the getLicense call to be allowed in either the top level drm parameter, or drm inside source parameter

### Motivation

Fix #4041  

## Test plan

Successful playback of DRM protected content on iOS device

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -531,7 +531,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
             data.contentId,
             data.licenseUrl,
             data.loadedLicenseUrl,
-            );
+            )
           ).catch(() => {
             throw new Error('fetch error');
           });
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -526,12 +526,13 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
             throw new Error('No spc received');
           }
           // Handles both scenarios, getLicenseOverride being a promise and not.
-          const license = await Promise.resolve(selectedDrm.getLicense(
-            data.spcBase64,
-            data.contentId,
-            data.licenseUrl,
-            data.loadedLicenseUrl,
-            )
+          const license = await Promise.resolve(
+            selectedDrm.getLicense(
+              data.spcBase64,
+              data.contentId,
+              data.licenseUrl,
+              data.loadedLicenseUrl,
+            ),
           ).catch(() => {
             throw new Error('fetch error');
           });
@@ -545,7 +546,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
               data.loadedLicenseUrl,
             );
           }
-        } catch (e) { 
+        } catch (e) {
           const msg = e instanceof Error ? e.message : 'fetch error';
           if (nativeRef.current) {
             NativeVideoManager.setLicenseResultErrorCmd(
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -522,15 +522,22 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
         }
         const data = event.nativeEvent;
         try {
-          if (!data?.spcBase64) throw new Error('No spc received');
+          if (!data?.spcBase64) {
+            throw new Error('No spc received');
+          }
           // Handles both scenarios, getLicenseOverride being a promise and not.
           const license = await Promise.resolve(selectedDrm.getLicense(
             data.spcBase64,
             data.contentId,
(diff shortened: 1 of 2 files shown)


moskalakamil:
Hi, thanks for your PR! I think your pr title should be probably fix(ios) - not android, right?

src/Video.tsx
```diff
@@ -524,14 +525,24 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+                NativeVideoManager.setLicenseResultCmd(
+                  getReactTag(nativeRef),
+                  license,
+                  data.loadedLicenseUrl,
+                );
+              }
+              return;
+            } else {
+              result = 'Empty license result';
+            }
```
YangJonghun:
```ts
            if (typeof license === 'string') {
              if (nativeRef.current) {
                NativeVideoManager.setLicenseResultCmd(
                  getReactTag(nativeRef),
                  license,
                  data.loadedLicenseUrl,
                );
              }
              return;
            } else {
              result = 'Empty license result';
            }
```
I'm not sure above change is really necessary.

EETVApps: Without it, setLicenseResultCmd() was never called - the licence is never imported, and therefore playback never starts

YangJonghun: https://github.com/TheWidlarzGroup/react-native-video/blob/master/src/Video.tsx#L541-L547
I think previous code seemed to work fine. Is it not called from above?

EETVApps: That's calling setLicenseResult**Error**Cmd() - i.e. indicating that the licence call failed.

YangJonghun: I get it! thank you👍

YangJonghun: nit;
I've rewrite some code to make it a bit improve readability. It's just suggestion. you don't have to apply it :) Thank you for contributing.
```ts
        try {
          if (!data?.spcBase64) throw new Error('No spc received');
          // Handles both scenarios, getLicenseOverride being a promise and not.
          const license = await Promise.resolve(selectedDrm.getLicense(
            data.spcBase64,
            data.contentId,
            data.licenseUrl,
            data.loadedLicenseUrl,
          )).catch(() => { throw new Error('fetch error') });
          if (typeof license !== 'string') throw Error('Empty license result')
          if (nativeRef.current) {
            NativeVideoManager.setLicenseResultCmd(
              getReactTag(nativeRef),
              license,
              data.loadedLicenseUrl,
            );
          }
        } catch (e) { 
          const msg = e instanceof Error ? e.message : 'fetch error'
          if (nativeRef.current) {
            NativeVideoManager.setLicenseResultErrorCmd(
              getReactTag(nativeRef),
              msg,
              data.loadedLicenseUrl,
            );
          }
        }
```

YangJonghun (approved):

freeboub (approved):

src/Video.tsx
```diff
@@ -512,41 +512,51 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
...
+          }
+          if (nativeRef.current) {
+            NativeVideoManager.setLicenseResultCmd(
+              getReactTag(nativeRef),
+              license,
+              data.loadedLicenseUrl,
+            );
+          }
+        } catch (e) { 
+          const msg = e instanceof Error ? e.message : 'fetch error';
```
freeboub, line 550: It would be better to avoid hard coded string here, but I approve anyway.
I don't think it exist better way to proceed
There is a lack of error code spec/management in the package...

KrzysztofMoch (approved):

PR #4049, opened 2024-08-02
chore(android): remove onBackPressed function in FullScreenPlayerView

### LINKED ISSUE #4047: [BUG]: Android navigation bar stays hidden after video using fullscreen no longer mounted
### Version

6.4.3

Android

### System Version

Android 13

Simulator

### Architecture

Old architecture

### What happened?

After playing the video or closing the screen with the video using `fullscreen={true}`, the status bar reappears but the navigation bar doesn't.

Maybe it is related to [this issue](https://github.com/TheWidlarzGroup/react-native-video/pull/3978).

### Reproduction

_No response_

### Reproduction

- Show a screen with Video component with `fullscreen={true}`
- Close the screen

seyedmostafahasani:
## Summary
Removed the onBackPressed function.

### Motivation
I fixed a potential issue where the `setFullscreen` method was not being called if the user did not press the back button.
fix #4047, #4057

## Test plan


KrzysztofMoch:
@seyedmostafahasani correct me if I am wrong but this should fix #4047 ?

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -60,10 +60,6 @@ class FullScreenPlayerView(
     init {
         setContentView(containerView, generateDefaultLayoutParams())
     }
-    override fun onBackPressed() {
-        super.onBackPressed()
-        onBackPressedCallback.handleOnBackPressed()
-    }
```
KrzysztofMoch, line 88: Why removing this? (just for my knowledge)

seyedmostafahasani:
> @seyedmostafahasani correct me if I am wrong but this should fix #4047 ?

I think this PR can resolve it.

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
seyedmostafahasani: If the user presses the back button, the `onStop` method will eventually be called as the activity transitions to a stopped state. Therefore, we can call the logic for backButtonPressed inside the `onStop` function.

sntg-p:
What happens if the user does not press the system back button but they press a button that invokes `navigation.goBack()`? In this case `navigation` comes from React Navigation but it could be whatever routing solution the app is using.

seyedmostafahasani:
> What happens if the user does not press the system back button but they press a button that invokes `navigation.goBack()`? In this case `navigation` comes from React Navigation but it could be whatever routing solution the app is using.

@sntg-p I just want to be sure that I get you correctly. For full-screen mode you navigate to another screen and use the prop `fullscreen={true}`. To exit this mode you call the `goBack()` function.

sntg-p:
What happens if the user does not press the system back button but they press a button that invokes `navigation.goBack()`? In this case `navigation` comes from React Navigation but it could be whatever routing solution the app is using.

sntg-p:
> @sntg-p I just want to be sure that I get you correctly. For full-screen mode you navigate to another screen and use the prop `fullscreen={true}`. To exit this mode you call the `goBack()` function.

This is currently not working properly as I mentioned in https://github.com/TheWidlarzGroup/react-native-video/issues/4047. Not really sure if this PR fixes that, as I didn't test it.

seyedmostafahasani:
@sntg-p would you please share your code with me if it's possible or sample of your code?

mkbhdana:
> What happens if the user does not press the system back button but they press a button that invokes `navigation.goBack()`? In this case `navigation` comes from React Navigation but it could be whatever routing solution the app is using.

That's exactly i am using and then statusbar and navbar gets hidden

seyedmostafahasani:
@sntg-p @mkbhdana 
Could you please use the `setFullScreen` method as shown below?
```
const exitFullScreen = () => {
    navigation.goBack();
    setFullScreen(false);
}
```

mkbhdana:
> @sntg-p @mkbhdana Could you please use the `setFullScreen` method as shown below?
> 
> ```
> const exitFullScreen = () => {
>     navigation.goBack();
>     setFullScreen(false);
> }
> ```

Already added in my task list to try this in my available time.

freeboub:
I think there is a side effect here. If you go to full screen, leave the app (press home), and go back to the app, full screen will be dismissed as onStop has been called.
Can you please double check that ?

mkbhdana:
> I think there is a side effect here. If you go to full screen, leave the app (press home), and go back to the app, full screen will be dismissed as onStop has been called. Can you please double check that ?

i tried it on my app and fullscreen stays

seyedmostafahasani:
> I think there is a side effect here. If you go to full screen, leave the app (press home), and go back to the app, full screen will be dismissed as onStop has been called. Can you please double check that ?

I tested your scenario, and the app works correctly. I pressed the home button, then I went back to the app, but the app was in full-screen mode. `onStop` doesn't get called when I press the home button or background the app.

freeboub (approved):

seyedmostafahasani:
> > I think there is a side effect here. If you go to full screen, leave the app (press home), and go back to the app, full screen will be dismissed as onStop has been called. Can you please double check that ?
> 
> i tried it on my app and fullscreen stays

Would you please share a sample of your code with me?

seyedmostafahasani:
@mkbhdana 
Would you please share a sample of your code with me?
I think your issue is not related to this PR.

seyedmostafahasani:
@sntg-p @mkbhdana 
I think this PR (#4112) can resolve your issue.

mkbhdana:
> @mkbhdana Would you please share a sample of your code with me? I think your issue is not related to this PR.

well setFullscreen also not resolved my issue 

i just used react-native-system-navigation-bar this library and calling SystemNavigationBar.fullScreen(false); on onBack handler and this solves my entire problem with status and navigation bar while keeping fullscreen={true} in react native video component

mkbhdana:
@sntg-p also I find if I keep fullscreen = true then then navigation bar adds some space above it after dismissing video so to solve it I set fullscreen = false in video commponent and instead I use SystemNavigationBar.fullScreen(true); and SystemNavigationBar.immersive(); inside useEffect so it helped to resolve the issue.

seyedmostafahasani:
@KrzysztofMoch 
If you have enough time please check out this PR.

KrzysztofMoch (approved):

freeboub:
@seyedmostafahasani I fix conflicts manually, can you double it is still working on your side and we can merge ?

seyedmostafahasani:
@freeboub
First of all, thank you for resolving the conflicts. I tested it, and everything works correctly. It's ready to merge.

Issue #4006, opened 2024-07-14
[BUG]: hideShutterView does not work


FrederickEngelhardt:
### Version

6.4.1

### What platforms are you having the problem on?

Android

### System Version

Android 14

### On what device are you experiencing the issue?

Real device

### Architecture

Old architecture

### What happened?

`hideShutterView` prop for android use to work. I upgraded recently from `6.0.0-alpha.6` to `6.4.1` and ran into the problem where the shutter shows up as a black window (on a white background with a white background video that loads).

```tsx
      <RNVideo
        {...props}
        hideShutterView // should hide blackscreen on load for android (does not work on 6.4.1)
      />
```

## Workaround

Add a colored view overlay until the video player is ready. Interesting part is `onLoaded` and `onReadyForDisplay` still require a debounce of anywhere between 20-200ms for the black screen to not show.

```tsx
import { View, StyleSheet } from 'react-native'
import RNVideo from 'react-native-video'

// or a variant of this, could easily be a simple JS timer.
import debounce from 'lodash.debounce'

const component = () => {
    const [isLoaded, setIsLoaded] = useState(false)

  const debouncedOnLoad = useMemo(
    () => debounce(() => setIsLoaded(true), 200),
    [],
  )

  return (
    <View>
      <RNVideo
        hideShutterView // hides blackscreen
      />
      {!isLoaded && Platform.OS === 'android' && (
        <View
          style={[
            StyleSheet.absoluteFillObject,
            { backgroundColor: 'white', zIndex: 9999 },
          ]}
        />
      )}
    </View>
  )
}
```

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are: 
- Use 6.4.1 on android device. My device was a Note 20 Ultra. But should apply to all android devices using exoplayer

freeboub:
can you try `shuttercolor={'transparent'}` prop ? : https://thewidlarzgroup.github.io/react-native-video/component/props#shuttercolor

freeboub:
I tested hideShutterView in the sample app (6.4.3). It works as expected. I close the issue, but if I need to reopen, please provide a sample app.
and regarding to your expectation, I think you can put `shuttercolor={'white'}` !

Issue #2974, opened 2022-12-27
Black screen while initiating player on Android Devices


ajayr82:
# Bug

While initializing the player for playback a black screen appears for a moment - the duration depends on the device type. The below code seems to be causing this issue. Commenting the same seems to fix the issue. 

https://github.com/react-native-video/react-native-video/blob/master/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java:77

Can the background be set to transparent? Or perhaps prop can be provided to set the bg color

Which player are you experiencing the problem on:
* Android

hassenlazli:
Same issues

freeboub:
For sure, it make sens to be able to make background color configurable.
You will need to create a property dedicated to this feature.

moGeekHex:
+2

rmyr:
Having the same issue using the latest alpha. Kinda need to upgrade to the alpha as well, cause seeking is not working well in 5.2 (Android).

This is a major issue for our app, cause you can switch quickly between videos, so the black flicker becomes very noticable.

pgpankajbigstep:
Any solution for this issue?

brunoro:
Also seeing here when using the exoplayer on 5.2

brunoro:
@pgpankajbigstep the solution is to delete that line that sets the shutter background color on ExoPlayerView

wolfxpertlab:
same issue please tell me any solution for this issue

please help me

barghi:
shutterColor="transparent" prop
fixed the issue for me

freeboub:
Thank you @barghi . This ticket can be closed

Issue #2483, opened 2021-09-11
hideShutterView prop doesn't work


KyoungHwanLee:
# Bug

Even when set `hideShutterView` to true, the shutter view is not hidden.

https://github.com/react-native-video/react-native-video/blob/bf11502fdb83751ad826f3e4b448732075292805/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java#L131-L149
https://github.com/react-native-video/react-native-video/blob/bf11502fdb83751ad826f3e4b448732075292805/android-exoplayer/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java#L196-L210
142th line and 209th line ignore `hideShutterView` prop and force the visibility of shutter view to VISIBLE.
These codes should be replace to `this.updateShutterViewVisibility();`

## Platform
* Android ExoPlayer

## Environment info

Library version: 5.1.1

## Steps To Reproduce

Just set `hideShutterView` to true.

## Expected behaviour

1. Shutter view should be hidden

ldabiralai:
Looks like this was fixed as part of https://github.com/react-native-video/react-native-video/commit/6d74eb686f9c97924cd33dee288d4fd7a258944d any idea when this will be released? 🙏

tironiigor:
@ldabiralai I'm not sure, but you can always use the latest changes in your project!

PR #4060, opened 2024-08-06
fix(android): hide surfaceView for loading time when shutter is hidden

### LINKED ISSUE #4045: [BUG]: Black screen on Android until video fully loaded
### Version

6.4.3

Android

### System Version

android 12+

Real device, Simulator

### Architecture

Old architecture

### What happened?

I have a static photo and wanted to add a "live photo". This is a regular mp4 video that is uploaded via a link and activated.

But the problem is that in the latest versions of the library at the moment of "onLoadStart" a black canvas is called, which is shown until the video is loaded. I saw shutterColor and hideShutterView in the documentation, but using them doesn't solve the problem.

### Reproduction

_No response_

### Reproduction

Just try to loaded video from remote uri: for example https://d23dyxeqlo5psv.cloudfront.net/big_buck_bunny.mp4

moskalakamil:
## Summary
Fix black screen when `hideShutterView` is true
Closes #4045
### Motivation
Fix #4045, #4006, #2974, #2483
### Changes
- Hide surface view for loading time, when shutter view is hidden
- Allow to enter 'transparent' in `shutterColor` prop
## Test plan
run example app on android with Video `hideShutterView={true}` prop - there is no black screen then


android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -161,8 +161,22 @@ public void updateSurfaceView(@ViewType.ViewType int viewType) {
         }
     }
 
+    private void hideShutterView() {
+        shutterView.setVisibility(INVISIBLE);
+        ((SurfaceView)surfaceView).setAlpha(1);
```
freeboub: What about when the user configures the texture view ?

freeboub (changes requested):

bradfx (approved):

KrzysztofMoch (approved):

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
KrzysztofMoch: this was cast only - it's removed so it should be fine now

freeboub (approved):

freeboub:
I approve because code is OK, but I am not sure this is the expected behavior.
In attached ticket, peoples are asking to get previous video image when they load a new video ... here it will hide the previous video (opacity 0) !

CMLCNL:
When you can release this version? thank you.

PR #4061, opened 2024-08-06
infra: add issue validator and stale action

KrzysztofMoch:
## Summary
This PR introduce (simple) issue validator and stale action

#### Issue Validator
Issue validator will allow us to focus on "good reported" issues. It will inform users about missing fields or in case that all fields are fine will "thanks" users
You can see example issue here https://github.com/KrzysztofMoch/react-native-video/issues/6

#### Stale Action
Stale action will close issues that are inactive (I made it to don't close issue with label https://github.com/TheWidlarzGroup/react-native-video/labels/Accepted won't be staled)

### Changes
- Add action for issue validator
- Add action for issue stale

## Test plan
I have tested it on my fork


KrzysztofMoch:
@seyedmostafahasani cloud you also also look at it? I can't add you to reviewer

.github/scripts/validate.js
```diff
@@ -0,0 +1,209 @@
...
+      .map((field) => `- ${field.replace('missing-', '')}`)
+      .join(
+        '\n',
+      )}\n\nPlease update your issue with this information to help us address it more effectively. 
+      \n > Note: issues without complete information have a lower priority`;
+  },
+};
+
+const getFieldValue = (body, field) => {
+  if (!FIELD_MAPPINGS[field]) {
```
seyedmostafahasani, line 66: We can set it in a variable as shown below:
```suggestion
  const fieldMapping = FIELD_MAPPINGS[field];
  if(!fieldMapping) {
```
Then, we can use this variable in the function instead of `FIELD_MAPPINGS[field]`.

.github/scripts/validate.js
```diff
@@ -0,0 +1,209 @@
...
+    {
+      name: 'Version',
+      invalidValue: 'Put the exact version from your package.json',
+    },
+    {
+      name: 'SystemVersion',
+      invalidValue:
+        'What version of the system is using device that you are experiencing the issue?',
+    },
+    {name: 'DeviceType'},
```
seyedmostafahasani, line 129: What is your opinion on adding invalidValue for `DeviceType` and `Architecture`?

seyedmostafahasani:
@KrzysztofMoch 
LGTM! I left 2 small comments.

freeboub (approved):

freeboub:
I approve, but be careful, a ticket with no activity doesn't mean that this is not an issue. Just that nobody handled it !

.github/scripts/validate.js
KrzysztofMoch: Those values don't have "default" value so we don't need to check if values are "provided by default template"

seyedmostafahasani (approved):

moskalakamil (approved):

PR #4067, opened 2024-08-07
feat(android): handle navigation bar status in full-screen mode

### LINKED ISSUE #4055: Feature: Hiding Nav Bar
### Version

6.4.3

Android

### System Version

Android

Real device, Simulator

### Architecture

Old architecture

### What happened?

When the full-screen view is activated, the default controls are shown, and the mobile bottom navigation bar remains visible, even in full-screen mode. To address this issue, consider upgrading the FullScreenPlayerView found under com => brentvatne => exoplayer.

Code Implementation
To enable immersive mode and hide the navigation bar, override the onAttachedToWindow method:

override fun onAttachedToWindow() {
    super.onAttachedToWindow()
    // Enable immersive mode to hide the navigation bar
    window.decorView.systemUiVisibility = (
        View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY
        or View.SYSTEM_UI_FLAG_FULLSCREEN
        or View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
        or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
        or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
        or View.SYSTEM_UI_FLAG_LAYOUT_STABLE
    )

    if (reactExoplayerView.preventsDisplaySleepDuringVideoPlayback) {
        mKeepScreenOnHandler.post(mKeepScreenOnUpdater)
    }
}

Resetting UI Visibility
To reset the UI visibility to default, override the onDetachedFromWindow method:

override fun onDetachedFromWindow() {
    super.onDetachedFromWindow()
    // Reset UI visibility to default
    window.decorView.systemUiVisibility = (
        View.SYSTEM_UI_FLAG_LAYOUT_STABLE
        or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
        or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
    )
}

### Reproduction

repository link

### Reproduction

Step to reproduce this bug are: When the full-screen view is activated

seyedmostafahasani:
## Summary
Handle the status of the navigation bar in full-screen mode.

### Motivation
Fix #4055

### Changes
Add a new property to `controlsStyle` called `hideNavigationBarOnFullScreenMode`

## Test plan
If you set this property to `true`, the navigation bar will be hidden; if you set it to `false`, it will be visible.


android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -127,4 +134,29 @@ class FullScreenPlayerView(
...
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                window?.setDecorFitsSystemWindows(true)
+                window?.insetsController?.let {
+                    it.show(WindowInsets.Type.navigationBars())
+                }
+            } else {
+                @Suppress("DEPRECATION")
+                window?.decorView?.systemUiVisibility = View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
+            }
+        }
```
moskalakamil: You can move `if (controlsConfig.hideNavigationBarOnFullScreenMode) {` inside `window?.insetsController?.let`, then we avoid repeating ->
```tsx
if(Build.VERSION.SDK_INT>=Build.VERSION_CODES.R){
    window?.setDecorFitsSystemWindows(false)
    window?.insetsController?.let{
        if(controlsConfig.hideNavigationBarOnFullScreenMode){
            it.hide(WindowInsets.Type.navigationBars())
            it.systemBarsBehavior=WindowInsetsController.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
        }else{
            it.show(WindowInsets.Type.navigationBars())
        }
    }
}else{
    @Suppress("DEPRECATION")
    window?.decorView?.systemUiVisibility=View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
}
```

seyedmostafahasani: Thanks @moskalakamil for your suggestion, but I think you missed something. I have updated your suggestion.

moskalakamil (approved):

YangJonghun:
@seyedmostafahasani
Thanks for assign me as reviewer:)
I'm not sure this approach is good idea
Android has many different cases, such as status bar, navigation bar, immersive mode, etc. ([fyi](https://github.com/kadiraydinli/react-native-system-navigation-bar))
Unless you plan to support all of them, stickt immersive mode seems to be acceptable.
On top of that, it's affect whole application state, so I'm not sure if it will affect lifecycle of other libraries.

seyedmostafahasani:
@YangJonghun 
Thanks for reviewing the PR.
I think the code works correctly. however, it might not be robust enough to handle all possible Android UI configurations.
If you agree, I will apply your suggestion of using `sticky immersive mode`.

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -127,4 +134,25 @@ class FullScreenPlayerView(
         layoutParams.setMargins(0, 0, 0, 0)
         return layoutParams
     }
+
+    private fun navigationBarVisibility() {
```
freeboub: can you rename this function to : updateNavigationBarVisibility please ?

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
seyedmostafahasani: For sure.

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -127,4 +134,25 @@ class FullScreenPlayerView(
         layoutParams.setMargins(0, 0, 0, 0)
         return layoutParams
     }
+
+    private fun updateNavigationBarVisibility() {
```
freeboub, line 139: reviewing again the code, I wonder why did you choose the manage vibility here, where there is a similar code in: 
`public void setFullscreen(boolean fullscreen) {`
I am afraid we will be lost if we split 'fullscreen' responsibility...

seyedmostafahasani: In the `setFullScreen` function, we handle showing the navigation bar and status bar on the first screen, but now we want to handle it in Dialog (full-screen) mode. I think this is the correct way to handle it in this class `(FullScreenPlayerView)`.

freeboub (approved):

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -127,4 +134,25 @@ class FullScreenPlayerView(
...
+            }
+        } else {
+            @Suppress("DEPRECATION")
+            window?.decorView?.systemUiVisibility = if (controlsConfig.hideNavigationBarOnFullScreenMode) {
+                View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
+            } else {
+                View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
+            }
+        }
+    }
```
YangJonghun, line 157: I suggest using `WindowInsetsControllerCompat(window, window.decorView).hide(WindowInsetsCompat.Type.systemBars())`

docs/pages/component/props.mdx
```diff
@@ -157,6 +158,7 @@ controlsStyles={{
   hideSeekBar: false,
   hideDuration: false,
   seekIncrementMS: 10000,
+  hideNavigationBarOnFullScreenMode: false,
```
YangJonghun: Do we really need this prop? I think people expect `fullscreen` already include this.
I think giving freedom to features that are not closely related to video would increase the number of management points and leave a lot of room for side effects.🤔

YangJonghun: I think this feature may cause of conflict with screen management libraries like react-native-screens.

seyedmostafahasani: @freeboub @KrzysztofMoch 
WDYT?

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
seyedmostafahasani: Sorry for the delay in responding. I was busy the last few days. I applied your suggestion. Please check it out.

docs/pages/component/props.mdx
freeboub: @seyedmostafahasani I agree with this prop. more the package is flexible better it is.
BTW, I think it should be true by default :) but I am not sure, we should keep current behavior by default to avoid regression (behavior change) in apps

freeboub:
I found an issue with this PR.
can be reproduced with the sample:
I boot the app I have this UI:

Capture d’écran 2024-09-03 à 09 03 00

I click on hide decoration and leave fullscreen. now the UI looks like this:
Capture d’écran 2024-09-03 à 09 04 28

freeboub (changes requested):
See my previous comment

seyedmostafahasani:
@freeboub 
I applied your suggestion. The problem occurred before merging this PR (#4112) but should be fixed now.

freeboub:
Thank you for applying my proposal. BTW, I don't think it will fix the issue, but just hide it ...
We need to save initial state to be able to restore after leaving the full screen mode.
I will check what I can do, I have already made some changes locally

seyedmostafahasani:
I don't think we need to save the initial state because when we go to full-screen mode, we don't change anything that we would want to restore. We handle the navigation bar and status bar in full-screen mode within the dialog. Normal mode and full-screen mode are separate.

PR #4074, opened 2024-08-09
fix(ios): crash on source change after coming back from background

blazlew:
## Summary
This works around the crash on source change, if the app was previously put in the background. The crash occurs on iOS >= 16 and I blame apple for this 😅. They seem to be adding some observer logic on the player instance, causing it to explode in mentioned scenario.
This might not be a super sophisticated solution, perhaps there is a better more granular way to handle this, but on another hand it's simple and works well.
### Motivation
Fix for https://github.com/TheWidlarzGroup/react-native-video/issues/3900
### Changes
- temporary disabled [feature that allows for copying texts/subjects out of a video frame](https://developer.apple.com/documentation/avkit/avplayerview/allowsvideoframeanalysis) in the conflicting code part
## Test plan
Author of the issue [provided a crash demo app](https://github.com/TheWidlarzGroup/react-native-video/issues/3900#issuecomment-2169703076). 
I was able to run it with:
```sh
yarn
(cd ios && pod update)
yarn ios --udid=<your_sim_udid>
```
Once succesfuly ran:
1. play video
2. go background
3. go foreground
4. change video
5. notice the crash
6. replace rn-video dependency with suggested change
7. test again

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -488,7 +488,17 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
                 NowPlayingInfoCenterManager.shared.registerPlayer(player: _player!)
             }
         } else {
+            if #available(iOS 16.0, *) {
+                // This feature caused crashes, if the app was put in bg, before the source change
+                // https://github.com/TheWidlarzGroup/react-native-video/issues/3900
+                self._playerViewController?.allowsVideoFrameAnalysis = false
+            }
+
             _player?.replaceCurrentItem(with: playerItem)
+
+            if #available(iOS 16.0, *) {
+                self._playerViewController?.allowsVideoFrameAnalysis = true
+            }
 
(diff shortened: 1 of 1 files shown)


KrzysztofMoch:
👀 Good catch! I think we can upgrade it with `playInBackground` check  - WDYT ?

jaybayley:
Confirming this is still an issue in 6.4.4. Applying @blazlew fix via a patch file resolves the issue for me. @KrzysztofMoch Could we look at getting this merged please?

blazyliszek:
Thanks @KrzysztofMoch although the steps to reproduce mention backgrounding the app, the issue itself doesn't seem to be related to `playInBackground` flag. It's set to `false` in the demo 
https://github.com/ammar20305/Dummy/blob/8273127732173223a221b9ee4feeccbf737b9dd7/App.tsx#L80

freeboub (approved):

KrzysztofMoch (approved):
Okay, it LGTM so let's merge

PR #4090, opened 2024-08-15
feat(android): support hiding Exoplayer video duration on android

ashlyWeiting:
## Summary
This PR adds support for hiding the duration on Android devices.

### Changes
  -  Implemented a new prop to hide the duration on Android.
  -  Updated relevant Android-specific components to respect the new prop.

## Test plan
    1. Apply the new prop hideDuration = true to hide the duration.
    2. Verify that the video player does not display the duration when the prop is true.


seyedmostafahasani:
@ashlyWeiting 
I don't understand why we need this PR. In particular, when the user hides the seek bar, it indicates that the video is live, so we hide both the seek bar and the duration. However, if the user only wants to hide the duration, it might cause confusion.
Why would the user need to hide the duration?

ashlyWeiting:
I know this request is unusual, but over 10 Android users of my product have asked for it. Users think seeing the current position is sufficient. 
If you think this PR is unnecessary, I will close it.

Thank you for taking time to review my pull request!

seyedmostafahasani:
@freeboub @KrzysztofMoch 
WDYT?

freeboub:
I think this is not up to rnv to choose UI requirements. Btw, the pr is clean and easy to maintain, so ok for me !

freeboub (changes requested):
I just think there is a missing typescript change in the component file. Can you double check please?

seyedmostafahasani:
@ashlyWeiting 
Please add the missing TypeScript to this file `src/specs/VideoNativeComponent.ts`

```
type ControlsStyles = Readonly<{
  hideSeekBar?: boolean;
  hideDuration?: boolean;
  seekIncrementMS?: Int32;
}>;
```

src/types/video.ts
```diff
@@ -223,6 +223,7 @@ export type AudioOutput = 'speaker' | 'earpiece';
 
 export type ControlsStyles = {
   hideSeekBar?: boolean;
+  hideDuration?: boolean;
```
seyedmostafahasani, line 226: Please add the missing TypeScript to this file `src/specs/VideoNativeComponent.ts`
```
type ControlsStyles = Readonly<{
  hideSeekBar?: boolean;
  hideDuration?: boolean;
  seekIncrementMS?: Int32;
}>;
```

seyedmostafahasani:
@ashlyWeiting 
Please update the document accordingly.

ashlyWeiting:
Thanks for the review! I've made the updates you requested. Could you please take a look and let me know if everything looks good?

freeboub (approved):

seyedmostafahasani (approved):

seyedmostafahasani:
@KrzysztofMoch 
I think it's better to update the commit as shown below.
`feat(android): add support for hiding video duration`

WDYT?

KrzysztofMoch (approved):

PR #4101, opened 2024-08-19
fix(ios): build fail due to an unwrapped value

### LINKED ISSUE #4099: [BUG]: Value of optional type 'AVMediaSelectionGroup??' not unwrapped
### Version

6.4.3

iOS

### System Version

MacOS Sonoma (14.4.1)

Simulator

### Architecture

Old architecture

### What happened?

Hello, 

I am facing some issues when updating my iOS App platform version from `13.0` to `15.0`, this issue was not previously being triggered and it happens when I try to do a normal build on a Simulator

The error and code section is:
```
node_modules/react-native-video/ios/Video/Features/RCTVideoUtils.swift:12:20 Value of optional type 'AVMediaSelectionGroup??' not unwrapped; did you mean to use 'try!' or chain with '?'?
```
Captura de Tela 2024-08-19 às 11 31 48

My Xcode version is 15.4 and the simulator is an iPhone 15 Pro

I already did all pod related stuff like, `deintegrate`, `install -repo-update`, removed all DerivedData and Pods/ folder (also Podfile.lock), deleted the .xcode.env.local, but nothing changes this result

### Reproduction

no repository

### Reproduction

I don't know how you will be able to reproduce this issue since I am adding react-native into a native iOS project and not a common react-native project

Fausto95:
Fixes #4099

iOS builds were failing because of AVMediaSelectionGroup's value not being unwrapped.

We first have to check if `asset` is nil before proceeding to call its methods.

### DIFF
--- a/ios/Video/Features/RCTVideoUtils.swift
+++ b/ios/Video/Features/RCTVideoUtils.swift
@@ -9,6 +9,7 @@ enum RCTVideoAssetsUtils {
         for mediaCharacteristic: AVMediaCharacteristic
     ) async -> AVMediaSelectionGroup? {
         if #available(iOS 15, tvOS 15, visionOS 1.0, *) {
+            // swiftlint:disable shorthand_optional_binding
             guard let asset = asset else {
                 return nil
             }
@@ -17,6 +18,7 @@ enum RCTVideoAssetsUtils {
             } else {
                 return nil
             }
+            // swiftlint:enable shorthand_optional_binding
         } else {
             #if !os(visionOS)
                 return asset?.mediaSelectionGroup(forMediaCharacteristic: mediaCharacteristic)
--- a/ios/Video/Features/RCTVideoUtils.swift
+++ b/ios/Video/Features/RCTVideoUtils.swift
@@ -9,20 +9,30 @@ enum RCTVideoAssetsUtils {
         for mediaCharacteristic: AVMediaCharacteristic
     ) async -> AVMediaSelectionGroup? {
         if #available(iOS 15, tvOS 15, visionOS 1.0, *) {
-            return try? await asset?.loadMediaSelectionGroup(for: mediaCharacteristic)
+            guard let asset = asset else {
+                return nil
+            }
+            return try? await asset.loadMediaSelectionGroup(for: mediaCharacteristic)
         } else {
             #if !os(visionOS)
                 return asset?.mediaSelectionGroup(forMediaCharacteristic: mediaCharacteristic)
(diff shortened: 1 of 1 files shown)


freeboub (approved):

Fausto95:
imo the first [commit](https://github.com/TheWidlarzGroup/react-native-video/pull/4101/commits/88c5f3f8c782c35a9eb7267294c27365df2bb44a) should be the way to go but the linter yells at me to use shorthand optional binding

ios/Video/Features/RCTVideoUtils.swift
```diff
@@ -9,7 +9,16 @@ enum RCTVideoAssetsUtils {
         for mediaCharacteristic: AVMediaCharacteristic
     ) async -> AVMediaSelectionGroup? {
         if #available(iOS 15, tvOS 15, visionOS 1.0, *) {
-            return try? await asset?.loadMediaSelectionGroup(for: mediaCharacteristic)
+            // swiftlint:disable shorthand_optional_binding
```
freeboub: Is this swiftlint directive mandatory ? looks like you want to bypass something ?

Fausto95: If we don't disable the rule for these specific lines, the linter forces us to use optional binding, which results in: 

```swift
try? await asset?.loadMediaSelectionGroup(for: mediaCharacteristic)
```

so the code gets unchanged and we still have errors because of the unwrapped value.

ios/Video/Features/RCTVideoUtils.swift
```diff
@@ -9,7 +9,16 @@ enum RCTVideoAssetsUtils {
...
+            guard let asset = asset else {
+                return nil
+            }
+            if let mediaSelectionGroup = try? await asset.loadMediaSelectionGroup(for: mediaCharacteristic) {
+                return mediaSelectionGroup
+            } else {
+                return nil
+            }
+            // swiftlint:enable shorthand_optional_binding
```
KrzysztofMoch: @Fausto95 How about this ?
```suggestion
            do {
                guard let asset else {
                    return nil
                }

                return try await asset.loadMediaSelectionGroup(for: mediaCharacteristic)
            } catch {
                return nil
            }
```

Fausto95: Yeah, should work too..
I'll apply the changes and see if the linter won't error.

Fausto95:
@KrzysztofMoch my app builds successfully. we're good to go

Issue #4039, opened 2024-07-29
[BUG]: (Android) HLS in background stops and shows buffering animation forever in example/basic app on physical devices


paulrinaldi:
### Version

^6.2.4

### What platforms are you having the problem on?

Android

### System Version

14

### On what device are you experiencing the issue?

Real device

### Architecture

Old architecture

### What happened?

Using an HLS livestream such as this [Democracy Now DVR stream](https://democracynow-hls.secdn.net/democracynow-live/play/democracynow.smil/playlist.m3u8) (available from 8-9am United States Eastern Time) or
this [DVR FPV Drone stream](https://5b44cf20b0388.streamlock.net:8443/dvr/dvr/playlist.m3u8?DVR) (always available) in the `example/basic` app on a physical device does not play properly in the background.

### Reproduction

https://github.com/paulrinaldi/react-native-video/tree/master

1. Start the app on a physical android 14 device
2. Navigate channel up to one channel past the red bull stream
3. Wait until the video loads
4. Background the app
5. Pull down the notification controls so you can observe AV (not necessarily needed, you can reproduce this without viewing the notif media controls)
6. * Watch the notification's progress bar move until the circle hits the end of the progress bar and switches out the play/pause icon for the buffering animation 

Actual:
7. Observe no AV and the buffering animation continue nonstop*

Desired:
7. Observe the playable duration increase and the progress bar's ticker move backward slightly because the new buffered section loaded in the progress bar and no interruption in the audio while backgrounded.

OR

Add either or both of these entries to `examples/basic/src/constants/general.ts` this DVR HLS:
```
  {
    description: 'democracy now',
    uri: 'https://democracynow-hls.secdn.net/democracynow-live/play/democracynow.smil/playlist.m3u8',
  },
  {
    description: 'Drone B-Roll',
    uri: 'https://5b44cf20b0388.streamlock.net:8443/dvr/dvr/playlist.m3u8?DVR',
  },
```

Actual:
1. Start the app on a physical android 14 device
2. Tap notification controls
3. Navigate channel up to the added source
4. Background the app
5. Pull down the notification controls so you can observe AV (not necessarily needed, you can reproduce this without viewing the notif media controls)
6. Observe no AV and the buffering animation continue nonstop*

Desired:
7. Observe the playable duration increase and the progress bar's ticker move backward slightly because the new buffered section loaded in the progress bar and no interruption in the audio while backgrounded.

Other notes:
[...]

PR #4105, opened 2024-08-21
fix(android): HLS DVR media should update playable duration in the background

paulrinaldi:
e.g. wait until onDestroy

When unmounting a component holding the the videoplayer from RNV component, I saw an error that crashed my app: deleting the notification channel while the foreground service is still running is not allowed.

## Summary
Use service starting methods properly. Remove notification channel properly.

### Motivation
The exploration started from looking in to #4039 which describes that HLS DVR streams stop loading playable duration when in the backgrond.
Add `startForegroundService` since the foreground service seemed to not be working correctly, warning logs I saw about buffer deallocations and windows closing and observations above in #4039, and  https://stackoverflow.com/questions/45525214/are-there-any-benefits-to-using-context-startforegroundserviceintent-instead-o. Add `startForeground` (required by using `startForegroundService`).

### Changes
See PR diff.

## Test plan
Test on physical android devices with the diff below and HLS DVR streams ("https://democracynow-hls.secdn.net/democracynow-live/play/democracynow.smil/playlist.m3u8", "https://5b44cf20b0388.streamlock.net:8443/dvr/dvr/playlist.m3u8?DVR").

Open the `examples/basic` videoplayer project, edit with this diff:
[background.patch](https://github.com/user-attachments/files/16692970/background.patch)
Also add HLS DVR streams to `general.ts` to view other than the HLS nonDVR redbull stream.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
@@ -63,6 +63,7 @@ class VideoPlaybackService : MediaSessionService() {
 
         mediaSessionsList[player] = mediaSession
         addSession(mediaSession)
+        startForeground(mediaSession.player.hashCode(), buildNotification(mediaSession))
     }
 
     fun unregisterPlayer(player: ExoPlayer) {
@@ -95,6 +96,10 @@ class VideoPlaybackService : MediaSessionService() {
 
     override fun onDestroy() {
         cleanup()
+        val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager.deleteNotificationChannel(NOTIFICATION_CHANEL_ID)
+        }
         super.onDestroy()
     }
 
@@ -209,9 +214,6 @@ class VideoPlaybackService : MediaSessionService() {
     private fun hideAllNotifications() {
         val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
         notificationManager.cancelAll()
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-            notificationManager.deleteNotificationChannel(NOTIFICATION_CHANEL_ID)
-        }
     }
 
     private fun cleanup() {
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -923,7 +923,11 @@ public class ReactExoplayerView extends FrameLayout implements
         Intent intent = new Intent(themedReactContext, VideoPlaybackService.class);
         intent.setAction(MediaSessionService.SERVICE_INTERFACE);
 
-        themedReactContext.startService(intent);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            themedReactContext.startForegroundService(intent);
+        } else {
+            themedReactContext.startService(intent);
+        }
 
         int flags;
         if (Build.VERSION.SDK_INT >= 29) {


android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -95,6 +96,10 @@ class VideoPlaybackService : MediaSessionService() {
 
     override fun onDestroy() {
         cleanup()
+        val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
```
freeboub, line 99: won't it be better to move this part of code inside cleanup function ? clean up is used more intensivelly then onDestroy ...

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -63,6 +63,7 @@ class VideoPlaybackService : MediaSessionService() {
 
         mediaSessionsList[player] = mediaSession
         addSession(mediaSession)
+        startForeground(mediaSession.player.hashCode(), buildNotification(mediaSession))
```
freeboub, line 66: startForground need a specific permission starting from android P:

Apps targeting API Build.VERSION_CODES.P or later must request the permission android.Manifest.permission.FOREGROUND_SERVICE in order to use this API.

I think you need at least to update the documentation to indicate to enable this option

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
paulrinaldi: I think that would cause a problem because the foreground service/notification could be still alive when cleanup() is ran. I can test this though and report back.

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
paulrinaldi: Already present, good eye, though! https://thewidlarzgroup.github.io/react-native-video/component/props#shownotificationcontrols

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
freeboub: After double check, this code doesn't seem necessary:
cleanup() function calls hideAllNotifications() which is doing exactly the same code  + notificationManager.cancelAll()

I think it can just be removed in fact

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -95,6 +96,10 @@ class VideoPlaybackService : MediaSessionService() {
 
     override fun onDestroy() {
         cleanup()
+        val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager.deleteNotificationChannel(NOTIFICATION_CHANEL_ID)
+        }
```
freeboub, line 102:
```suggestion
```

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
freeboub: In fact I am disturbed by the logic.
Notification channel is created in onUpdateNotification and destroyed in onDestroy

For exemple shouldn't you add this code also in : (which calls cleanup)
unregisterPlayer(player: ExoPlayer) { 

in fact (maybe), you need to replace cleanup() by hideAllNotifications() in onTaskRemoved and unregisterPlayer() ?

Notice that if you unregisterPlayer and registerPlayer just after the notification channel is not re created...

onDestroy seems ok for me to cleanup.

freeboub: @paulrinaldi gentle reminder on this point, It would be great to merge this PR for next release !

freeboub:
should decrease repeatability of https://github.com/TheWidlarzGroup/react-native-video/issues/3897

cordovalegacy:
@freeboub @paulrinaldi my team has had tons of users reporting this issue, could we please get this fix merged so we can update with the fix? Thanks

jdredman:
Our app is experiencing this issue and we really need this to improve our UX. Any idea when we can expect a fix?

roliverplate:
Hi,
I’d like to emphasize the importance of this fix. We're relying on this update to resolve a key issue in our app, and we're currently facing a critical deadline. The fix is blocking us, so we’d greatly appreciate it if this could be prioritized ASAP.
Thanks for all your hard work!

freeboub:
I have one remaining comment, but I am Ok to merge it.
@KrzysztofMoch can you please review this PR ?

paulrinaldi:
Currently using a patch with this as the fix. In a submission to the play store, we did see some crashes when a test user from google play seemed to create a race between reusing and destroying a media session and when trying to play two videos at once. I would recommend we carefully consider the destruction and creation method details.

freeboub:
@paulrinaldi Thank you for the feedback! I think the 2 video in parallel is not necessary a usual use case.
If you confirm this patch is better than master branch let's go with it. (I think it is better, just the last comment disturbing me)

KrzysztofMoch (approved):
It Okay for me, would be nice to resolve @freeboub comment but not required for this PR

paulrinaldi:
The patch I'm using is still equivalent to the changes in these commits.

PR #4122, opened 2024-08-29
chore(sample): additionnal sample cleanup

freeboub:
## Summary
The goal of this PR is to remove circular dependency in the sample app which is causing a warning at every startup

### Motivation
The goal of this PR is to remove circular dependency in the sample app which is causing a warning at every startup

### Changes
move 2 files in components
rework imports / exports
fix linter warning

## Test plan
test on sample


examples/basic/src/components/Overlay.tsx
```diff
@@ -23,14 +21,15 @@ import {
...
-} from '../components';
+
+import {toast} from './Toast';
+import {Seeker} from './Seeker';
+import {AudioTrackSelector} from './AudioTracksSelector';
+import {VideoTrackSelector} from './VideoTracksSelector';
+import {TextTrackSelector} from './TextTracksSelector';
+import {TopControl} from './TopControl';
+import {ToggleControl} from './ToggleControl';
+import {MultiValueControl} from './MultiValueControl';
```
seyedmostafahasani, line 32: I think it's better to import as shown below:
```
import {
  toast,
  Seeker,
  AudioTrackSelector,
  TextTrackSelector,
  VideoTrackSelector,
  TopControl,
  ToggleControl,
   MultiValueControl
} from '../components';
```

seyedmostafahasani:
I just left a small comment.

examples/basic/src/components/Overlay.tsx
freeboub: But will produce the circular dependency issue.. as ../component/index importss Overlay, Overlay will try to import itself...

freeboub: But, maybe overlay shouldn't be in component?!

seyedmostafahasani: You're right. I approved it.

seyedmostafahasani (approved):

PR #4123, opened 2024-08-30
fix(ios): ensure behavior is correct with empty text track list

### LINKED ISSUE #4044: [BUG]: Setting textTracks to empty array when playing a mp3 file in videoplayer breaks the player in example/basic
### Version

6.4.3

iOS

### System Version

17.4

Simulator

### Architecture

Old architecture

### What happened?

Likely also happens on real devices.

Supply prop `textTracks={[]}`and an mp3 for the source uri, and watch the player behave innapropriately. In my app, I was onProgress was getting called with currentTime equal to the end of the video (on 6.4.2), in the videoplayer example/basic an error was logged and the app navigated back to the first source, and in a sample app I made with expo it seemed to load forever. 

What I would expect to happen is either:
1. The error is caught in js and gets propagated to the user and we see a stacktrace of an error like "mp3 doesn't allow textTracks"
2. RNV's Video player silently ignores texttracks when mp3 is used
And probably:
3. A note in the documentation about not setting textTracks on audio uri's (though mp4a's are handled just fine by iOS with subtitles)

Other notes:
One can supply textTracks={[]} with a source of a mp4a uri and this inappropriate behavior doesn't occur.

### Reproduction

repository link

### Reproduction

Edit `example/basic`'s `src/constants/general.ts` `srcAllPlatformList` to include:
```
  {
    description: 'mp3 with texttrack',
    uri: 'https://traffic.libsyn.com/democracynow/wx2024-0702_SOT_DeadCalm-LucileSmith-FULL-V2.mxf-audio.mp3', // an mp3 file, necessary for reproduction
    textTracks: [] // necessary for reproduction
  },
```
Navigate to the video, observe:
 LOG  {"error":{"domain":"AVFoundationErrorDomain","localizedFailureReason":"This media format is not supported.","localizedDescription":"Cannot Open","code":-11828,"localizedRecoverySuggestion":""},"target":3}

freeboub:
## Summary
Ensure textTracks list behavior is correct when the list is empty 

### Motivation
fix: https://github.com/TheWidlarzGroup/react-native-video/issues/4044

### Changes
To fix the issue, I rework external tracks list management to ensure is it never null:
- It simplify the code (avoid ?.)
- it simplify tests to check if app should handle external tracks (just need to check if it is empty instead of != nill && !empty) 

## Test plan
Can be tested with sample, I added the mp3 track.
important to test: you need to comment: export const textTracksSelectionBy = 'index';
as selection by index doesn't work on sideLoaded tracks

### DIFF
--- a/examples/basic/src/constants/general.ts
+++ b/examples/basic/src/constants/general.ts
@@ -79,6 +79,11 @@ export const srcAllPlatformList = [
     startPosition: 50000,
   },
   {
+    description: 'mp3 with texttrack',
+    uri: 'https://traffic.libsyn.com/democracynow/wx2024-0702_SOT_DeadCalm-LucileSmith-FULL-V2.mxf-audio.mp3', // an mp3 file
+    textTracks: [], // empty text track list
+  },
+  {
     description: 'BigBugBunny sideLoaded subtitles',
     // sideloaded subtitles wont work for streaming like HLS on ios
     // mp4
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -979,7 +979,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     }
 
     func setTextTracks(_ textTracks: [TextTrack]?) {
-        if (textTracks == nil) {
+        if textTracks == nil {
             _textTracks = []
         } else {
             _textTracks = textTracks!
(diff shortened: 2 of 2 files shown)


seyedmostafahasani:
@freeboub 
What do you think about adding a comment for the `textTracksSelectionBy` variable?

freeboub:
> @freeboub What do you think about adding a comment for the `textTracksSelectionBy` variable?

You are totally right !
I added it in https://github.com/TheWidlarzGroup/react-native-video/pull/4129
Thank you for the feedback !

seyedmostafahasani (approved):

KrzysztofMoch (approved):

PR #4133, opened 2024-09-01
fix(android): add subtitleStyle.subtitlesFollowVideo prop to control subtitles positionning

freeboub:
## Summary
add subtitleStyle.subtitlesFollowVideo prop to control subtitles positionning
2 options: follow video or follow main view

Here is description of the new property:

`subtitlesFollowVideo` helps to determine how the subtitles are positionned.
To understand this prop you need to understand how views management works.
The main View style passed to react native video is the position reserved to display the video component.
It may not match exactly the real video size.
For exemple, you can pass a 4:3 video view and render a 16:9 video inside.
So there is a second view, the video view.

Subtitles are managed in a third view.

First react-native-video resize the video to keep aspect ratio (depending on `resizeMode` property) and put it in main view.

* When putting subtitlesFollowVideo to true, the subtitle view will be adapt to the video view.
It means that if the video is displayed out of screen, the subtitles may also be displayed out of screen.

* When putting subtitlesFollowVideo to false, the subtitle view will keep adapting to the main view.
It means that if the video is displayed out of screen, the subtitles may also be displayed out of screen.

This prop can be changed on runtime.

### Motivation
Fix following regression: https://github.com/TheWidlarzGroup/react-native-video/issues/4019 (will be fixed by default)
Regression introduced by following PR: https://github.com/TheWidlarzGroup/react-native-video/pull/3830
Linked tickets: https://github.com/TheWidlarzGroup/react-native-video/issues/3662 and https://github.com/TheWidlarzGroup/react-native-video/issues/4043

### Changes
dynamically configure subtitle position with a new prop added in subtitleStyle

## Test plan
Can be tested with the sample. I didn't add a button to trigger change, but I tested with manually change the value

Note, it is not implemented on ios, I think it should ...

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -114,7 +114,7 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
     }
 
     public void setSubtitleStyle(SubtitleStyle style) {
-        // ensure we reset subtile style before reapplying it
+        // ensure we reset subtitle style before reapplying it
         subtitleLayout.setUserDefaultStyle();
         subtitleLayout.setUserDefaultTextSize();
 
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
@@ -48,6 +48,8 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
     private @ViewType.ViewType int viewType = ViewType.VIEW_TYPE_SURFACE;
     private boolean hideShutterView = false;
 
+    private SubtitleStyle localStyle = SubtitleStyle.parse(null);
+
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.java
```diff
@@ -48,6 +48,8 @@ public final class ExoPlayerView extends FrameLayout implements AdViewProvider {
     private @ViewType.ViewType int viewType = ViewType.VIEW_TYPE_SURFACE;
     private boolean hideShutterView = false;
 
+    private SubtitleStyle localStyle = SubtitleStyle.parse(null);
```
seyedmostafahasani: I think it would make more sense to initialize `localStyle` in the constructor.
WDYT?

freeboub: You are right, it makes more sense ! This is fixed now

seyedmostafahasani (approved):

KrzysztofMoch (approved):

PR #4143, opened 2024-09-03
feat: Correct isBehindLiveWindow Error Handling

whdudtod1273:
## Summary
```
@Override
    public void onPlayerError(@NonNull PlaybackException e) {
        String errorString = "ExoPlaybackException: " + PlaybackException.getErrorCodeName(e.errorCode);
        String errorCode = "2" + e.errorCode;
        switch(e.errorCode) {
            case PlaybackException.ERROR_CODE_DRM_DEVICE_REVOKED:
            case PlaybackException.ERROR_CODE_DRM_LICENSE_ACQUISITION_FAILED:
            case PlaybackException.ERROR_CODE_DRM_PROVISIONING_FAILED:
            case PlaybackException.ERROR_CODE_DRM_SYSTEM_ERROR:
            case PlaybackException.ERROR_CODE_DRM_UNSPECIFIED:
                if (!hasDrmFailed) {
                    // When DRM fails to reach the app level certificate server it will fail with a source error so we assume that it is DRM related and try one more time
                    hasDrmFailed = true;
                    playerNeedsSource = true;
                    updateResumePosition();
                    initializePlayer();
                    setPlayWhenReady(true);
                    return;
                }
                break;
            default:
                break;
        }
        eventEmitter.onVideoError.invoke(errorString, e, errorCode);
        playerNeedsSource = true;
        if (isBehindLiveWindow(e)) {
            clearResumePosition();
            initializePlayer();
        } else {
            updateResumePosition();
        }
    }
```

Currently, when the isBehindLiveWindow error (ERROR_CODE_BEHIND_LIVE_WINDOW) occurs in onPlayerError, initializePlayer is executed. However, when initializePlayer is executed, the player is reinitialized, causing it to reload, which results in flickering and feels unnatural.

I believe this error may occur frequently in 3G environments or when the network is poor. In such cases, it would be better to resume playback from the latest live position rather than reinitializing the player.

### Changes
Then, instead of initializePlayer(), I will execute player?.prepare() and player?.seekToDefaultPosition() to resume the live stream.

## Test plan
This is an issue that occurs when playing HLS live streaming in low network mode on Android.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1742,7 +1742,8 @@ public class ReactExoplayerView extends FrameLayout implements
         playerNeedsSource = true;
         if (isBehindLiveWindow(e)) {
             clearResumePosition();
-            initializePlayer();
+            player?.seekToDefaultPosition();
+            player?.prepare();
         } else {
             updateResumePosition();
         }
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1742,8 +1742,10 @@ public class ReactExoplayerView extends FrameLayout implements
         playerNeedsSource = true;
         if (isBehindLiveWindow(e)) {
             clearResumePosition();
-            player?.seekToDefaultPosition();
-            player?.prepare();
+            if (player != null) {
+                player.seekToDefaultPosition();
+                player.prepare();
+            }
         } else {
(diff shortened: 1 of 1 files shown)


freeboub:
I agree with this change, I will merge it.
BTW the behavior may still change in the future.
To give you more background, There are 2 ways to generate this error with live contents:
- The use case you describe, having a very short manifest (only few chunks inside) and connectivity issues can cause the problem -> Your fix is well addressing the issue! thank you.
- Another can happen when you have a pausable live stream. maybe you live is 15min long. in that case it is also reproduced after 15min of pause. This use case is a bit different and I am not 100% sure you put the best behavior here.
Anyway my point is just to keep in mind. Thank you

freeboub (approved):

PR #4148, opened 2024-09-05
fix(sample): remove warning on ios with NavigationBar

freeboub:
## Summary
fix(sample): remove warning on ios with NavigationBar, It is not defined in ios

### Motivation
Keep sample clean without warning

### Changes
call the function only on android

## Test plan
launch ios sample

### DIFF
--- a/examples/basic/src/VideoPlayer.tsx
+++ b/examples/basic/src/VideoPlayer.tsx
@@ -34,7 +34,7 @@ import Video, {
 } from 'react-native-video';
 import styles from './styles';
 import {type AdditionalSourceInfo} from './types';
-import {bufferConfig, srcList, textTracksSelectionBy} from './constants';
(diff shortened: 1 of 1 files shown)


examples/basic/src/VideoPlayer.tsx
```diff
@@ -106,7 +106,9 @@ const VideoPlayer: FC<Props> = ({}) => {
   }, [goToChannel, srcListId]);
 
   useEffect(() => {
-    NavigationBar.setVisibilityAsync('visible');
+    if (Platform.OS === 'android') {
```
seyedmostafahasani: We can use the variable `isAndroid` instead of `Platform.OS === 'android'` throughout the entire app, or use `isIos` where applicable.

freeboub: Right, This is done, thank you !

seyedmostafahasani (approved):

PR #4160, opened 2024-09-11
chore(android): move contentStartTime into source prop

freeboub:
## Summary
 move contentStartTime into source prop

### Motivation
fix next item for ticket: https://github.com/TheWidlarzGroup/react-native-video/issues/2693

### Changes
Just an internal refactor + doc clean up

## Test plan
N/A

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1428,7 +1428,7 @@ public class ReactExoplayerView extends FrameLayout implements
             ArrayList<Track> audioTracks = getAudioTrackInfo();
             ArrayList<Track> textTracks  = getTextTrackInfo();
 
-            if (source.getContentStartTime() != -1L) {
+            if (source.getContentStartTime() != -1) {
                 ExecutorService es = Executors.newSingleThreadExecutor();
                 es.execute(() -> {
                     // To prevent ANRs caused by getVideoTrackInfo we run this on a different thread and notify the player only when we're done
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -1429,7 +1428,7 @@ private void videoLoaded() {
             ArrayList<Track> audioTracks = getAudioTrackInfo();
             ArrayList<Track> textTracks  = getTextTrackInfo();
 
-            if (this.contentStartTime != -1L) {
+            if (source.getContentStartTime() != -1L) {
```
seyedmostafahasani:
```suggestion
     source.getContentStartTime() != -1
```
Because contentStartTime is Int.

src/specs/VideoNativeComponent.ts
```diff
@@ -344,7 +345,6 @@ export interface VideoNativeProps extends ViewProps {
   debug?: DebugConfig;
   showNotificationControls?: WithDefault<boolean, false>; // Android, iOS
   bufferConfig?: BufferConfig; // Android
-  contentStartTime?: Int32; // Android
```
seyedmostafahasani, line 347: I think it's better to revert it and add a comment, similar to the one in the file: `src/types/video.ts`
`/** @deprecated Use source.contentStartTime */`
WDYT?

KrzysztofMoch: This is for native component - users does not see this as it is only for codegen (that we are not using yet)

KrzysztofMoch (approved):

seyedmostafahasani:
@freeboub 
This branch has conflicts.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: yes make sense ! Done

seyedmostafahasani (approved):

PR #4171, opened 2024-09-14
fix(JS): improve loader api

freeboub:
## Summary
Improve renderLoader API
update api to allow function prop in renderLoader instead of Element.

### Motivation
It is a performence motivation
With element, the element is created on every render.
With function, the element is rendered only when needed

### Changes
Only JS changes

## Test plan
Can be tested with the sample


KrzysztofMoch (approved):
Looks fine for me

Issue #4167, opened 2024-09-12
[BUG]: Expo Config Plugin for Notification Controls adds duplicate <service> tag to AndroidManifest on every prebuild


aminmeshk:
### Version

6.5.0

### What platforms are you having the problem on?

Android

### System Version

macOs 14.1

### On what device are you experiencing the issue?

Simulator

### Architecture

Old architecture

### What happened?

When I set `"enableNotificationControls": true` in my `app.json` and run `npx expo prebuild`, it adds the code below to the `AndroidManifest.xml` file:
```xml
    <service android:name="com.brentvatne.exoplayer.VideoPlaybackService" android:exported="false" android:foregroundServiceType="mediaPlayback">
      <intent-filter>
        <action android:name="androidx.media3.session.MediaSessionService"/>
      </intent-filter>
    </service>
```
which is the expected behavior. The issue is that if I run `npx expo prebuild` again, it adds the full tag to the manifest once more! It adds it every time I run `prebuild` without cleaning.

### Reproduction

Step to reproduce this bug are: 
1. Add `react-native-video` package to a Expo project that uses [Prebuild](https://docs.expo.dev/workflow/prebuild/)
2. In `app.json` file, add `"react-native-video"` to the `plugins` array like with `"enableNotificationControls": true` parameter. 
```json
"plugins": [
      [
        "react-native-video",
        {
          "enableNotificationControls": true
        }
      ]
]
```
3. Run `npx expo prebuild`
4. Check `AndroidManifest.xml`. The code below should be added to the manifest:
```xml
    <service android:name="com.brentvatne.exoplayer.VideoPlaybackService" android:exported="false" android:foregroundServiceType="mediaPlayback">
      <intent-filter>
        <action android:name="androidx.media3.session.MediaSessionService"/>
      </intent-filter>
    </service>
```
5. Run `npx expo prebuild` again.
6. Check `AndroidManifest.xml`. You can see there are now two of the exact same tag:
```xml
    <service android:name="com.brentvatne.exoplayer.VideoPlaybackService" android:exported="false" android:foregroundServiceType="mediaPlayback">
      <intent-filter>
        <action android:name="androidx.media3.session.MediaSessionService"/>
      </intent-filter>
    </service>
    <service android:name="com.brentvatne.exoplayer.VideoPlaybackService" android:exported="false" android:foregroundServiceType="mediaPlayback">
      <intent-filter>
[...]

PR #4172, opened 2024-09-14
fix(expo-plugin): add check for existing service in AndroidManifest for notification controls 

aminmeshk:
## Summary
This PR fixes a minor issue with Expo Config Plugins (the one related to notification controls) that caused having duplicate `<service>` tags in the generated `AndroidManifest.xml` file when running non-cleaning `npx expo prebuild` multiple times.

### Motivation
Fix this issue: #4167 

### Changes
 I added a check in `withNotificationControls.ts` file for existing service in to avoid having duplicate code in `AndroidManifest.xml`.

## Test plan

### DIFF
--- a/src/expo-plugins/withNotificationControls.ts
+++ b/src/expo-plugins/withNotificationControls.ts
@@ -23,7 +23,10 @@ export const withNotificationControls: ConfigPlugin<boolean> = (
       if (!application.service) {
         application.service = [];
       }
-
+// We check if the VideoPlaybackService is already defined in the AndroidManifest.xml
+// to prevent adding duplicate service entries. If the service exists, we will remove 
+// it before adding the updated configuration to ensure there are no conflicts or redundant
+// service declarations in the manifest.
       const existingServiceIndex = application.service.findIndex(
         (service) =>
           service?.$?.['android:name'] ===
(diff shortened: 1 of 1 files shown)


freeboub (approved):

freeboub:
Looks easy :) Thank you for the patch !

src/expo-plugins/withNotificationControls.ts
```diff
@@ -24,6 +24,15 @@ export const withNotificationControls: ConfigPlugin<boolean> = (
         application.service = [];
       }
```
seyedmostafahasani, line 26: @freeboub
I think it's better to add a log to explain this logic.
```suggestion
// We check if the VideoPlaybackService is already defined in the AndroidManifest.xml
// to prevent adding duplicate service entries. If the service exists, we will remove 
// it before adding the updated configuration to ensure there are no conflicts or redundant
// service declarations in the manifest.
```
WDYT?

seyedmostafahasani (approved):

seyedmostafahasani:
@aminmeshk 
Would you please fix the lint errors?

freeboub:
> @aminmeshk 
> Would you please fix the lint errors?

You miss the spaces in your proposal... I commit it without thinking of it 🙄

aminmeshk:
I fixed the spaces

KrzysztofMoch (approved):
Thanks!

PR #4174, opened 2024-09-14
chore(android): add null check for id of videoFormat

seyedmostafahasani:
## Summary
Add a null check for `videoFormat.id`.

### Motivation
Fix a potential issue where `videoFormat.id` could be null.

## Test plan
The sample app should no longer crash when playing RTSP video types.

### DIFF
--- a/android/src/main/java/com/brentvatne/common/react/VideoEventEmitter.kt
+++ b/android/src/main/java/com/brentvatne/common/react/VideoEventEmitter.kt
@@ -355,5 +355,4 @@ class VideoEventEmitter {
 
             putString("orientation", orientation)
         }
-
 }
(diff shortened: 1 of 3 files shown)


freeboub (commented):
This track id is not documented in onLoad... It would make sense to document it !

In a more general matter, don't you think it would be better to fix it inside onVideoBandwidthUpdate and onVideoLoad.
If track id is provided, add the field, if null do put the field in the event (and the same for height/width).
These kind of default/fallback values are coming from the hat and doesn't really make sens ', 0, 0, "-1" '

that said, I think this code:
`
            int width = videoFormat != null ? (isRotatedContent ? videoFormat.height : videoFormat.width) : 0;
            int height = videoFormat != null ? (isRotatedContent ? videoFormat.width : videoFormat.height) : 0;
``
should also be present before onVideoBandwidthUpdate call ...
I know it is not your code...

Thank you.

freeboub:
small safety PR, 3 feedbacks around the code, sorry ...

freeboub (approved):

seyedmostafahasani:
@freeboub 
Thank you for your suggestions!

KrzysztofMoch (approved):

PR #4176, opened 2024-09-15
chore(android): refactor audio focus

freeboub:
## Summary
WIP
refactor audio focus listener to kotlin

### Changes
Move audio focus management in a dedicated kotlin file
create RNVPlayerInterface to be able to handle generic code
Add lazy init of audioFocusDelegate (will not be instanciated if not necessary)

## Test plan


android/src/main/java/com/brentvatne/common/toolbox/AudioManagerDelegate.kt
```diff
@@ -0,0 +1,106 @@
...
+import android.media.AudioManager
+import android.media.AudioManager.OnAudioFocusChangeListener
+import com.brentvatne.common.api.RNVPlayerInterface
+import com.facebook.react.uimanager.ThemedReactContext
+
+/**
+ * Delegate audio management to this class
+ * This is an helper to group all generic android code which do not depend on player implementation
+ */
+class AudioManagerDelegate(player: RNVPlayerInterface, themedReactContext: ThemedReactContext) {
```
seyedmostafahasani, line 16: We can add `private val` because they are not expected to change after initialization.
```suggestion
class AudioManagerDelegate(private val player: RNVPlayerInterface, private val themedReactContext: ThemedReactContext
)```

PR #4183, opened 2024-09-19
feat(android): enable to hide specific controls

### LINKED ISSUE #3193: Provide props to hide specific controllers when control `true`
# Feature Request

## Why it is needed
 In certain cases, when we only need basic functions like play, pause, and seek for video control, so in that case we have two options: customizing the controls ourselves or relying on an external library like react-native-video-control.

moskalakamil:
## Summary
Create new feature which allow user to hide specific controls on android devices

Closes #3193
### Motivation
#3193 
### Changes
update function to allow hide more controls
## Test plan


src/specs/VideoNativeComponent.ts
```diff
@@ -296,11 +296,18 @@ export type OnAudioFocusChangedData = Readonly<{
 }>;
 
 type ControlsStyles = Readonly<{
+  hidePosition?: WithDefault<boolean, false>;
+  hidePlayPause?: WithDefault<boolean, false>;
+  hideForward?: WithDefault<boolean, false>;
+  hideRewind?: WithDefault<boolean, false>;
+  hideNext?: WithDefault<boolean, false>;
```
freeboub, line 303: I wonder if the next and previous buttons should be hidden by default, they are always displayed but never usable 🤔 @KrzysztofMoch @seyedmostafahasani  what are your point of views ?

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -525,38 +527,85 @@ private void reLayout(View view) {
...
+        }
+
+        final ImageButton previousButton = playerControlView.findViewById(R.id.exo_prev);
+        if (controlsConfig.getHidePrevious()) {
+            previousButton.setImageAlpha(0);
+            previousButton.setClickable(false);
+        } else {
+            previousButton.setImageAlpha(255);
+            previousButton.setClickable(true);
+        }
```
moskalakamil, line 581: setVisibility doesn't work correctly for these items

freeboub (approved):

src/specs/VideoNativeComponent.ts
seyedmostafahasani: I agree with hiding the previous and next buttons by default. However, I don't understand why we would allow the user to hide the play and pause buttons.

KrzysztofMoch: Background videos - it is just an option
> I wonder if the next and previous buttons should be hidden by default

I agree we can hide them by default

KrzysztofMoch (approved):
Good Job @moskalakamil!

seyedmostafahasani (approved):

src/specs/VideoNativeComponent.ts
freeboub: @seyedmostafahasani depending of UI requierements, it make sense (live content, or not pausable video , why not :) )
@moskalakamil can you just change hideNext and  hidePrevious to true

seyedmostafahasani: @freeboub 
Thank you for the explanation.
@moskalakamil 
Thank you for creating this PR.

PR #4185, opened 2024-09-19
feat: add setSource API function fix ads playback

### LINKED ISSUE #3928: [BUG]:  Ad control not working on android
### Version

"react-native-video": "6.1.2"

Android

### System Version

Any android version

Real device, Simulator

### Architecture

Old architecture

### What happened?

- The ad overlay with play/pause button and skip ad doesn't appear always
- When playing an ad on `android`, the adEvents `PAUSED` and `RESUMED` get triggered, but do not actually pause or resume the ad. The issue is only with ads.

### Reproduction

_No response_

### Reproduction

Try any video on android with ads.

freeboub:
## Summary
feat: Add setSource API
fix(android): playback start up

### Motivation
Add setSource API, linked to: https://github.com/TheWidlarzGroup/react-native-video/issues/3352
fix: https://github.com/TheWidlarzGroup/react-native-video/issues/3928

### Changes
Add setSource command: easy on ios

On Android:
I moved management of playback stop from : ReactExoplayerViewManager.kt to ReactExoplayerView.java
I added concept of running source. I allows to avoid race conditions. during asynchronious process of startup, we are now stopping it if start has been canceled

For ads, I avoid releasing this adsLoader when the player is not destroyed.

## Test plan
Sample has been patched to integrate this source change


freeboub:
@KrzysztofMoch If you want to do a first review on this PR it would be great, There is a remaining resize glitch on sample when we change video on android, I will fix that.

freeboub:
> @KrzysztofMoch If you want to do a first review on this PR it would be great, There is a remaining resize glitch on sample when we change video on android, I will fix that.

OK, this is fixed !

seyedmostafahasani:
@freeboub 
Thanks for this PR. I really like it! 
Nice work!

seyedmostafahasani (approved):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -772,9 +776,9 @@ private void initializePlayerCore(ReactExoplayerView self) {
         player.addListener(self);
         player.setVolume(muted ? 0.f : audioVolume * 1);
         exoPlayerView.setPlayer(player);
-        if (adsLoader != null) {
-            adsLoader.setPlayer(player);
-        }
+
+        adsLoader.setPlayer(player);
+
```
KrzysztofMoch: We can't have this without checks, if we disable ads (in example they are enabled by default!) this will cause error - but my past _me_ left us an great option
```
if (BuildConfig.USE_EXOPLAYER_IMA) {
    adsLoader.setPlayer(player);
}
```
With this we can check if ads are enabled

KrzysztofMoch (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: All Ads related "things" we can wrap with this but it's optional I guess - only setPlayer cause playback error

KrzysztofMoch:
PR Looks Good For Me - only this ADS thing need to be fixed and we can merge

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: Yep I know, but as this is just empty function in the stub file, I think this is not necessary, it will just overload the code... (I made the choice not to add it). Do you really think I should add these additionnel tests ?
The stub will need to be implemented anyway ...

freeboub: OK @KrzysztofMoch I handled your point in: https://github.com/TheWidlarzGroup/react-native-video/pull/4220

freeboub:
@KrzysztofMoch Done, thank you for the feedback !

KrzysztofMoch (approved):

PR #4186, opened 2024-09-20
fix onAudioTracks : selected payload

bsbimalbose:
## Summary
Fix selected payload of onAudioTracks event

### Motivation
The selected payload of the onAudioTracks event was always coming as false, making it impossible to know which audio track is currently selected.

### Changes
change trackIndex to 0 for the isTrackSelected call

## Test plan


freeboub:
If I remember well I changed that to fix another issue :(
Can you provide the sample to reproduce the issue, I would like to triple check !

PR #4190, opened 2024-09-21
feat(android): add live video label

seyedmostafahasani:
## Summary
Display a `Live` label for live videos.

### Motivation
This allows the user to easily identify whether the video is live.

## Test plan
Play a live video in the sample app to verify the label appears correctly.


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -525,41 +524,40 @@ private void reLayout(View view) {
...
+        // Determine if the content is live
+        if (!timeline.isEmpty()) {
+            Timeline.Window window = new Timeline.Window();
+            timeline.getWindow(player.getCurrentMediaItemIndex(), window);
+            isLive = window.isLive();
         }
+
+        // Handle visibility for live content or hideSeekBar
+        if (controlsConfig.getHideSeekBar() || isLive) {
+            exoProgress.setVisibility(View.INVISIBLE);
```
freeboub: When you have a live content, you cannot always hide the progress bar.
You can have live with timeshift !

freeboub (changes requested):

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -525,41 +524,40 @@ private void reLayout(View view) {
...
+        Timeline timeline = player.getCurrentTimeline();
+
+        // Determine if the content is live
+        if (!timeline.isEmpty()) {
+            Timeline.Window window = new Timeline.Window();
+            timeline.getWindow(player.getCurrentMediaItemIndex(), window);
+            isLive = window.isLive();
         }
+
+        if(isLive){
```
freeboub: instead of getting isLive from stream, can you add a text property which define the label to display ?
--> then you will display live text, if the player isLive (like you do ) and the label is provided by the application.
It will add the new feature, avoid UI change for existing user and allow translation support !

seyedmostafahasani: I agree with you. We can set a new text property for this, with `Live` as the default value.
WDYT?

freeboub: No, if there is no value, you will keep current behavior!

android/src/main/java/com/brentvatne/common/api/ControlsConfig.kt
```diff
@@ -9,6 +9,7 @@ class ControlsConfig {
     var hideDuration: Boolean = false
     var hideNavigationBarOnFullScreenMode: Boolean = true
     var hideNotificationBarOnFullScreenMode: Boolean = true
+    var liveLabel: String? = ""
```
freeboub: It should be better with this I think:
```suggestion
    var liveLabel: String? = null
```

android/src/main/java/com/brentvatne/common/api/ControlsConfig.kt
```diff
@@ -21,6 +22,7 @@ class ControlsConfig {
                 config.hideDuration = ReactBridgeUtils.safeGetBool(src, "hideDuration", false)
                 config.hideNavigationBarOnFullScreenMode = ReactBridgeUtils.safeGetBool(src, "hideNavigationBarOnFullScreenMode", true)
                 config.hideNotificationBarOnFullScreenMode = ReactBridgeUtils.safeGetBool(src, "hideNotificationBarOnFullScreenMode", true)
+                config.liveLabel = ReactBridgeUtils.safeGetString(src, "liveLabel", "")
```
freeboub: Then
```suggestion
                config.liveLabel = ReactBridgeUtils.safeGetString(src, "liveLabel", null)
```

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -525,41 +525,42 @@ private void reLayout(View view) {
...
+        Timeline timeline = player.getCurrentTimeline();
+
+        // Determine if the content is live
+        if (!timeline.isEmpty()) {
+            Timeline.Window window = new Timeline.Window();
+            timeline.getWindow(player.getCurrentMediaItemIndex(), window);
+            isLive = window.isLive();
         }
+
+        if(isLive && !Objects.equals(controlsConfig.getLiveLabel(), "")){
```
freeboub: And then:
```suggestion
        if(isLive && controlsConfig.getLiveLabel() != null){
```

With this change it would allow client to put "" live label

freeboub:
One more remark and after I think we are good !

freeboub:
and you forgot to add the new field in the documentation !

seyedmostafahasani:
You're right.
I will add it in a while.

seyedmostafahasani:
@freeboub 
Thanks for reviewing. I applied your suggestions.

android/src/main/java/com/brentvatne/common/api/ControlsConfig.kt
```diff
@@ -21,6 +22,7 @@ class ControlsConfig {
                 config.hideDuration = ReactBridgeUtils.safeGetBool(src, "hideDuration", false)
                 config.hideNavigationBarOnFullScreenMode = ReactBridgeUtils.safeGetBool(src, "hideNavigationBarOnFullScreenMode", true)
                 config.hideNotificationBarOnFullScreenMode = ReactBridgeUtils.safeGetBool(src, "hideNotificationBarOnFullScreenMode", true)
+                config.liveLabel = ReactBridgeUtils.safeGetString(src, "liveLabel", null)
```
KrzysztofMoch: Can we rename `src` to something else ? like `controlsConfig` or sm else

seyedmostafahasani: Yeah, we can do it. Let me handle that.

seyedmostafahasani: @KrzysztofMoch 
Done.

KrzysztofMoch (approved):

freeboub:
@seyedmostafahasani there are some conflict due to another pr merged.
Reviewing again the changes, can you please refactor this kind of duplicated code in a dedicated function:

```java
            previousButton.setImageAlpha(0);
            previousButton.setClickable(false);
```

Thank you

seyedmostafahasani:
> @seyedmostafahasani there are some conflict due to another pr merged. Reviewing again the changes, can you please refactor this kind of duplicated code in a dedicated function:
> 
> ```java
>             previousButton.setImageAlpha(0);
>             previousButton.setClickable(false);
> ```
> 
> Thank you

@freeboub 
Anytime!
It's done.

freeboub (approved):

Issue #4226, opened 2024-10-09
[Feature]: Paused state from ref


maksymhcode-care:
### Description

Not crucial, but do you think getting play/pause state from ref can be added?

### Why it is needed ?

Having ability to get this state locally can help with optimization

seyedmostafahasani:
@maksymhcode-care 
You can use these methods: [pause](https://thewidlarzgroup.github.io/react-native-video/component/methods#pause) and [resume](https://thewidlarzgroup.github.io/react-native-video/component/methods#resume)

cordovalegacy:
Yes this would be a great thing to have

freeboub:
This should be implemented on js side. Register on the event in Video.tsx and save the state to a ref. The accessor just needs to return the value on the ref. (Can also be implemented on the app side, but it can but integrated in the package). Ping @seyedmostafahasani I know you start something similar.

paulrinaldi:
Yeah this could be implemented as @freeboub suggests.
The way the exoplayer lets us keep track of state is by `isPlayingChanged` events in any class that implements `Player.Listener` (in rnv, the class `ReactExoplayerView` does). 

I believe you @maksymhcode-care could implement in your video player containing component the following currently though:
```
const videoIsPlaying = useRef(false);

<Video
  onPlaybackStateStateChanged={(data) => videoIsPlaying.current = data.isPlaying}
  />
  ```

maksymhcode-care:
> Yeah this could be implemented as @freeboub suggests. The way the exoplayer lets us keep track of state is by `isPlayingChanged` events in any class that implements `Player.Listener` (in rnv, the class `ReactExoplayerView` does).
> 
> I believe you @maksymhcode-care could implement in your video player containing component the following currently though:
> 
> ```
> const videoIsPlaying = useRef(false);
> 
> <Video
>   onPlaybackStateStateChanged={(data) => videoIsPlaying.current = data.isPlaying}
>   />
> ```

Yeah, something like that would works, thanks

PR #4229, opened 2024-10-10
feat: add method for getting current playback status

seyedmostafahasani:
## Summary
Add method for getting current playback status.

### Motivation
Implement feature request #4226.

## Test plan
Run the sample app, add a button get play status with the use of function `getCurrentPlaybackStatus`.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -802,22 +802,32 @@ public class ReactExoplayerView extends FrameLayout implements
     public void getCurrentPlaybackStatus(Promise promise) {
         if (player != null) {
             int playbackState = player.getPlaybackState();
-            switch (playbackState){
-                case Player.STATE_IDLE -> promise.resolve("idle");
-                case Player.STATE_BUFFERING -> promise.resolve("buffering");
-                case Player.STATE_READY -> {
-                    if(isPaused){
+            switch (playbackState) {
+                case Player.STATE_IDLE:
+                    promise.resolve("idle");
+                    break;
+                case Player.STATE_BUFFERING:
+                    promise.resolve("buffering");
+                    break;
+                case Player.STATE_READY:
+                    if (isPaused) {
                         promise.resolve("paused");
+                    } else {
+                        promise.resolve("playing");
                     }
-                    promise.resolve("playing");
-                }
-                case Player.STATE_ENDED -> promise.resolve("ended");
-                default -> promise.resolve("unknown");
+                    break;
+                case Player.STATE_ENDED:
+                    promise.resolve("ended");
+                    break;
+                default:
+                    promise.resolve("unknown");
+                    break;
             }
         } else {
-            promise.reject("PLAYER_NOT_AVAILABLE", "Player is not initialized.");
+            promise.resolve("released");
         }
     }
+
 
     private void initializePlayerCore(ReactExoplayerView self) {
         ExoTrackSelection.Factory videoTrackSelectionFactory = new AdaptiveTrackSelection.Factory();
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -799,6 +799,26 @@ public class ReactExoplayerView extends FrameLayout implements
         }
     }
 
+    public void getCurrentPlaybackStatus(Promise promise) {
+        if (player != null) {
+            int playbackState = player.getPlaybackState();
+            switch (playbackState){
+                case Player.STATE_IDLE -> promise.resolve("idle");
+                case Player.STATE_BUFFERING -> promise.resolve("buffering");
+                case Player.STATE_READY -> {
+                    if(isPaused){
+                        promise.resolve("paused");
+                    }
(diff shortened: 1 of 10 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -799,6 +799,26 @@ public void getCurrentPosition(Promise promise) {
         }
     }
 
+    public void getCurrentPlaybackStatus(Promise promise) {
+        if (player != null) {
+            int playbackState = player.getPlaybackState();
+            switch (playbackState){
+                case Player.STATE_IDLE -> promise.resolve("idle");
```
freeboub: I think you should add breaks in the switch case ?!

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -799,6 +799,26 @@ public void getCurrentPosition(Promise promise) {
...
+                case Player.STATE_BUFFERING -> promise.resolve("buffering");
+                case Player.STATE_READY -> {
+                    if(isPaused){
+                        promise.resolve("paused");
+                    }
+                    promise.resolve("playing");
```
freeboub: in a else please

ios/Video/RCTVideo.swift
```diff
@@ -1685,6 +1685,44 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
...
+                        }
+                    }
+                } else {
+                    resolve("unknown")
+                }
+            @unknown default:
+                resolve("unknown")
+            }
+        } else {
+            reject("PLAYER_NOT_AVAILABLE", "Player is not initialized.", nil)
```
freeboub: resolve("released")

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -799,6 +799,26 @@ public void getCurrentPosition(Promise promise) {
...
+                }
+                case Player.STATE_ENDED -> promise.resolve("ended");
+                default -> promise.resolve("unknown");
+            }
+        } else {
+            promise.reject("PLAYER_NOT_AVAILABLE", "Player is not initialized.");
```
freeboub: resolve("released") ?

src/types/video.ts
```diff
@@ -40,6 +40,15 @@ export type ReactVideoSourceProperties = {
   textTracks?: TextTracks;
 };
 
+export type TPlaybackStatus = {
```
freeboub, line 43: PlaybackStatus would be enough ?

docs/pages/component/methods.mdx
```diff
@@ -127,6 +127,17 @@ On iOS, this displays the video in a fullscreen view controller with controls.
...
+
+This function retrieves the current playback status of the media and can return one of the following states:
+
+- **idle** (Android only): The media is not yet prepared, and the player must be prepared before it can play the media.
+- **paused**: The media is loaded but not actively playing.
+- **playing**: The media is currently playing.
+- **buffering**: The media is loading or waiting for data to continue playback.
+- **ended**: The media has finished playing.
+- **unknown**: The media has unknown status.
+
```
freeboub, line 151: you didn't document the exception, but As I said a released state would be better.
THhank you

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
seyedmostafahasani: It doesn't need but I added just to be on the safe side.

src/types/video.ts
seyedmostafahasani: I think it's better to add the prefix `T` before type names, meaning playbackStatus would be a type. Similarly, we can add an `I` before interface names
FX: type TSample = {} and interface ISample = {}
WDYT?

YangJonghun:
@seyedmostafahasani 
It's a great idea and something I've been wanting.
But in new arch, functions called via ref can't have a return value. Current implementation of `save` or `getCurrentPosition` is tricky and may be deprecated in the future via RN Core, so using callback props may be a better approach for now.
(This is why I made it so that library authors see deprecation warning when using `getReactTag`.)

seyedmostafahasani:
@YangJonghun 
Thanks for informing me. I think we can use this method until the RN core team decides on it. If they deprecate it, we can implement a new approach using callback props.
WDYT?

freeboub:
@seyedmostafahasani if it is deprecated, It is nopt a good idea to use it.
BTW a possible implementation would be to implement the event callback on native side.
Register an handler in Video.tsx and the save the last state in a Ref.
Then the getter will just return the ref value.

YangJonghun:
> Thanks for informing me. I think we can use this method until the RN core team decides on it. If they deprecate it, we can implement a new approach using callback props.
WDYT?

Same as above comment

src/types/video.ts
IslamRustamov: As Robert Martin wrote in his "Clean Code" - T, I, Enum, List (and so on) prefixes/suffixes before types, interfaces, enums (and so on) are meaningless.

We are not necessarily need to follow what he writes in his books, but it's better to follow guidelines that are accepted in the project you are working on (for example, as I can see in this project suffixes/prefixes are not used)

PR #4235, opened 2024-10-13
fix: remove warning and refactor & fix ad workflow

freeboub:
## Summary
Clean up ads workflow. and remove minor warnings

### Motivation
centralize ads management

## Test plan

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -856,9 +856,9 @@ public class ReactExoplayerView extends FrameLayout implements
     }
 
     private AdsMediaSource initializeAds(MediaSource videoSource, Source runningSource) {
-        AdsMediaSource mediaSourceWithAds = null;
-        AdsProps adProps = source.getAdsProps();
-        if (adProps != null) {
+        AdsProps adProps = runningSource.getAdsProps();
+        Uri uri = runningSource.getUri();
+        if (adProps != null && uri != null) {
             Uri adTagUrl = adProps.getAdTagUrl();
             if (adTagUrl != null) {
                 exoPlayerView.showAds();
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -876,6 +855,42 @@ private void initializePlayerCore(ReactExoplayerView self) {
         }
     }
 
+    private AdsMediaSource initializeAds(MediaSource videoSource, Source runningSource) {
+        AdsMediaSource mediaSourceWithAds = null;
```
seyedmostafahasani: Initializing this variable to null is redundant because you immediately assign it a value later on.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -876,6 +855,42 @@ private void initializePlayerCore(ReactExoplayerView self) {
...
+        AdsProps adProps = source.getAdsProps();
+        if (adProps != null) {
```
seyedmostafahasani: I think we can change it as shown below for simplicity.
```
AdsProps adProps = source.getAdsProps();
if (adProps == null || adProps.getAdTagUrl() == null) {
        exoPlayerView.hideAds();
        return null;
    }
     exoPlayerView.showAds();
     ...
```

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: yes you are right, I also fix a potential issue with the source vs running source.
This is fixed !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
freeboub: I undestand your point, but there are 3 differents cases where I want to call `exoPlayerView.showAds();` so I prefer current version

PR #4245, opened 2024-10-20
Adds paused state to ref

genesy:
Simple addition to the ref.

### Motivation
#4226


freeboub:
Why do you need that ? Pause function already exists https://thewidlarzgroup.github.io/react-native-video/component/methods#pause
Moreover adding a state here will cause unnecessary renders ...

PR #4272, opened 2024-11-03
Add support for HLS videos caching on android

IslamRustamov:
## Summary
This PR adds a conditional check whether to include HLS video caching or not for Android devices. We already have a **bufferConfig** prop that enables caching in most cases, but we don't utilise the **useCache** variable in case of HLS video streaming.

### Motivation
Every time the same HLS video is being played - it gets loaded again and again which is counter-efficient.

### Changes

- Added a conditional check with **useCache** to whether enable HLS video caching or not.

## Test plan
Open App Inspector in Android Studio (or proxyman or any other sniffing tool) and check whether HLS videos are getting downloaded with and without **bufferConfig** prop.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1169,7 +1169,8 @@ public class ReactExoplayerView extends FrameLayout implements
 
                 mediaSourceFactory = new HlsMediaSource.Factory(
                         dataSourceFactory
-                ).setAllowChunklessPreparation(source.getTextTracksAllowChunklessPreparation());                break;
+                ).setAllowChunklessPreparation(source.getTextTracksAllowChunklessPreparation());
+                break;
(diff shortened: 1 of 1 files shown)


freeboub (approved):

freeboub:
looks good to me

douzal34:
Hi @freeboub,

I have seen the new version **6.8.0** and the new HLS cache managing from this PR. However, i've seen on this [link](https://docs.thewidlarzgroup.com/react-native-video/other/caching) that caching is not available for .m3u8 yet.

Does HLS caching managing is possible on rnv component or not ? If yes, can you share an example of it like a basic rnv component with the option needed.

Thanks for your time and have a nice day !

IslamRustamov:
@douzal34 probably docs were not updated

douzal34:
Still no update on the doc, it's normal @freeboub ?

guidodipaolo:
> Still no update on the doc, it's normal @freeboub ?

So its available or not ?

IslamRustamov:
@guidodipaolo it is available

PR #4293, opened 2024-11-18
fix(ios): handle async player access in text track selection

tarikfp:
# Fix Potential Race Condition in Video Player Text Track Selection

## Description
Potentially fixed a crash that happens when text tracks get selected while the player is being destroyed. Super hard to reproduce but shows up in crash logs. Basically the player exists when we start the async operation but might be gone by the time we try to use it.

## Changes
- Added weak self in async block
- Added guard check for player instance
- Made sure we have strong reference to player during async operation

## Steps to Maybe Reproduce
Really hard to get this consistently, but here's what might trigger it:
1. Switch videos really fast
2. Change text tracks while video is loading
3. Background the app during text track changes

## Type of Change
- [ ] Breaking change (fix or feature that would break existing functionality)

## Screenshots
ios-rctvideo-crash

## Checklist
- [x] Code follows project style guidelines
- [x] Edge cases tested where possible
- [x] No new warnings introduced


KrzysztofMoch (approved):
LGTM, Thanks for PR!

PR #4302, opened 2024-11-23
Fix(android): restart issue react76

### LINKED ISSUE #2693: Move startup props into a single configuration

### LINKED ISSUE #4261: [BUG]: toggling muted prop causes full rerender on RN 0.76.0
### Version

6.7.0

Android TV

### System Version

12

Simulator

### Architecture

### What happened?

Toggling the muted prop causes the whole component to rerender and start the video again.

Downgrading RN back to 0.75.2 resolves the issue.

### Reproduction Link

_No response_

### Reproduction

Toggle the muted prop

freeboub:
## Summary
fix: https://github.com/TheWidlarzGroup/react-native-video/issues/4261
fix: https://github.com/TheWidlarzGroup/react-native-video/issues/2693

### Motivation
Fix issue on react native 0.76

### Changes
- move bufferConfig inside source 
- fix restart issue on react 0.76
- fix constness
- deprecate bufferConfig in root props
- update documentation
- upgrade expo exemple to expo 54 with new arch

## Test plan
Can be test with the expo sample

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -91,9 +91,10 @@ def configStringPath = ExoplayerDependencies
         .concat("buildFromSource:$media3_buildFromSource")
         .md5()
 
-if (isNewArchitectureEnabled()) {
-    apply plugin: "com.facebook.react"
-}
+// commented as new architecture not yet fully supported
+// if (isNewArchitectureEnabled()) {
+//     apply plugin: "com.facebook.react"
+// }
 
 android {
     if (supportsNamespace()) {
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,5 +1,5 @@
 RNVideo_kotlinVersion=1.8.0
-RNVideo_minSdkVersion=23
+RNVideo_minSdkVersion=24
 RNVideo_targetSdkVersion=34
 RNVideo_compileSdkVersion=34
 RNVideo_ndkversion=26.1.10909125
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,5 +1,5 @@
 RNVideo_kotlinVersion=1.8.0
-RNVideo_minSdkVersion=24
+RNVideo_minSdkVersion=23
 RNVideo_targetSdkVersion=34
 RNVideo_compileSdkVersion=34
 RNVideo_ndkversion=26.1.10909125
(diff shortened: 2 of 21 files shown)


KrzysztofMoch (approved):
Amazing job! Can you tell me in short what part fixed issue on 0.76? Thank You!

freeboub:
> Amazing job! Can you tell me in short what part fixed issue on 0.76? Thank You!

I don't know why but we receive multiple times setBufferConfig callback. As there was a "force restart playback" inside this function playback was restarting 🤷
Maybe an issue in react native but I did already plan to change that (it was possible that it causes some race conditions in initial start up).

freeboub:
@KrzysztofMoch I will release that tonight!

moskalakamil (approved):
Nice!

hongjae-eum:
@freeboub 

I wasn't using bufferConfig at all, and I didn't encounter this restart issue in version 6.8.0.

However, after the patch update to version 6.8.1, when I pause a video, the onLoad callback triggers, causing the video to return to the starting point.

Could there be any suspected changes in the patch that might be related to this?

freeboub:
@hongjae-eum no it should not ... Do you use the last react native version ? 0.75 with newArch enabled ?

freeboub:
@hongjae-eum maybe this line: https://github.com/TheWidlarzGroup/react-native-video/blob/dd78241b0d997e87a65f2f898c7059a72951e532/android/src/main/java/com/brentvatne/common/api/Source.kt#L104
Can you try to remove the condition locally ? I think it can be the issue. I missed to overload the equals in BufferConfig ..

hongjae-eum:
@freeboub 

Commenting out that part(line 104) resolves the restart issue! If that part is indeed the problem, it would be great if a patch could be released quickly.

freeboub:
Yep, I will do it tonight, ping @KrzysztofMoch

freeboub:
@hongjae-eum here is the fix: https://github.com/TheWidlarzGroup/react-native-video/pull/4305
It has already been released !
Thank you for the quick feedback !

PR #4320, opened 2024-12-04
fix: _paused is updated when video playback pause

giodevcoding:
## Summary
Keep paused state in sync with playing state.

### Motivation
I found that the video resumes from being paused automatically when foregrounding the app, which is an unexpected behavior. I wanted the playback state from the background, whether playing or paused, to persist from background to foreground to create a better experience for the users of my app.

Also a fix for #4321

### Changes
Update `_paused` on RCTVideo alongside `_isPlaying` when the video playback (or time control status) changes.

I found that this was behavior was due to `_paused` on RCTVideo not being updated when the video is paused with native controls, as it is only updated by the `setPause()` method. The consequences of this are that `setPause()` is called within `applyModifiers()` which is called within `applicationWillEnterForeground().` This call uses the current value of `_paused` (`setPause(_paused)`), which is false, and `setPause()` will play the video if the value given to it is false.

Updating `_paused` alongside `_isPlaying` keeps it update with the state of the player and fixes the issue.

## Test plan
1. Open Video
2. Pause Video
3. Background App
4. Foreground App
5. Video should remain paused (or whatever it's background state was when backgrounded)

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -1508,6 +1508,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
         guard _isPlaying != isPlaying else { return }
         _isPlaying = isPlaying
+        _paused = !isPlaying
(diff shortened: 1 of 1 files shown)


KrzysztofMoch (approved):
Good catch!

PR #4323, opened 2024-12-05
chore: fix type in generic folder

freeboub:
## Summary
minor refactor of view data type

### Motivation
do not use exoplayer outside of exoplayer folder

### Changes
move type to Any

## Test plan
Build


seyedmostafahasani:
@freeboub 
What is your opinion on using a generic class instead of `Any`?

KrzysztofMoch (approved):
This align android with iOS - also generic type have no sense here since we don't care about - ideally we should have interface of player but let's leave it for future

freeboub:
I confirm we cannot use an interface here. The goal is that plugins uses the 'native handle', in the future, it will not be possible to extract a common inteface between VLC and exoplayer.
This is just a minor clean !

PR #4333, opened 2024-12-12
Allow overriding style on web

M-i-k-e-l:
## Summary
Allow overriding style on web

### Motivation
When overriding the `style` with `{}` we can allow the (web) video to resize to its real size without further effort.

### Changes
Sending `style` will now affect video on web.

## Test plan
I assume that if someone is already sending `style` this can affect them, however the release of the web support was quite recent so I'm not sure how risky this is; not sure how to test it though...

### DIFF
--- a/src/types/video.ts
+++ b/src/types/video.ts
@@ -9,7 +9,7 @@ import type {
   ImageURISource,
   ImageStyle,
 } from 'react-native';
-import type {ReactNode} from 'react';
+import type {CSSProperties, ReactNode} from 'react';
 import type VideoResizeMode from './ResizeMode';
 import type FilterType from './FilterType';
 import type ViewType from './ViewType';
@@ -350,3 +350,9 @@ export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
   allowsExternalPlayback?: boolean; // iOS
   controlsStyles?: ControlsStyles; // Android
 }
+
+export interface WebStyleProp {
+  style?: CSSProperties;
+}
+
+export type WebReactVideoProps = Omit<ReactVideoProps, 'style'> & WebStyleProp;
--- a/src/Video.web.tsx
+++ b/src/Video.web.tsx
@@ -7,7 +7,7 @@ import React, {
   type RefObject,
   useState,
 } from 'react';
-import type {VideoRef, ReactVideoProps, VideoMetadata} from './types';
+import type {VideoRef, WebReactVideoProps, VideoMetadata} from './types';
 
 // stolen from https://stackoverflow.com/a/77278013/21726244
 const isDeepEqual = <T,>(a: T, b: T): boolean => {
@@ -25,10 +25,11 @@ const isDeepEqual = <T,>(a: T, b: T): boolean => {
   );
 };
 
-const Video = forwardRef<VideoRef, ReactVideoProps>(
+const Video = forwardRef<VideoRef, WebReactVideoProps>(
   (
     {
       source,
+      style = videoStyle,
       paused,
       muted,
       volume,
@@ -127,7 +128,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     const setFullScreen = useCallback(
       async (
         newVal: boolean,
-        orientation?: ReactVideoProps['fullscreenOrientation'],
+        orientation?: WebReactVideoProps['fullscreenOrientation'],
         autorotate?: boolean,
       ) => {
         orientation ??= fsPrefs.current.fullscreenOrientation;
@@ -359,7 +360,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           onVolumeChange?.({volume: nativeRef.current.volume});
         }}
         onEnded={onEnd}
-        style={videoStyle}
+        style={style}
       />
     );
   },


moskalakamil:
I'm just wondering- if you have a react-native app you have one <Video /> component for web and native. Does typescript work as well to change the type of props depends on Platform.select()? And even if, what is the default type for this prop?

M-i-k-e-l:
> I'm just wondering- if you have a react-native app you have one

If you're referring to the testing, then I've tested it; I assumed the the `Test plan` is for unit test of some sort

moskalakamil:
Sorry, I'm not sure why my comment was cut off. What I meant was: if you have a Video component in your app that works for both web and native, do the types work correctly? Do they change, for example, when you use Platform.select? Also, what is the default type in that case?

M-i-k-e-l:
> Sorry, I'm not sure why my comment was cut off. What I meant was: if you have a Video component in your app that works for both web and native, do the types work correctly? Do they change, for example, when you use Platform.select? Also, what is the default type in that case?

If this is about the typing change, then I did not test that, I needed to send an empty style `{}`...
I made the change to the type because of the error in this project, not because of a usage issue - perhaps the correct thing was to ignore it.

src/types/video.ts
```diff
@@ -350,3 +350,9 @@ export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
   allowsExternalPlayback?: boolean; // iOS
   controlsStyles?: ControlsStyles; // Android
 }
+
+export interface WebStyleProp {
+  style?: CSSProperties;
+}
+
+export type WebReactVideoProps = Omit<ReactVideoProps, 'style'> & WebStyleProp;
```
freeboub, line 358: why don't you patch directly ReactVideoProps instead of creating WebStyleProp in this common file ?

freeboub: something like:

```JS
type CustomViewProps = Omit<ViewProps, 'style'> & {
  style?: ViewProps['style'] | CSSProperties;
};

export interface ReactVideoProps extends ReactVideoEvents, CustomViewProps {
```

freeboub:
+ ping @zoriya

zoriya:
I purposely omitted the style props because this is not compatible with react native styles. this means you'll need a different style for web & native. Since types are different, `Platform.select` is bothersome to use & styling libs would just error.
Existing projects using the style property for native would probably have a type error due to the new style typing.

Maybe we could add a `webStyle` instead to prevent those issues.

M-i-k-e-l:
Another option is to add a prop that will have `hug` (`style={{}}`) and `fill` (`style={videoStyle}`) as values.

freeboub:
just a ping on this PR, I think we definitively need to integrate such change !

M-i-k-e-l:
Hi @freeboub,
Not sure if the ping is for me :)
I would love to get the prop name selected before changing the PR.
Technically `resizeMode` can be used since it's not used in web ATM (so maybe `contain` (for `hug` - `style={{}}`) and `stretch` (for `fill` - `style={videoStyle}`).
If you don't like this option, then maybe you'll like one of these `layoutMode` or `wrapStyle`?

freeboub:
@M-i-k-e-l Yes it was for you :)

I think you just need to check the patch in Video.ts 
please change:

```JS
export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
```

to export
```JS
interface ReactVideoProps extends ReactVideoEvents, Omit<ViewProps, 'style'> {
   style?: CSSProperties | Pick<ViewProps, 'style'>
```

I am not fan of having some Web dedicated interface. I think we should keep the api common (even if here we don't really have the choice)

Would that work for you ?

zoriya:
Wouldn't it make more sense to use https://necolas.github.io/react-native-web/docs/unstable-apis/#use-with-existing-react-dom-components to use the style prop from react-native with the web element?

freeboub:
@zoriya I am not confident in using unstable apis ... I am not a web dev, what do we gain by using this api ?

If it is only to support in 1 line: accessibilityLabel, accessible, style, I would prefer to handle it prop per prop, it gives a thiner control on the api and allow easier integration when we will enable them on   android/ios

zoriya:
This unstable api allows using react native's style prop on the web. Without using it, a lot of style are different. This also means that all native styling libraries (rnw's default, native-wind, tamagui, unistyle...) won't work on the web (and give type errors on native).

This api is marked as unstable but it hasn't changed in a lot of years, I think it's safe to use there honestly.

freeboub:
OK, @zoriya it looks good to me, but not sure on how to integrate it :D  @M-i-k-e-l

zoriya:
Should be as simple as adding

```tsx
import { unstable_createElement } from 'react-native-web';
const Video = (props) => unstable_createElement('video', props);
```

at the top of the file and changing the component from `<video>` to `<Video>`. (might need to type the props in the declaration above, idk)

M-i-k-e-l:
I think that the two major use-cases for style are `hug` and `fit` that I've mentioned [here](https://github.com/TheWidlarzGroup/react-native-video/pull/4333#issuecomment-2614249961), it would be nice to support a general style but I'm not familiar with `unstable_createElement` either and TBH I'd have the same concern about it.

zoriya:
What are your concerns? `unstable_createElement` would allow setting the `resizeMode` like you desire.

M-i-k-e-l:
Hi,

Apologies for not replying sooner.

Generally, I would feel uncomfortable adding code that has `unstable_` in it and since I do not know the API that was suggested and have no time to research it ATM I have to decline adding it.

Please feel free to close the PR and\or make the changes you've suggested.

jonathanj:
Is there any reason not to keep this change absolutely minimal and do `style={[videoStyle, style]}`, then at least the user can override the default style? At the moment the `style` prop is thrown away altogether on web, which is not the case for the native component.

The type changes seem academic because TypeScript will ignore `Video.web.tsx` when you `import Video from 'react-native-video`, since the `.web` mechanism is part of the RN babel tooling, not part of TypeScript.

zoriya:
passing native styles to a web component will either not render appropriately. we could support web only styles with your proposition but every style could be handled using the `unstable_createElement` api.

jonathanj:
Okay, I tried patching my local `node_modules` to use `unstable_createElement` and then also do `style={[videoStyle, style]}`, and as far as I can tell everything still seems to work and additionally the `style` prop is respected.

I think supporting `resizeMode` is out of scope for a PR that wants to allow overriding styles, I would think that `resizeMode` would just select different/better values for `videoStyle` but `style` should still be able to override those.

jonathanj:
As per the original author's [comment](https://github.com/TheWidlarzGroup/react-native-video/pull/4333#issuecomment-2677843129), I have opened a new PR https://github.com/TheWidlarzGroup/react-native-video/pull/4528 which uses `unstable_createElement` and preserves the original `videoStyle` so as not to introduce a breaking change.

KrzysztofMoch:
Implemented in https://github.com/TheWidlarzGroup/react-native-video/pull/4528

Issue #2884, opened 2022-10-05
Refresh expired bearer token in drm access


danieldulcettiBN:
### Current behavior
I need to refresh my url (streaming video) periodically since it has a token that expires after some time.
I have already managed this flow (I ask the new url before it expires) but screen appears blank for about a second after url refreshing, probably caused by source reloading.
I also tried putting the token in the source header, but nothing.

### Reproduction steps
1. Change source (same url with different token) while it plays

### Expected behavior
The source must be changed without the user noticing the transition.

### Platform
* Android TV
* tvOS

### Version
5.2.0

Thanks to everyone!

freeboub:
It is expected that changing the url change playback.
For sure, you should use header to achieve that, but I don't know how to do it or if it is really possible...
Notice that as nobody can reproduce the issue as your server api is not open, you will have to investigate it

falk-stefan:
@freeboub How can we refresh an access token if it expires mid-stream?

Currently I am providing the token by setting `headers` via `source` like this: 

```javascript
source={{
  uri: trackUri,
  headers: {
    Authorization: `Bearer ${accessToken}`,
  },
}}
```

But how can we refresh it on expiry? :thinking:

freeboub:
@falk-stefan unfortunately I don't have any solution for now ... I will reopen the ticket to track it, but as I don't have any way to reproduce the issue, I cannot investigate... 
In terms of api I can do some suggestions if needed.

falk-stefan:
@freeboub I have a different requirement though- I do not need to refresh the URL - but the `Authorization` header once a request reteurns HTTP 401 Token Expired. 

It's not the exact same as what @danieldulcettiBN tries to do.

freeboub:
Yes, I agree but it is more or less the same use case. 
The change would be to add a callback onAuthError. When this callback is called application would need to call another function api to refresh bearer token (or the url)

falk-stefan:
Hm, a more generic way would be to just return the full response on error and let the user do whatever is right. Perhaps the onAuthError could be a convenience thing but generally speaking it would probably be nice to have a general way to either intercept requests and/or responses and open a path to retry a request if needed.

freeboub:
You mean the onError callback should be enough. I think it makes sense. Another ticket exists about error code clean up, they should be linked then.
Something I am not sure about and should be experienced, about the timing, would JS be fast enough to renew a token fastly enough and provide to react native video to get a seemless renew ...

[...]

PR #4352, opened 2024-12-28
Add HLS playback support for web browsers

bhyoo99:
## Summary
Add HLS playback support for web browsers

### Motivation
Some web browsers do not support native HLS playback, requiring hls.js implementation for cross-browser compatibility

### Changes
- Add `hls.js` package
- Implement HLS stream validation logic
- Add HLS player initialization for browser compatibility

## Test plan
1. HLS stream playback testing:
   - Verify HLS stream playback in Chrome and Firefox
   - Confirm proper functionality in browsers without native HLS support
   - Check for memory leaks during stream load/unload

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,72 +1,72 @@
 {
-  "name": "react-native-video",
-  "version": "6.8.2",
-  "description": "A <Video /> element for react-native",
-  "main": "lib/index",
-  "source": "src/index.ts",
-  "react-native": "src/index",
-  "license": "MIT",
-  "author": "Community Contributors",
-  "homepage": "https://docs.thewidlarzgroup.com/react-native-video/",
-  "repository": {
-    "type": "git",
-    "url": "git@github.com:TheWidlarzGroup/react-native-video.git"
-  },
-  "resolutions": {
-    "@types/react": "~18.0.0"
-  },
-  "devDependencies": {
-    "@expo/config-plugins": "^8.0.5",
-    "@jamesacarr/eslint-formatter-github-actions": "^0.2.0",
-    "@react-native/eslint-config": "^0.72.2",
-    "@release-it/conventional-changelog": "^7.0.2",
-    "@types/jest": "^28.1.2",
-    "@types/react": "~18.0.0",
-    "@types/react-native": "0.72.3",
-    "@typescript-eslint/eslint-plugin": "^6.7.4",
-    "eslint": "^8.19.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
         "@typescript-eslint/eslint-plugin": "^6.7.4",
         "eslint": "^8.19.0",
         "eslint-plugin-jest": "^27.4.2",
+        "hls.js": "^1.5.18",
```
KrzysztofMoch, line 29: I think we should add this also as __optional__ peer dependency

KrzysztofMoch (changes requested):
Hey, Thank You for the PR!

freeboub:
+ ping @zoriya
I am not sure of the implementation, what will happen when switching from hls to dash and the opposite ...

package.json
bhyoo99: @KrzysztofMoch You mean `peerDependencies`?

bhyoo99: @KrzysztofMoch like this?

```
"peerDependencies": {
        "react": "*",
        "react-native": "*",
        "hls.js": "^1.5.18"
 },
```

freeboub: @bhyoo99 The best would be to put it in optionnalDependency: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies
It will allow to not embed the package is not needed.
It would be interesting to put the information in the documentation also.

zoriya (changes requested):
This implementation lacks some options, namely:
- source containing headers (using hls's `xhrSetup`)
- video/audio/subtitles tracks reporting (with the `onInfo` prop)
- selection of video/audio/subtitle (`selectedVideoTrack`, `selectedAudioTrack` & `selectedTextTrack` props)
- error handling
- `startPosition` support

For headers, it could be nice to consider #2884.
I also know that error handling is planned to be reworked, idk if there's somewhere we can refer to on how to implement this for hls?


Regarding HLS.js as a whole, it was my first guess since I never needed more than hls for kyoo, but for rnv we might want to consider something a bit more feature-full. I considered both [shaka](https://github.com/shaka-project/shaka-player) and [video.js](https://github.com/videojs/video.js) when I first did #3958 (and I think I preferred shaka, but not sure anymore tbh.) Does anyone have an opinion on this? If not, we might want to research a bit and list drawbacks/advantages of all solutions before actually implementing anything.

PR #4355, opened 2024-12-31
fix: avoid memory leak on iOS

vladvlasov256:
## Summary

The PR fixes memory leaks on iOS. 

Screenshot 2024-12-31 at 10 19 08

The `RCTVideo` and `ReactNativeVideoManager` create a reference cycle because `ReactNativeVideoManager` retains the views but never release them.

### Motivation

Reducing the memory usage in the apps.

### Changes

Removes redundant `instanceList` in `ReactNativeVideoManager` but keep the counter logic. React Native retain the view references so no need to have an extra list.

## Test plan

1. In the basic app, push and pop the video screen 8 times. 
2. Check `RCTVideo` instances in the memory graph.

**Expected result**: only a single `RCTVideo` instance is presented.

Screenshot 2024-12-31 at 10 29 33

<br><br>

> ⚠️ *In the sample app, the `CLANG_ADDRESS_SANITIZER` is enabled in debug. So Xcode doesn't show a proper object list. I had to remove `Yes` value for test purposes.* Screenshot 2024-12-31 at 10 41 01

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -256,7 +256,6 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
-        ReactNativeVideoManager.shared.registerView(newInstance: self)
         #if USE_GOOGLE_IMA
             _imaAdsManager = RCTIMAAdsManager(video: self, pipEnabled: isPipEnabled)
         #endif
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -256,6 +256,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
+        ReactNativeVideoManager.shared.registerView(newInstance: self)
         #if USE_GOOGLE_IMA
             _imaAdsManager = RCTIMAAdsManager(video: self, pipEnabled: isPipEnabled)
         #endif
--- a/ios/Video/ReactNativeVideoManager.swift
+++ b/ios/Video/ReactNativeVideoManager.swift
@@ -17,9 +17,9 @@ public class ReactNativeVideoManager: RNVPlugin {
     private var pluginList: [RNVPlugin] = Array()
 
     /**
-      * register a new view
+     * register a new view
      */
-    func registerView(newInstance: RCTVideo) {
+    func registerView(newInstance _: RCTVideo) {
         if instanceCount > expectedMaxVideoCount {
             DebugLog("multiple Video displayed ?")
         }
@@ -29,7 +29,7 @@ public class ReactNativeVideoManager: RNVPlugin {
     /**
      * unregister existing view
      */
-    func unregisterView(newInstance: RCTVideo) {
+    func unregisterView(newInstance _: RCTVideo) {
         instanceCount -= 1
     }
 
--- a/ios/Video/ReactNativeVideoManager.swift
+++ b/ios/Video/ReactNativeVideoManager.swift
@@ -6,33 +6,31 @@
 import Foundation
 
 public class ReactNativeVideoManager: RNVPlugin {
-    private let expectedMaxVideoCount = 10
+    private let expectedMaxVideoCount = 2
 
     // create a private initializer
     private init() {}
 
     public static let shared: ReactNativeVideoManager = .init()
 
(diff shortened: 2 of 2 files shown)


vladvlasov256:
There might be a bug in the instance counter logic on both platforms. Please take a look at the Android code:

```
 fun registerView(newInstance: ReactExoplayerViewManager) {
    if (instanceList.size > 2) {
        DebugLog.d(TAG, "multiple Video displayed ?")
    }
    instanceList.add(newInstance)
}
```

As I understand, a user gets a debug output only after adding the 4th video to the screen:

Added the 1st video. `instanceList.size = 0`, no warning.
Added the 2nd video. `instanceList.size = 1`, no warning.
Added the 3rd video. `instanceList.size = 2`, no warning.
Added the 4th video. `instanceList.size = 3`, the warning is displayed.

Should the condition be `instanceList.size > 0` instead? So then the user gets the warning after adding the 2nd video.

@freeboub what do you think?

ios/Video/RCTVideo.swift
```diff
@@ -256,6 +256,7 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
+        ReactNativeVideoManager.shared.registerView(newInstance: self)
```
freeboub: Why is there 2 init functions 🤔 
BTW looks good to me

freeboub (approved):

ios/Video/RCTVideo.swift
KrzysztofMoch: That one with coder is for compatibility with older versions of swift and storyboard - this init in general should be never called

freeboub:
> There might be a bug in the instance counter logic on both platforms. Please take a look at the Android code:
> 
> ```
>  fun registerView(newInstance: ReactExoplayerViewManager) {
>     if (instanceList.size > 2) {
>         DebugLog.d(TAG, "multiple Video displayed ?")
>     }
>     instanceList.add(newInstance)
> }
> ```
> 
> As I understand, a user gets a debug output only after adding the 4th video to the screen:
> 
> Added the 1st video. `instanceList.size = 0`, no warning. Added the 2nd video. `instanceList.size = 1`, no warning. Added the 3rd video. `instanceList.size = 2`, no warning. Added the 4th video. `instanceList.size = 3`, the warning is displayed.
> 
> Should the condition be `instanceList.size > 0` instead? So then the user gets the warning after adding the 2nd video.
> 
> @freeboub what do you think?


Yes I may do better. I put this, as is, because, from my experience, 2 parallel video decoding on a device is working pretty fine in most of case... but It is not absolutely true. In another fork, I implemented a limitation to ensure we don't use more than 'X' Video in the same time by droping previous instance (for safety).

ios/Video/RCTVideo.swift
vladvlasov256: Removed this call. I added it just for init/deinit symmetry.

vladvlasov256:
> Yes I may do better. I put this, as is, because, from my experience, 2 parallel video decoding on a device is working pretty fine in most of case... but It is not absolutely true

@freeboub, in this PR, the maximum value is changed from 10 to 2 to align it with Android. Does this make sense, or should I revert it?

freeboub:
> > Yes I may do better. I put this, as is, because, from my experience, 2 parallel video decoding on a device is working pretty fine in most of case... but It is not absolutely true
> 
> @freeboub, in this PR, the maximum value is changed from 10 to 2 to align it with Android. Does this make sense, or should I revert it?

No that's good, ios is really more robust at this point (but it's not a reason) :)

ios/Video/ReactNativeVideoManager.swift
```diff
@@ -6,33 +6,31 @@
...
+    private var pluginList: [RNVPlugin] = Array()
 
     /**
-      * register a new ReactExoplayerViewManager in the managed list
+     * register a new view
      */
-    func registerView(newInstance: RCTVideo) {
-        if instanceList.count > expectedMaxVideoCount {
+    func registerView(newInstance _: RCTVideo) {
+        if instanceCount > expectedMaxVideoCount {
```
KrzysztofMoch, line 23: Is this breaking change? If it is let's add "!" to commit -> `fix!: avoid memory leak on iOS`

KrzysztofMoch (approved):
It's fine for me

ios/Video/ReactNativeVideoManager.swift
vladvlasov256: No, it doesn't break the existing API. But the argument is unused so `_` makes the linter happy.

freeboub:
@KrzysztofMoch I confirm there is not breaking change in this PR, let's merge it !

PR #4366, opened 2025-01-09
feat: enhance react-native-video plugins [Plugins API Breaking]

KrzysztofMoch:
## Summary
1. I have changed DRM logic to allow plugins to implement custom drm provider. It should not have impact for current users
2. Add missing `unregisterPlugin` function on iOS (I tried, but it is impossible to add it without breaking change. Now plugins are experimental so it's good moment to do it)
3. Add type-safe wrappers for ExoPlayer/AVPlayer `onInstanceCreated`/`onInstanceRemoved` methods

### Motivation
Enhance plugins by allowing plugins to implement custom drm provider. Eg. implement DRM SDK form 3th party DRM providers

## Test plan
I have ran our DRM example on both iOS & Android physical devices to confirm that "default" DRM works


examples/bare/package.json
```diff
@@ -18,32 +18,32 @@
     "windows": "react-native run-windows --sln windows/BareExample.sln"
   },
   "dependencies": {
-    "@callstack/react-native-visionos": "^0.73.0",
-    "@react-native-picker/picker": "2.8.1",
-    "react": "18.2.0",
-    "react-native": "0.73.2",
```
freeboub: here you upgrade react native, split your pr please !

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -997,6 +1004,27 @@ private void initializePlayerSource(Source runningSource) {
...
 
+    private DrmSessionManager initializePlayerDrm() {
+        DrmSessionManager drmSessionManager = null;
+        DRMProps drmProps = source.getDrmProps();
+        // need to realign UUID in DRM Props from source
+        if (drmProps != null && drmProps.getDrmType() != null) {
+            UUID uuid = Util.getDrmUuid(drmProps.getDrmType());
+            if (uuid != null) {
+                try {
+                    DebugLog.w(TAG, "drm buildDrmSessionManager");
```
freeboub: Not your code, but I think we should remove this warning (which is not a warning in fact)

```suggestion
                    DebugLog.d(TAG, "drm buildDrmSessionManager");
```

freeboub (changes requested):
Code looks good, but I am not fan of this solution (I understand the need by the way).
My issue here is that you introduced a dependency between RNVplugin and Exoplayer or AVPlayer.
The day we will support other players, The RNVPlugin will not be compatible with new player implementation.

We have 2 choices to handle that, either
- remove dependency from RNVPlugin and make a generic api (which looks really tricky on my side).
- move the DRM management to RNVExoplayerDrmPlugin / RNVAVPlayerDrmPlugin which clearly highlight the player dependency

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
KrzysztofMoch: Done ✅

examples/bare/package.json
KrzysztofMoch: Moved out ✅

KrzysztofMoch:
@freeboub I have moved this custom DRM logic into `RNVAVPlayerPlugin`/`RNVExoplayerPlugin` - it is fine now ?

android/src/main/java/com/brentvatne/exoplayer/DRMManager.kt
```diff
@@ -0,0 +1,57 @@
...
+            // Set DRM headers
+            val keyRequestPropertiesArray = drmProps.drmLicenseHeader
+            for (i in keyRequestPropertiesArray.indices step 2) {
+                drmCallback.setKeyRequestProperty(keyRequestPropertiesArray[i], keyRequestPropertiesArray[i + 1])
+            }
+
+            val mediaDrm = FrameworkMediaDrm.newInstance(uuid)
+            if (hasDrmFailed) {
+                // When DRM fails using L1 we want to switch to L3
+                mediaDrm.setPropertyString("securityLevel", "L3")
```
freeboub, line 39: I know it is not you but this is a bad behavior in terms of security ...
It would be to add a flag in API to configure this behavior. (not blocking, just to inform you :) )

examples/react-native-video-plugin-sample/android/src/main/java/com/videopluginsample/VideoPluginSampleModule.kt
```diff
@@ -35,16 +35,12 @@ class VideoPluginSampleModule(reactContext: ReactApplicationContext) :
   }
 
 
-  override fun onInstanceCreated(id: String, player: Any) {
-    if (player is ExoPlayer) {
```
freeboub, line 39: would be better to keep this test: if (player is ExoPlayer) {
in case of another player integration, we will not have to readd it :)

examples/react-native-video-plugin-sample/android/src/main/java/com/videopluginsample/VideoPluginSampleModule.kt
```diff
@@ -35,16 +35,12 @@ class VideoPluginSampleModule(reactContext: ReactApplicationContext) :
...
-      player.addAnalyticsListener(debugEventLogger)
-      player.addListener(this)
-    }
+  override fun onInstanceCreated(id: String, player: ExoPlayer) {
+    player.addAnalyticsListener(debugEventLogger)
+    player.addListener(this)
   }
 
-  override fun onInstanceRemoved(id: String, player: Any) {
-    if (player is ExoPlayer) {
```
freeboub, line 46: same here

freeboub:
some minor feedback,  but it looks good to me !
There are some remaining exoplayer reference in Manager, but as it is very located, it is OK for me !
You put [breaking], if it breaks only plugin api, no need to bump major version number.
Sorry for loooonnnngggg delay to review...

freeboub (approved):

android/src/main/java/com/brentvatne/exoplayer/DRMManager.kt
KrzysztofMoch: Thanks for letting me know! tho it's not my code, I copied this from our impl of DRM - Will leave comment in code to fix this

examples/react-native-video-plugin-sample/android/src/main/java/com/videopluginsample/VideoPluginSampleModule.kt
KrzysztofMoch: Here I have deleted this, because we are extending `RNVExoplayerPlugin`. `RNVExoplayerPlugin` implements `fun onInstanceCreated(id: String, player: ExoPlayer)` so there is no need to check for ExoPlayer type. If we in future and more players I will update this, but I think this is better for now

examples/react-native-video-plugin-sample/android/src/main/java/com/videopluginsample/VideoPluginSampleModule.kt
KrzysztofMoch: https://github.com/TheWidlarzGroup/react-native-video/pull/4366#discussion_r1991382053

PR #4381, opened 2025-01-20
fix conversion of string to Stretch enum

darthLeviN:
The current code probably never worked for anyone. I don't know why no one has realized this yet.
You will need to convert string to enum values.


freeboub (approved):

freeboub:
low usage on windows, thank you for the fix !

PR #4390, opened 2025-01-25
Sideload audio

pigeonmal:
## Summary
A source.audioTracks for sideload audio tracks.
I updated the patch for work on 0.77 react-native.

### Motivation
https://github.com/TheWidlarzGroup/react-native-video/issues/4374
https://github.com/TheWidlarzGroup/react-native-video/issues/3265
### Changes
Get sideload audio and merge all in one source for exoplayer, check commits.
## Test plan
It's not perfect for now, for example in my code I didn't setup language of audio in the Reactexoplayerview.java. 
In 6.8.2 I had a poster bug, when audio tracks and poster are set : Audio work, but the poster stay over the video.
I don't test in 6.9 if this is fixed, I will test.

### DIFF
--- a/android/src/main/java/com/brentvatne/common/api/SideLoadedAudioTrack.kt
+++ b/android/src/main/java/com/brentvatne/common/api/SideLoadedAudioTrack.kt
@@ -1,0 +1,34 @@
+package com.brentvatne.common.api
+
+import android.net.Uri
+import com.brentvatne.common.toolbox.ReactBridgeUtils
+import com.facebook.react.bridge.ReadableMap
+
+/**
+* Class representing a sideLoaded audio track from application
+*/
+class SideLoadedAudioTrack {
+   var url: Uri = Uri.EMPTY
+   var title: String? = null
+   var language: String? = null
+   var sampleMimeType: String? = null
+
+   companion object {
+       val SIDELOAD_AUDIO_TRACK_URL = "url"
+       val SIDELOAD_AUDIO_TRACK_TITLE = "title"
+       val SIDELOAD_AUDIO_TRACK_LANGUAGE = "language"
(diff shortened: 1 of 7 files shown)


freeboub:
Can you please add a section in documentation please ? I would like to test your change in source creation (side loaded subtitles) but it looks great 👍 thank you !

pigeonmal:
No problem, but I have 3 problems :
1) The name of the prop is source.audioTracks not just audioTrack. Except that for now in the documentation, there is no grouping of source props. So this can lead to confusion?

2) In sampleMimeType, how to add all available mime types check: https://github.com/androidx/media/blob/release/libraries/common/src/main/java/androidx/media3/common/MimeTypes.java

3) In my code I put that title, language, mimetype are not mandatory, because I have implemented only the mimetype for now. Should they be mandatory?

Sorry for my bad English, I am French

freeboub:
1/ there is a source prop here : https://github.com/TheWidlarzGroup/react-native-video/blob/master/docs/pages/component/props.mdx#source
But I agree this part of doc shall be clean up! it is already confusing :)
Bad doc is bette than no doc !

2/ The easiest is to redirect to exoplayer doc, no issue with that !

3/ you mean in that place:
```JS
            MediaItem audioItem = new MediaItem.Builder()
                            .setUri(track.getUrl())
                            .setMimeType(track.getSampleMimeType())
                            .build();
```

you can add setTitle and/or setLanguage of something like this ?
If that is the case, just add it and it should work !

4/ pas de soucis avec ton anglais, je cause franglais aussi :D

KrzysztofMoch:
I have fixed kotlin lint action on master, can you rebase/pull from it? Thanks

pigeonmal:
Sorry I don't really have time anymore, I have medical studies on the side, I wanted to rename the branch it closed the pull requests

PR #4391, opened 2025-01-25
fix(sample): remove duplicate code in sample

freeboub:
## Summary
remove duplicated code in sample

### Motivation
remove unnecessary code to avoid any mismatch

### Changes
move common code in exemple/common

## Test plan
tested: bare android & ios + expo bare & ios


freeboub:
next step will be to fix broken seek in the sample

PR #4398, opened 2025-01-29
Audio decoder issue fix for android.

rahulnainwal107:
## Summary
I was facing an issue related to the audio decoder when the lang attribute was missing in manifest file for AdaptationSet tag and having multiple AdaptationSet, where 0th position audio is not supported by the device but 1st position audio is supported.

### Changes

1- Default groupIndex updated.

`int groupIndex = C.INDEX_UNSET; // default if no match`

2- Added a helper function for checking supported audio codes.

```
public boolean isCodecSupported(String format) {
        List<String> codecList = new ArrayList<>();
        int codecCount = MediaCodecList.getCodecCount();
        for (int i = 0; i < codecCount; i++) {
            android.media.MediaCodecInfo codecInfo = MediaCodecList.getCodecInfoAt(i);
            for (String type : codecInfo.getSupportedTypes()) {
                codecList.add(type);
            }
        }
        if (codecList.size() > 0) {
            return codecList.contains(format);
        }
        return false;
    }
````

3- If audio is supported only then updating groupIndex.

```
private int getGroupIndexForDefaultLocale(TrackGroupArray groups) {
        if (groups.length == 0){
            return C.INDEX_UNSET;
        }

        int groupIndex = C.INDEX_UNSET; // default if no match
        String locale2 = Locale.getDefault().getLanguage(); // 2 letter code
        String locale3 = Locale.getDefault().getISO3Language(); // 3 letter code
        for (int i = 0; i < groups.length; ++i) {
            Format format = groups.get(i).getFormat(0);
            String language = format.language;
            Boolean isSupported = isCodecSupported(format.sampleMimeType);
            if (isSupported) {
                if (language != null && (language.equals(locale2) || language.equals(locale3))) {
                    groupIndex = i;
                    break;
                } else {
                    groupIndex = i;
                }
            }
        }
        return groupIndex;
    }
```

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -2202,19 +2202,21 @@ public class ReactExoplayerView extends FrameLayout implements
         }
 
         int groupIndex = C.INDEX_UNSET; // default if no match
-        String locale2 = Locale.getDefault().getLanguage(); // 2 letter code
-        String locale3 = Locale.getDefault().getISO3Language(); // 3 letter code
+        Locale defaultLocale = Locale.getDefault();
+        String locale2 = defaultLocale.getLanguage();  // 2-letter code
+        String locale3 = defaultLocale.getISO3Language(); // 3-letter code
         for (int i = 0; i < groups.length; ++i) {
             Format format = groups.get(i).getFormat(0);
+
+            // Skip if codec is not supported
+            if (!isCodecSupported(format.sampleMimeType)) {
+                continue;
+            }
+
             String language = format.language;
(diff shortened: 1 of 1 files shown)


freeboub:
First thank you for the PR!
I don't think this will be OK, the function getGroupIndexForDefaultLocale is used for both audio and text tracks. Checking codec for text track doesn't make sense ...
But I am not sure on how to handle it (maybe just add track type to getGroupIndexForDefaultLocale and check it only for audio)

rahulnainwal107:
@freeboub Thanks for reviewing it. 
I will update it to check the codec only for audio.

freeboub:
> @freeboub Thanks for reviewing it. I will update it to check the codec only for audio.

No problem, I think you should also rename the function to isCodecAudioSupported then !

PR #4409, opened 2025-02-09
fix(ios): the video has no audio by default

### LINKED ISSUE #4400: [BUG]: There is no sound
### Version

6.10.0

iOS

### System Version

ios 18.2, 18.0 simulator and real device iPhone 12 Max Pro

Real device, Simulator

### Architecture

### What happened?

Video is playing nicely in ios environment but sound is not coming for some videos. I have tested it is working fine for 6.8.0 but starting from 6.9.0 it is not working anymore. It is working fine on both arch for 6.8.0. Here is a sample video URL for testing

### Reproduction Link
use this video to reproduce
https://geniusreachstorage.blob.core.windows.net/companyfiles/discover.mp4

### Reproduction

use the reproducing link and test it on 6.8.0 and later versions. Later versions won't have sound for this video

mlazari:
This reverts the change in https://github.com/TheWidlarzGroup/react-native-video/pull/4319, I don't think that change is correct.

According to [docs](https://docs.thewidlarzgroup.com/react-native-video/component/props#selectedtexttrack) the default value for `selectedTextTrack.type` is "system" which means the video should respect user's preference in Settings > Accesibility > Subtitles & Captioning > Closed Captions + SDH. And this is exactly what happens by default without the changes in that PR. I think the author of that PR had this setting on, that's why it was showing the subtitles without setting something for `selectedTextTrack`, which IMO is the expected result. In order to explicitly turn off subtitles you have to set `selectedTextTrack={{ type: 'disabled' }}`. I tested it with this video and the settings option on and it works (disables the subtitles) -> https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_ts/master.m3u8

IMG_0443

Another reason why I think it should be reverted is because it also affects `selectedAudioTrack`. By default it should also have 'system' type according to [docs](https://docs.thewidlarzgroup.com/react-native-video/component/props#selectedaudiotrack), which means "Play the audio track that matches the system language. If none match, play the first track.", but the change in that PR turns the audio off by default unless you explicitly turn it on with `selectedAudioTrack={{ type: 'system' }}`.

Fixes #4400

### DIFF
--- a/ios/Video/Features/RCTPlayerOperations.swift
+++ b/ios/Video/Features/RCTPlayerOperations.swift
@@ -27,7 +27,7 @@ enum RCTPlayerOperations {
 
         var selectedTrackIndex: Int = RCTVideoUnset
 
-        if (type == "disabled") || (type == "none") || (type == "") {
+        if type == "disabled" {
             // Select the last text index which is the disabled text track
             selectedTrackIndex = trackCount - firstTextIndex
         } else if type == "language" {
@@ -92,7 +92,7 @@ enum RCTPlayerOperations {
             return
         }
 
-        if (type == "disabled") || (type == "none") || (type == "") {
+        if type == "disabled" {
             // Do nothing. We want to ensure option is nil
         } else if (type == "language") || (type == "title") {
             let value = criteria?.value as? String


freeboub:
@mlazari Thank you very much for this PR ! I don't have much time to spend on react native video for now... I will release it soon !

PR #4447, opened 2025-03-03
chore: update utm links

moskalakamil:
## Summary
Update link leadings to external sites

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "react-native": "src/index",
     "license": "MIT",
     "author": "Community Contributors",
-    "homepage": "https://docs.thewidlarzgroup.com/react-native-video/",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "react-native": "src/index",
     "license": "MIT",
     "author": "Community Contributors",
-    "homepage": "https://docs.thewidlarzgroup.com/react-native-video/",
+    "homepage": "https://docs.thewidlarzgroup.com/react-native-video/?utm_source=rnv&utm_medium=npm&utm_campaign=homepage&utm_id=text",
```
freeboub: I don't think we should change the package.Json... it is not a place for analytics management.
There are some tools which parse the package.json to generate package list, the analytics info shall not be there!?

freeboub (changes requested):

package.json
KrzysztofMoch: Yeah I also think we shouldn't put it here

KrzysztofMoch (approved):

PR #4453, opened 2025-03-05
Fix: call startForeground() immediately to prevent ForegroundServiceDidNotStartInTimeException

mobily:
## Summary

This PR addresses a crash happening in the `VideoPlaybackService` when the app is offline and you switch between screens, resulting in the `ForegroundServiceDidNotStartInTimeException` error.

ADB logs:

```
02-25 14:52:37.242 31539 31539 E AndroidRuntime: FATAL EXCEPTION: main
02-25 14:52:37.242 31539 31539 E AndroidRuntime: Process: com.app.xyz, PID: 31539
02-25 14:52:37.242 31539 31539 E AndroidRuntime: android.app.ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{35eeb1f u0 com.app.xyz/com.brentvatne.exoplayer.VideoPlaybackService}
```

### Motivation

Android requires that any service started with `startForegroundService()` must call `startForeground()` within a limited short time frame. Previously, the implementation only called `startForeground()` when a player was successfully registered, which could be delayed or fail when the device was offline.

### Changes

- added an immediate call to `startForeground()` right in the `onStartCommand()` method using a placeholder notification
- created a placeholder notification for when the service starts up
- added a constant for the placeholder notification ID
- the original notification for the player still updates when a player registers

## Test Plan

1. Put the device in airplane mode (or turn off WiFi and mobile data).
2. Open the app and go to a screen with the video player.
3. Go back.
4. Check that the app doesn't crash with `ForegroundServiceDidNotStartInTimeException`.

Before this fix, the crash would happen every time the device was offline.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
@@ -63,7 +63,9 @@ class VideoPlaybackService : MediaSessionService() {
 
         mediaSessionsList[player] = mediaSession
         addSession(mediaSession)
-        startForeground(mediaSession.player.hashCode(), buildNotification(mediaSession))
+
+        val notificationId = player.hashCode()
+        startForeground(notificationId, buildNotification(mediaSession))
     }
 
     fun unregisterPlayer(player: ExoPlayer) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
```diff
@@ -224,7 +226,32 @@ class VideoPlaybackService : MediaSessionService() {
...
+                    NOTIFICATION_CHANEL_ID,
+                    NotificationManager.IMPORTANCE_LOW
+                )
+            )
+        }
+
+        // Return a minimal notification
+        return NotificationCompat.Builder(this, NOTIFICATION_CHANEL_ID)
+            .setSmallIcon(androidx.media3.session.R.drawable.media3_icon_circular_play)
+            .setContentTitle("Media playback")
```
freeboub: I don't like these hardcoded string ... (As I am french :) ) 
Can you either put empty string of extract to the resources (futur translations will be possible) ?

freeboub (changes requested):
looks good to me (just a small comment).
I ping @paul-rinaldi I think he will interested with this change

android/src/main/java/com/brentvatne/exoplayer/VideoPlaybackService.kt
mobily: @freeboub thanks for the suggestion @freeboub! it's all set now: https://github.com/TheWidlarzGroup/react-native-video/pull/4453/commits/1c4849331be7e98e06cef932821e0236bf3a59d8 and https://github.com/TheWidlarzGroup/react-native-video/pull/4453/commits/402706ed58aad296fd7e57dc4931b732e891eaca

freeboub (approved):

KrzysztofMoch (approved):
LGTM ✅
Thanks!

ArifSanaullah:
@mobily , We have implemented this patch in our app a couple of months back. It indeed fixed the `ForegroundServiceDidNotStartInTimeException` crash but also introduced another crash which stated this:

```
Fatal Exception: java.lang.RuntimeException
Unable to start service com.brentvatne.exoplayer.VideoPlaybackService@bb65e05 with null: android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service my.app.name/com.brentvatne.exoplayer.VideoPlaybackService
```

I BTW was calling `startForeground` without any version check. [Here](https://github.com/ArifSanaullah/react-native-video-anr) is the blank RN app with complete patch applied.

Apart from that, We have also seen ANRs spiking up when we made this change. [Here is the issue that I just created for ANRs.](https://github.com/TheWidlarzGroup/react-native-video/issues/4470)

cc: @KrzysztofMoch @freeboub

PR #4466, opened 2025-03-12
refactor(ios): add audio session manager

KrzysztofMoch:
## Summary
Created audio session manager that will ensure correct playback category. I should fix weird behavior where users are using conflict options that change playback category without checking if they can 

rework of https://github.com/TheWidlarzGroup/react-native-video/pull/3850

### Motivation
Fix yolo playback category overriding 

### Changes
- moved all audio session manager logic into on class that manage it across all views

## Test plan
- [x] Played with audio props in example app

### DIFF
--- a/ios/Video/AudioSessionManager.swift
+++ b/ios/Video/AudioSessionManager.swift
@@ -1,0 +1,325 @@
+import AVFoundation
+import Foundation
+
+class AudioSessionManager {
+    static let shared = AudioSessionManager()
+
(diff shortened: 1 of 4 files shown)


rustygloves:
I have confirmed that this resolves my issue https://github.com/TheWidlarzGroup/react-native-video/issues/4426#issuecomment-2718321066

freeboub (approved):

Issue #3640, opened 2024-04-02
[BUG]: Poster not working on android


LunatiqueCoder:
### Version

v6 (Beta)

### What platforms are you having the problem on?

Android

### Architecture

Old architecture

### What happened?

When using the `poster` prop, it works for a few seconds at the beginning, and then we have a blank screen.

### Reproduction

N/A

### Reproduction
N/A

### Solution: 
```diff
diff --git a/node_modules/react-native-video/src/Video.tsx b/node_modules/react-native-video/src/Video.tsx
index e496efc..5a97630 100644
--- a/node_modules/react-native-video/src/Video.tsx
+++ b/node_modules/react-native-video/src/Video.tsx
@@ -376,7 +376,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     );
 
     const _onReadyForDisplay = useCallback(() => {
-      setShowPoster(false);
+      setShowPoster(!!poster);
       onReadyForDisplay?.();
     }, [setShowPoster, onReadyForDisplay]);
```

freeboub:
Thank you for the patch, I will clarify the expected behavior soon. (But if you want to have a poster instead of the video, you should do it on the app side)

freeboub:
I open a PR to fix the issue. I think you should be able to test it !
changing source without remount didn't reset the setShowPoster value

LunatiqueCoder:
Hey @freeboub this seems to be reproducing on the latest version as well (v6.4.3) 😢

LunatiqueCoder:
@freeboub This is how I managed to fix it: 

```diff
diff --git a/node_modules/react-native-video/src/Video.tsx b/node_modules/react-native-video/src/Video.tsx
index 49ca1e9..5cc196c 100644
--- a/node_modules/react-native-video/src/Video.tsx
+++ b/node_modules/react-native-video/src/Video.tsx
@@ -452,7 +452,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     );
 
     const _onReadyForDisplay = useCallback(() => {
-      hasPoster && setShowPoster(false);
+      setShowPoster(hasPoster);
       onReadyForDisplay?.();
     }, [setShowPoster, hasPoster, onReadyForDisplay]);
 ```

Let me know if you want me to open a PR with this, it would be my pleasure to contribute.

moskalakamil:
[...]

PR #4478, opened 2025-03-19
bug/#3640-android-poster-not-working

LunatiqueCoder:
This PR should fix:
- #3640

### DIFF
--- a/src/Video.tsx
+++ b/src/Video.tsx
@@ -570,7 +570,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     );
 
     const _onReadyForDisplay = useCallback(() => {
-      hasPoster && setShowPoster(false);
+      hasPoster && setShowPoster(!!hasPoster);
       onReadyForDisplay?.();
     }, [setShowPoster, hasPoster, onReadyForDisplay]);
 


src/Video.tsx
```diff
@@ -570,7 +570,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     );
 
     const _onReadyForDisplay = useCallback(() => {
-      hasPoster && setShowPoster(false);
+      hasPoster && setShowPoster(!!hasPoster);
```
moskalakamil, line 573: Do you have any idea why it's works? Basically i can see, that !!hasPoster is always true here, so when the video is ready for display, we still show the poster?

LunatiqueCoder: Honestly, this was an issue we investigated long time ago and it was funny because it only happens on Android. If I remember right, yes, this was the behaviour.

`onReadyForDisplay` 👉 if it `hasPoster` then `setShowPoster(true)`
 
Unfortunately at the moment we don't have time do deep dive into it, but in our project we patched this change in all `react-native-video` versions in order to fix the `black-screen` poster issue on Android.

freeboub: I also don't understand how it can fix something... In your app, do you change the poster value during playback initialization? It may be a corner case, but setting it to false is the expected behavior here...

LunatiqueCoder: @freeboub Yes, depending if the file is a video or audio. If it's audio, we're looking for it's corresponding poster and update the poster after we find it. I believe this might happen during initialization, yes, just like you say.

KrzysztofMoch: I can see why this can fix issues. It's fine for me to merge this

moskalakamil: Maybe we should implement some more advanced logic? With this change, I think the poster will always be visible under the video whenever a user provides one

Issue #3074, opened 2023-03-24
Feature: HLS stream initial bitrate on iOS


chrstnwhlrt:
> I would like to provide an initial bandwidth for HLS streaming. Currently the Exoplayer in Android uses always the lowest bandwidth available by default and starts increasing the resolution after some chunks.
> 
> The Explayer allows to provide an initial bitrate estimate, which would solve this issue.
> Somehow a configuration parameter should be passed to the bandwidth meter in:
> 
> https://github.com/react-native-video/react-native-video/blob/edf7eee287ebe5fedd34aa1093fec91cdb7f1989/android/src/main/java/com/brentvatne/exoplayer/DefaultReactExoplayerConfig.java
> 
> Line 15: Add setInitialBitrateEstimate(magicParamComingFromReact) to the bandwidth meter builder param.

The initial bitrate issue has been resolved for Android (PR #4480), but iOS still lacks a way to control the starting stream quality.

mihailpresssport:
we have encountered this problem on IOS and it is still relevant to us. Video starts playing in terrible quality and improves over time, there must be a way to adjust this

vipulbhj:
Hello, any update on this ? For us, this is happening on ios

moskalakamil:
Hey @chrstnwhlrt, I created PR #4480 where I added this for Android. Unfortunately, on iOS, there's no straightforward way to achieve this, and we would need a workaround. I'll update this issue to focus on iOS only for now, but I don't have any plans to implement this on iOS at the moment so if anyone is interested in working on this, feel free to take it on! 🚀

PR #4480, opened 2025-03-20
feat(android): initial bitrate

moskalakamil:
## Summary
Added prop to set initial bitrate
### Motivation
#3074 
### Changes
- new prop `source.bufferConfig.initialBitrate`
- setInitialBitrate in ExoplayerConfig
- new Docs section
## Test plan
Run test android app with 
```
<Video
        // ...
        source={{
          // ...
          bufferConfig: {
            // ...
            initialBitrate: {number}
          }
        }}
        reportBandwidth={true}
        onBandwidthUpdate={e => console.log(e.bitrate)}
/>
```


freeboub (changes requested):
Initial bitrate should be a sub property of source (or even better source.bufferConfig).
The issue here under is when you mount the <Video/> component with a source and bitrate, setSrc and setInitialBitrate may not be synchronized, so in some corner case the initial bitrate may not be received before starting playback ...

KrzysztofMoch:
Yeah I agree - It would be better to move it into `source.bufferConfig` or s`source`

moskalakamil:
@freeboub Done

freeboub (approved):
Long time to approve but it looks good to me.

Just a note for you @moskalakamil.
If I remember well this initial bitrate will be apply on player initialization (I implemented similar feature in my past fork).
During your app life cycle, if you just change `<Video/>` source we don't release and recreate the whole player, then this parameter may not be used. It make sense as if exoplayer already know its expected bandwidth, no need to apply initial bandwidth.

Thank you!

PR #4489, opened 2025-03-26
update the json file.

ankitjainanktech:
## Test plan


freeboub:
It is an error I think ...

PR #4494, opened 2025-03-28
feat(android): add asset management functions to plugin

moskalakamil:
## Summary
Added support for 3 plugin hooks that allow native modules to:
- Override the `MediaItem.Builder`
- Override the `DataSource.Factory`
- Disable the caching system

### Motivation
Enable other native packages to inject custom logic — for example, to implement a custom caching system.

### DIFF
--- a/android/src/main/java/com/brentvatne/react/ReactNativeVideoManager.kt
+++ b/android/src/main/java/com/brentvatne/react/ReactNativeVideoManager.kt
@@ -1,5 +1,8 @@
 package com.brentvatne.react
 
+import androidx.media3.common.MediaItem
+import androidx.media3.datasource.DataSource
+import com.brentvatne.common.api.Source
 import com.brentvatne.common.toolbox.DebugLog
 import com.brentvatne.exoplayer.DRMManagerSpec
 import com.brentvatne.exoplayer.RNVExoplayerPlugin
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/brentvatne/react/ReactNativeVideoManager.kt
```diff
@@ -75,6 +78,48 @@ class ReactNativeVideoManager : RNVPlugin {
...
+                return null
+            }
+
+            val factory = plugin.overrideMediaDataSourceFactory()?.invoke(source, mediaDataSourceFactory)
+            if (factory != null) {
+                return factory
+            }
+        }
+        return null
+    }
```
KrzysztofMoch: Here is wrong logic, you are going through list of plugin and return __always__ first
So if we have list `[RNVPlugin, RNVExoplayerPlugin]` you will have `null`.
You should add variable before `pluginList.forEach` so you can set it from loop, and after loop return it 

It's same for all functions that you added

KrzysztofMoch (changes requested):
Left one comment, btw did you test it without any plugins to see if you didn't brake anything?

moskalakamil:
Updated. Yes, I tested it — the logic stays the same when no plugins are used, so it won’t affect the library unless plugins are in use

KrzysztofMoch (commented):
Great, please update docs and we are ready to go

KrzysztofMoch (approved):

PR #4500, opened 2025-04-02
chore(example): bump react-native & expo version

KrzysztofMoch:
## Summary
Bumped react-native versions in examples

### Changes
- bump expo sdk
- bump react-native
- bump `@react-native-picker/picker`

## Test plan
- run both examples


examples/expo/android/gradle.properties
```diff
@@ -35,7 +35,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=true
+newArchEnabled=false
```
freeboub, line 38: Is this really expected ?

examples/expo/package.json
```diff
@@ -18,35 +18,35 @@
   "dependencies": {
     "@expo/metro-runtime": "~4.0.0",
     "@react-native-picker/picker": "2.9.0",
-    "expo": "^52.0.7",
+    "expo": "^52.0.0",
```
freeboub, line 21: why did you downgrade expo version, I would expect: 52.0.44

freeboub (changes requested):
2 minor comments, but looks good

examples/expo/package.json
KrzysztofMoch: This was made automatically by expo, but let me explain
If we have `^52.0.0`, npm will search for latest `52.0.X` version - you can read more here https://docs.npmjs.com/about-semantic-versioning

examples/expo/android/gradle.properties
KrzysztofMoch: Yes, we don't support new arch (we do with interop layer, but it's very bugy) so we shouldn't have it in example

PR #4522, opened 2025-04-18
feat(ios): add `overridePlayerAsset` to `AVPlayerPlugin`

KrzysztofMoch:
## Summary
add `overridePlayerAsset` to `AVPlayerPlugin`

### Motivation
Allow plugins to override assets (eg. to change url, or replace with new one)

### DIFF
--- a/ios/Video/RNVAVPlayerPlugin.swift
+++ b/ios/Video/RNVAVPlayerPlugin.swift
@@ -37,7 +37,10 @@ open class RNVAVPlayerPlugin: RNVPlugin {
     open func onInstanceRemoved(id _: String, player _: AVPlayer) { /* no-op */ }
 
     /**
-     * Function called when a AVPlayer instance is being removed
+     * Function called when creating a new AVPlayerItem
+     * @param source: The VideoSource describing the video (uri, type, headers, etc.)
+     * @param asset: The AVAsset prepared by the player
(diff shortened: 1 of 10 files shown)


ios/Video/RNVAVPlayerPlugin.swift
```diff
@@ -36,6 +36,11 @@ open class RNVAVPlayerPlugin: RNVPlugin {
      */
     open func onInstanceRemoved(id _: String, player _: AVPlayer) { /* no-op */ }
 
+    /**
+     * Function called when a AVPlayer instance is being removed
+     */
```
moskalakamil: update comment please

moskalakamil (changes requested):
Left one comment - just to confirm, have you checked whether these plugins are called every time a new source is set? It's important to dynamically override the asset

ios/Video/RNVAVPlayerPlugin.swift
KrzysztofMoch: Done ✅

KrzysztofMoch:
> just to confirm, have you checked whether these plugins are called every time a new source is set? It's important to dynamically override the asset

It's working like on Android. We look for plugin that does not return null in `overridePlayerAsset`, and we are taking only one result from first plugin that return non null value

PR #4528, opened 2025-04-24
feat(web): allow styling overrides via the `style` prop

jonathanj:
## Summary

The existing `videoStyle` style is retained, but users can now override that with their own `style` prop.

It uses the `unstable_createElement` API from react-native-web, which allows the resulting component to use React Native's accessibility and style properties.

<https://necolas.github.io/react-native-web/docs/unstable-apis/#use-with-existing-react-dom-components>

### Motivation

The `Video` component on web claims to accept `style` (the type allows it), but that prop is entirely disregarded and instead of a fixed style is applied to the native `video` element. This makes it hard to style the video to let the browser choose an acceptable size, or to apply any app-specific styling.

There is an PR https://github.com/TheWidlarzGroup/react-native-video/pull/4333 to make a similar change, but the author indicated they are not comfortable using `unstable_createElement`, and suggests creating a new PR to make that change. Additionally, the old `videoStyle` has been removed, which seems potentially breaking for any existing web code that was depending on that styling to be in place.

### Changes

- Use `unstable_createElement` in a new wrapper component, to create a `video` component that accepts both web and React Native `style`

## Test plan

- Existing React Native code using `Video` does not see any new type errors
- Existing (React Native) web code using `Video` does not see any new type errors, but will now honour the `style` prop

### DIFF
--- a/src/Video.web.tsx
+++ b/src/Video.web.tsx
@@ -6,8 +6,26 @@ import React, {
   useRef,
   useState,
   type RefObject,
+  type CSSProperties,
 } from 'react';
+import {StyleProp, ViewStyle} from 'react-native';
+import {unstable_createElement} from 'react-native-web';
 import type {ReactVideoProps, VideoMetadata, VideoRef} from './types';
+
+// Define a style prop that is accepted and transformed by React Native Web
+// for the native `video` element.
+interface WebVideoElementProps
+  extends Omit<React.ComponentProps<'video'>, 'style'> {
+  style?: StyleProp<ViewStyle | CSSProperties>;
+}
+
+// Wrap the native `video` element to accept both React Native styles and CSS
+// styles.
+//
+// See <https://necolas.github.io/react-native-web/docs/unstable-apis/#use-with-existing-react-dom-components>
+function WebVideo(props: WebVideoElementProps) {
+  return unstable_createElement('video', props);
+}
 
 // stolen from https://stackoverflow.com/a/77278013/21726244
 const isDeepEqual = <T,>(a: T, b: T): boolean => {
@@ -28,6 +46,7 @@ const isDeepEqual = <T,>(a: T, b: T): boolean => {
 const Video = forwardRef<VideoRef, ReactVideoProps>(
   (
     {
+      style,
       source,
       paused,
       muted,
@@ -306,7 +325,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
     useMediaSession(src?.metadata, nativeRef, showNotificationControls);
 
     return (
-      <video
+      <WebVideo
         ref={nativeRef}
         src={src?.uri as string | undefined}
         muted={muted}
@@ -410,7 +429,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
           onVolumeChange?.({volume: nativeRef.current.volume});
         }}
         onEnded={onEnd}
-        style={videoStyle}
+        style={[videoStyle, style]}
       />
     );
   },
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
         "@release-it/conventional-changelog": "^7.0.2",
         "@types/jest": "^28.1.2",
         "@types/react": "~19.0.0",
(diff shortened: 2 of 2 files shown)


src/types/video.ts
```diff
@@ -281,11 +281,13 @@ export interface ReactVideoRenderLoaderProps {
...
 
-export interface ReactVideoProps extends ReactVideoEvents, ViewProps {
+export interface ReactVideoProps
+  extends ReactVideoEvents,
+    Omit<ViewProps, 'style'> {
   source?: ReactVideoSource;
   /** @deprecated Use source.drm */
   drm?: Drm;
-  style?: StyleProp<ViewStyle>;
+  style?: StyleProp<ViewStyle | CSSProperties>;
```
jonathanj: This change was [suggested in the original PR](https://github.com/TheWidlarzGroup/react-native-video/pull/4333#issuecomment-2614437684), but causes an tsc error in the native `Video.tsx` component, since `View` doesn't accept `React.CSSProperties` to `style`.

There are at least two options:

1. Remove this style here, and require users to force the value of `style` to `any` or something else when using `style` on web with web-only styles, since the types in `Video.web.tsx` are not exposed as part of the public API conditional to the platform.
2. Force the type of `style` internally in `Video.tsx`, to satisfy `tsc`, which could technically be considered incorrect since `React.CSSProperties` are not supported by `View` but the types wouldn't indicate that.

zoriya: Since using `unstable_createElement`, you don't need to specify this `| CSSProperties` since `ViewStyle` will automatically be converted to css.

`ViewStyle` already contains some web-only props, if someone wants to use some css props not supported on `ViewStyle` they need to use `any` (like with any other RN component)

src/Video.web.tsx
```diff
@@ -7,8 +7,24 @@ import React, {
...
+// Define a style prop that is accepted and transformed by React Native Web
+// for the native `video` element.
+type WebVideoElementProps = Omit<React.ComponentProps<'video'>, 'style'> &
+  Pick<ReactVideoProps, 'style'>;
+
+// Wrap the native `video` element to accept both React Native styles and CSS
+// styles.
+//
+// See <https://necolas.github.io/react-native-web/docs/unstable-apis/#use-with-existing-react-dom-components>
+const WebVideo = forwardRef<HTMLVideoElement, WebVideoElementProps>(
```
zoriya: i'm not sure we need to `forwardRef` here

zoriya (approved):
thanks for tackling this!

src/Video.web.tsx
```diff
@@ -7,8 +7,24 @@ import React, {
   useState,
   type RefObject,
 } from 'react';
+import {unstable_createElement} from 'react-native-web';
```
jonathanj, line 12: @zoriya Unfortunately `@types/react-native-web` doesn't export this. I guess I'll need to submit a PR to get that added there, or `@ts-ignore` this.

jonathanj: I went for `@ts-expect-error` with a description for now.

src/Video.web.tsx
jonathanj: You're right, we don't. I was getting errors about the `ref` prop at some point, but I think that was before I was correctly using the intrinsic `video` props.

src/types/video.ts
jonathanj: Makes sense. I've removed the changes to this file.

src/Video.web.tsx
jonathanj: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72588

If the PR is accepted, the `@ts-expect-error` will start failing, and since this project doesn't have a lockfile that might be on some unrelated PR.

Maybe it makes sense to wait for the types PR?

jonathanj: The new version of `@types/react-native-web` has been published, I've updated the package.json and removed the `@ts-expect-error`.

freeboub (approved):

KrzysztofMoch (approved):

PR #4570, opened 2025-06-12
fix(RCTVideo.swift): add audiovisualBackgroundPlaybackPolicy when pla…

Dawqss:
### Summary of Changes

* **Fixed iOS background playback behavior**:
  Added logic to set `audiovisualBackgroundPlaybackPolicy` on the player instance to ensure video continues playing when the app goes into the background.

```swift
if (_playInBackground) {
    _player!.audiovisualBackgroundPlaybackPolicy = .continuesIfPossible
} else {
    _player!.audiovisualBackgroundPlaybackPolicy = .automatic
}
```

* Without this, iOS would consistently pause video playback when the app was backgrounded, even if background playback was expected.

---
How to test:

Just play any video with react-native-video player and go to background, video should pause
### Motivation

Video stopped playing in background mode even though playInBackground was set to true.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -574,6 +574,14 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
 
             _player!.replaceCurrentItem(with: playerItem)
 
+            if #available(iOS 15.0, *) {
+                if _playInBackground {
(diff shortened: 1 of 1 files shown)


freeboub (approved):

KrzysztofMoch (changes requested):
Thank you for PR!
Our current min version of iOS is `13.0`, but this property is `15.0+` (from what I can see in docs). Can you add version availability check? Thanks!

KrzysztofMoch (approved):
Thanks!

PR #4579, opened 2025-06-24
fix(ios): allow audio mixing when none of the players are playing

lukasjengo:
## Summary
Noticing that on video component mount if mixWithOthers is set to default value "inherit", the external app (i.e. Spotify) audio gets paused even if the video is not playing. I feel like the pausing of external app audio should happen only when the video starts playing. At least this is what I would expect and see in other apps like Messenger.

### Changes
Default to audio mixing if none of the players are playing.


KrzysztofMoch (approved):
Make sense, Thanks!

PR #4581, opened 2025-06-26
feat(android): replace custom VideoView with media3 `PlayerView`

Pieczasz:
### Changes
*   **Simplified ExoPlayerView:** Replaces manual management of video surfaces, UI controls, and ad overlays with PlayerView's integrated components.
*   **Enhanced Controls & Subtitles:** Integrates PlayerView's native controls for a more consistent user experience, and improves subtitle styling and dynamic visibility based on available tracks. External subtitle support is now handled directly through ExoPlayer's MediaItem for better compatibility.
*   **Improved Track Selection:** Refines the logic for selecting audio and text tracks, providing more accurate reporting and enabling better control over track preferences.
*   **Stability & Performance:** Introduces a layout fix to address potential rendering issues with React Native's layout system, ensuring internal player views are sized correctly. The example app also includes performance enhancements for track change detection.
*   **Modernized Base:** Adopting `PlayerView` aligns the player with best practices, simplifies future maintenance, and leverages the latest features of the ExoPlayer library.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
@@ -32,6 +32,11 @@ class ExoPlayerView @JvmOverloads constructor(
         controllerShowTimeoutMs = 5000
         // Don't show subtitle button by default - will be enabled when tracks are available
         setShowSubtitleButton(false)
+        // Enable proper surface view handling to prevent rendering issues
+        setUseArtwork(false)
+        setDefaultArtwork(null)
+        // Ensure proper video scaling
+        resizeMode = androidx.media3.ui.AspectRatioFrameLayout.RESIZE_MODE_FIT
     }
 
     init {
@@ -39,7 +44,24 @@ class ExoPlayerView @JvmOverloads constructor(
     }
 
     fun setPlayer(player: ExoPlayer?) {
-        playerView.player = player
+        val currentPlayer = playerView.player
+        
+        if (currentPlayer != null && currentPlayer != player) {
+            playerView.player = null
+            
+            playerView.invalidate()
+            
+            post {
+                playerView.player = player
+                
+                if (player != null) {
+                    playerView.invalidate()
+                    requestLayout()
+                }
+            }
+        } else {
+            playerView.player = player
+        }
     }
 
     fun getPlayerView(): PlayerView {
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
@@ -13,6 +13,13 @@ import com.brentvatne.common.api.ResizeMode
 import com.brentvatne.common.api.SubtitleStyle
 import com.brentvatne.common.api.ViewType
 import android.view.View.MeasureSpec
+import androidx.media3.common.Player
+import androidx.media3.common.Timeline
+import android.view.SurfaceView
+import android.view.TextureView
+import android.graphics.drawable.GradientDrawable
+import android.widget.TextView
+import androidx.media3.ui.DefaultTimeBar
 
 @UnstableApi
 class ExoPlayerView @JvmOverloads constructor(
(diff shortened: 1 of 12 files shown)


android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -142,17 +142,8 @@ class FullScreenPlayerView(
...
-            val desc = if (isFullscreen) {
-                context.getString(androidx.media3.ui.R.string.exo_controls_fullscreen_exit_description)
-            } else {
-                context.getString(androidx.media3.ui.R.string.exo_controls_fullscreen_enter_description)
-            }
-            imageButton.setImageResource(imgResource)
-            imageButton.contentDescription = desc
-        }
+        // Note: This method is no longer needed since we're using PlayerView's built-in controls
+        // PlayerView handles fullscreen button state automatically
```
KrzysztofMoch: Can't we just remove this function, if we don't use it anymore?

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
```diff
@@ -226,13 +217,7 @@ class FullScreenPlayerView(
...
-        if (controlsConfig.hideNotificationBarOnFullScreenMode) {
-            val liveContainer = playerControlView?.findViewById<LinearLayout?>(com.brentvatne.react.R.id.exo_live_container)
-            liveContainer?.let {
-                val layoutParams = it.layoutParams as LinearLayout.LayoutParams
-                layoutParams.topMargin = 40
-                it.layoutParams = layoutParams
-            }
-        }
+        // Note: Live container adjustment is no longer needed since we're using PlayerView's built-in controls
+        // PlayerView handles UI adjustments automatically
```
KrzysztofMoch, line 212: Did you test any live to see if `PlayerView` handle this?

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
```diff
@@ -1,337 +1,237 @@
...
-            ViewType.VIEW_TYPE_SURFACE, ViewType.VIEW_TYPE_SURFACE_SECURE -> {
-                if (surfaceView !is SurfaceView) {
-                    surfaceView = SurfaceView(context)
-                    viewNeedRefresh = true
-                }
-                (surfaceView as SurfaceView).setSecure(viewType == ViewType.VIEW_TYPE_SURFACE_SECURE)
-            }
+        // PlayerView handles surface view internally
+        // This method is kept for compatibility but does nothing
+        // as PlayerView manages the surface automatically
```
KrzysztofMoch: This is breaking change, we need to allow users to set surface view that they want

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -248,7 +245,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private float mProgressUpdateInterval = 250.0f;
     protected boolean playInBackground = false;
     private boolean mReportBandwidth = false;
-    private boolean controls;
+    private boolean controls = true;  // Default to showing controls
```
KrzysztofMoch: nit: controls should be disabled by default, otherwise there is possibility that they will "flash" when user will want to disable them

KrzysztofMoch (changes requested):
It looks good, but there is a lot of "empty" functions that should be removed. Also I see that you removed possibility of setting `SurfaceView` type - this should be implemented also

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
Pieczasz: I removed it

android/src/main/java/com/brentvatne/exoplayer/FullScreenPlayerView.kt
Pieczasz: I just implemented a custom basic live badge, since the live indicator isn’t handled automatically.

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
Pieczasz: Changed to false for default

android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
Pieczasz: Created corresponding implementation to previous one

KrzysztofMoch (changes requested):
Please remove those empty functions that have comments
```
PlayerView manages shutter view automatically
This method is kept for compatibility
```

And I think we will be ready to merge

KrzysztofMoch (approved):

PR #4621, opened 2025-07-21
Update "Resolves a retain cycle and memory leak involving the RCTVideo instance when using Google IMA ads in react-native-video."

MdAbubakar:
Update TheWidlarzGroup/react-native-video#4574

iOS Google IMA Ads Container Not Resizing on Fullscreen or Rotation — Fixed by Adding AutoresizingMask

After recent changes to Google IMA integration, the ad container on iOS was not resizing when the player entered/exited fullscreen or changed orientation when using custom controls.

Cause:
The adContainerView added inside requestAds() had a fixed frame without autoresizingMask, causing layout issues.

Fix:
Added:
adContainerView.autoresizingMask = [.flexibleWidth, .flexibleHeight]

let adContainerView = UIView(frame: _video.bounds)
adContainerView.backgroundColor = .clear
adContainerView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
_video.addSubview(adContainerView)

### DIFF
--- a/ios/Video/Features/RCTIMAAdsManager.swift
+++ b/ios/Video/Features/RCTIMAAdsManager.swift
@@ -30,8 +30,14 @@
 
         func requestAds() {
             guard let _video else { return }
+            // fixes RCTVideo --> RCTIMAAdsManager --> IMAAdsLoader --> IMAAdDisplayContainer --> RCTVideo memory leak.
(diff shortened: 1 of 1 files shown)


MdAbubakar (commented):
iOS Google IMA Ads Container Not Resizing on Fullscreen or Rotation — Fixed by Adding AutoresizingMask

PR #4641, opened 2025-08-04
feat: bump nitro modules version

KrzysztofMoch:
## Summary

Updates dependencies, replaces `clean` with `dispose` for better memory management, and improves error handling.

### Changes

- **Dependency bump**: Updated `react-native-nitro-modules` to `>=0.27.2` across all package files
- **Memory management**: Replaced `clean()` method with `dispose()` in Android/iOS `HybridVideoPlayer` implementations
- **Error handling**: Simplified `VideoPlayer.ts` error logic - only call `onError` when it exists
- **Legacy support**: Added patch for RN versions < 0.80 to fix Android error handling in nitro-modules

**Breaking**: `clean()` method removed from VideoPlayer interface - use `dispose()` instead. Now we require `react-native-nitro-modules` >= `0.27.2`


packages/react-native-video/android/src/main/java/com/video/hybrids/videoplayer/HybridVideoPlayer.kt
```diff
@@ -351,6 +347,10 @@ class HybridVideoPlayer() : HybridVideoPlayerSpec() {
     }
   }
 
+  override fun dispose() {
```
mrousavy, line 350: nice!

PR #4643, opened 2025-08-07
Disable allowsVideoFrameAnalysis v7

noma4i:
## Summary

In iOS and macOS, the system automatically analyzes paused video content for text and objects. This results in a text recognition overlay(LiveText) appearing whenever text is present in the paused frame. The overlay can be visually distracting and diminishes the viewing experience by interfering with the paused image.

Like that:

image

Documentation: https://developer.apple.com/documentation/avkit/avplayerviewcontroller/allowsvideoframeanalysis

### Motivation

No one would find LiveText on video appealing, particularly when displaying Overlay Video where the video player functions as a background element.

During the loading process, I encountered a specific issue when displaying a video. While the video plays and pauses at the end as intended, it contains text elements including the app name and slogan. This creates an unexpected behavior where the Live Text icon becomes visible as the app continues to load in the background.

image

### Changes

`allowsVideoFrameAnalysis` set to `false`


noma4i:
@KrzysztofMoch I hope that helps to ship v7.

KrzysztofMoch:
Hey thanks for PR! I think this should be behind some prop, so users can enable and disable this - do you want me to do it or will you try do it yourself? Thanks!

noma4i:
@KrzysztofMoch the feature has never been optional due to some issues like dealing with source change while in a background. It's not the first time `allowsVideoFrameAnalysis` has to be set `false`

noma4i:
https://github.com/TheWidlarzGroup/react-native-video/issues/3900

KrzysztofMoch:
Okay let's merge this - maybe later I will create some prop and handle it proper in other PR

PR #4666, opened 2025-08-20
fix(ios): disable audio session management when no views

qusaieilouti99:
This PR fixes a major issue for calling apps.
When there are no video views, it is considered `disableAudioSessionManagement` as false
Therefore, it's fighting any other library trying to configure the audio session forever, which is, in my case, a callkit library

This guarantees that when there is a video, we rely on the prop; otherwise, disable it, the app might have several libraries that can manage the audio session.

### DIFF
--- a/ios/Video/AudioSessionManager.swift
+++ b/ios/Video/AudioSessionManager.swift
@@ -9,6 +9,11 @@ class AudioSessionManager {
     private var remoteControlEventsActive = false
 
     private var isAudioSessionManagementDisabled: Bool {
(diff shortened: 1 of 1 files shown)


qusaieilouti99:
The checks are broken; it's obvious that my changes are 5 lines and formatted correctly
Why these failures?

KrzysztofMoch (approved):
LGTM, thanks! 
About checks seems that we need to bump device on CI - I will fix it! Sorry for it 😅

PR #4672, opened 2025-08-26
chore(android): move package to twg namespace

KrzysztofMoch:
## Summary
Move android code to `com.twg.video` namespace to avoid conflicts and confusion


BasixKOR:
Commenting since this is not released yet: Is there a particular reason you didn't go with `com.thewidlarzgroup` instead? Conventionally Java package names starts with the domain name reversed, and it seems like you don't have TWG.com and it's very likely the company with the same name might want to get that domain

PR #4691, opened 2025-09-12
chore(android): enable flexible page sizes in native build configuration

seyedmostafahasani:
## Summary
This PR adds support for 16 KB memory page size on Android 15+ devices.
To achieve this, the `build.gradle` file was updated to enable flexible page size support in the native build process by passing the following argument to CMake:
```
externalNativeBuild {
  cmake {
    arguments "-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON"
  }
}
```
This ensures the React Native app can run correctly on newer Android devices that use a 16 KB page size by default.

### Motivation
Starting with Android 15, some devices use 16 KB memory pages instead of the traditional 4 KB pages.
Without enabling flexible page size support, apps may experience native crashes, memory mapping issues, or incompatibility on those devices.
This change ensures the app remains stable and forward-compatible with upcoming Android versions.

### Changes

- Updated build.gradle to pass the -DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON flag to CMake.
- No JavaScript/TypeScript or runtime logic was modified.
- This is a build-time configuration change only, affecting the native Android build process.

## Test plan

1. Build and install the app on an Android 15 emulator or device configured with a 16 KB page size.
2. Verify that the app launches and runs without native crashes related to memory mapping or page alignment.
3. Optionally, test on devices with 4 KB page size to confirm backward compatibility.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -142,6 +142,11 @@ android {
         buildConfigField "boolean", "USE_EXOPLAYER_DASH", ExoplayerDependencies["useExoplayerDash"].toString()
         buildConfigField "boolean", "USE_EXOPLAYER_HLS", ExoplayerDependencies["useExoplayerHls"].toString()
         buildConfigField "boolean", "USE_EXOPLAYER_RTSP", ExoplayerDependencies["useExoplayerRtsp"].toString()
+        externalNativeBuild {
+            cmake {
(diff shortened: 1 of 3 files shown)


KrzysztofMoch (approved):
Thanks!

seyedmostafahasani:
Anytime!

Issue #4686, opened 2025-09-12
[BUG]: Controls briefly showed when opening video, despite controls={false}


ivanyufen:
### Version

6.16.1

### What platforms are you having the problem on?

Android

### System Version

16

### On what device are you experiencing the issue?

Simulator

### Architecture

Old architecture

### What happened?

Hi, just upgrade my react-native-video to 6.16.1 from 6.1.1 and found that there is a controls showing in a split second despite i give controls={false}. After that, controls is no longer shown, only the first time opening a video

<img width="351" height="537" alt="Image" src="https://github.com/user-attachments/assets/95c2d953-fe63-48af-b7db-65fe26b0d30c" />

Any idea how to hide the controls when video first opened and buffer? thanks!

### Reproduction Link

repository link

### Reproduction

Step to reproduce this bug are:

ersyhsyn:
same for me

CarlosOmarRamirezZaragoza:
Not sure why but if source is null at first time and controls ={false}, the controls will work as expected

wn-na:
I also experienced the same issue and have submitted a pull request with the fix.

KiyeopYang:
me too

johhansantana:
> I also experienced the same issue and have submitted a pull request with the fix.

hey @wn-na I saw the PR was merged. I'm not really sure how can I get the fix as it seems I have the latest 6.16.1 version but this bug still happens

wn-na:
@johhansantana It seems the modified code hasn't been released yet.

Would you like to try the patch below?

`react-native-video+6.16.1.patch`
```
diff --git a/node_modules/react-native-video/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java b/node_modules/react-native-video/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
index bea60c0..10955a6 100644
--- a/node_modules/react-native-video/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/node_modules/react-native-video/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -456,7 +456,7 @@ public class ReactExoplayerView extends FrameLayout implements
     private void updateControllerVisibility() {
         if (exoPlayerView == null) return;
             
-        exoPlayerView.setUseController(!controlsConfig.getHideFullscreen());
[...]

PR #4698, opened 2025-09-19
fix: android control option is not work when init video

wn-na:
## Summary
On Android, the controls are set when the video first loads, regardless of the `controls` values.

### Motivation
fix for #4686 

### Changes
fix for function `updateControllerVisibility` in setUseController when controls is true


KrzysztofMoch:
Thanks!

PR #4708, opened 2025-09-28
Handle multiples event listeners

### LINKED ISSUE #4631: [v7]: `useEvent` override other events
### Version

v7.0.0

iOS, Android

### System Version

/

Real device

### Architecture

### What happened?

Events listeners are unique for a player and this shouldn't be the case. Two components should be able to listen to the same event.

Example: If i have a `PlayButton` like this:

```ts
export const PlayButton = ({ player }: { player: VideoPlayer }) => {
	const [playing, setPlay] = useState(player.isPlaying);
	useEvent(player, "onPlaybackStateChange", (status) => {
		setPlay(status.isPlaying);
	});

	return (
		<IconButton
			icon={playing ? Pause : PlayArrow}
			onPress={() => {
				if (playing) player.pause();
				else player.play();
			}}
		/>
	);
};
```

only the last `PlayButton` instantiated gets playback change events, while the others never updates.

I think `useEvent` should look like this:
```ts
export const useEvent = <T extends Events>(
  player: VideoPlayer,
  event: T,
  callback: (...args: EventsParams<T>) => void
) => {
  const id = useId();
  useEffect(() => {
    // @ts-expect-error we narrow the type of the event
    player[event][id] = callback;

    return () => {
      if (event === 'onError') {
        // onError is not native event, so we can set it to undefined
        player.onError[id] = undefined;
      } else {
        player.clearEvent(event, id);
      }
    };
  }, [player, event, id, callback]);
};
```

### Reproduction Link

/

### Reproduction

/

zoriya:
## Summary

Use a `addEventListener`/`removeEventListener` based system instead of manually assigning `onEvent` parameters.

### Motivation

Fixes #4631 

## Test plan

I tested my branch on a project that suffered from this issue.

### DIFF
--- a/docs/docs/events/events.md
+++ b/docs/docs/events/events.md
@@ -87,17 +87,17 @@ import { VideoPlayer } from 'react-native-video';
 
 const player = new VideoPlayer('https://example.com/video.mp4');
 
-player.onLoad = (data) => {
+player.addEventListener('onLoad', (data) => {
   console.log('Video loaded! Duration:', data.duration);
-};
+});
 
-player.onProgress = (data) => {
+player.addEventListener('onProgress', (data) => {
   console.log('Current time:', data.currentTime);
-};
+});
 
-player.onError = (error) => {
+player.addEventListener('onError', (error) => {
   console.error('Player Error:', error.code, error.message);
-};
+});
 
 player.play();
 ```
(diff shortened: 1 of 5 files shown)


KrzysztofMoch:
Nice work! I will try to review it this week

KrzysztofMoch:
@zoriya I want to make release, so I will merge this as it is and apply some changes myself - Thanks!

PR #4713, opened 2025-09-30
fix(AudioSessionManager.swift): notification control hide after pausing video

Dawqss:
## Summary
Don't allow .mixWithOthers when notification control or background playback are enabled

### Motivation
Notification control hide when pressing pause and user cannot resume video anymore from notification

### Changes
Block possibility to add to options .mixWithOthers when any player uses notification controls

### Test Scenario

1. Mount component with video
2. Play start
3. Observe that control notification appears
4. Pause video by passing prop paused={true}
5. Observe that control notification vanish

### DIFF
--- a/ios/Video/AudioSessionManager.swift
+++ b/ios/Video/AudioSessionManager.swift
@@ -152,14 +152,16 @@ class AudioSessionManager {
             return view._mixWithOthers == "mix" || view._mixWithOthers == "duck"
         }
 
-        let canAllowMixing = anyPlayerWantsMixing || (!anyPlayerShowNotificationControls && !anyPlayerNeedsBackgroundPlayback)
+        let anyPlayerUsesNotificationControlFeature = anyPlayerShowNotificationControls && anyPlayerNeedsBackgroundPlayback
+
+        let canAllowMixing = anyPlayerWantsMixing || !anyPlayerUsesNotificationControlFeature
 
         if isAudioSessionManagementDisabled {
             // AUDIO SESSION MANAGEMENT DISABLED BY USER
             return
         }
 
-        if !anyPlayerPlaying {
+        if !anyPlayerPlaying && !anyPlayerUsesNotificationControlFeature {
             options.insert(.mixWithOthers)
         } else if canAllowMixing {
             let shouldEnableMixing = videoViews.allObjects.contains { view in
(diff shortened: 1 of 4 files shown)


ios/Video/Features/RCTPictureInPicture.swift
```diff
@@ -6,7 +6,7 @@ import React
 
 #if os(iOS)
     class RCTPictureInPicture: NSObject, AVPictureInPictureControllerDelegate {
-        public private(set) var _pipController: AVPictureInPictureController?
```
Dawqss, line 9: swiftformat changed this file

ios/Video/NowPlayingInfoCenterManager.swift
```diff
@@ -81,7 +81,7 @@ class NowPlayingInfoCenterManager {
         }
     }
 
-    public func cleanup() {
```
Dawqss, line 84: swiftformat changed this file

ios/Video/AudioSessionManager.swift
```diff
@@ -152,14 +152,16 @@ class AudioSessionManager {
...
+        let anyPlayerUsesNotificationControlFeature = anyPlayerShowNotificationControls && anyPlayerNeedsBackgroundPlayback
+
+        let canAllowMixing = anyPlayerWantsMixing || !anyPlayerUsesNotificationControlFeature
 
         if isAudioSessionManagementDisabled {
             // AUDIO SESSION MANAGEMENT DISABLED BY USER
             return
         }
 
-        if !anyPlayerPlaying {
```
Dawqss, line 162: https://github.com/TheWidlarzGroup/react-native-video/commit/2d7e03942d5e2525e5c71dc914926ba07f67d54a

cc: @lukasjengo 

will this change break functionality you added?

lukasjengo: I think this change might cause issues, but only for our specific use case. Correct me if I’m wrong, but with this PR the audio mixing logic would work like this:

If Spotify is playing in the background using the notification control feature, and react native video plays a video with mixWithOthers enabled, the video audio will be muted and only Spotify’s audio will come through.

Our expectation is that, when mixWithOthers is set, Spotify should continue playing alongside the react native video audio. That said, this might be a niche requirement on our side, and I’m not sure what Apple’s official guidelines are for this behavior.

KrzysztofMoch: Yeah I wonder the same, 
it is question if it should be allowed to use `mixWithOthers` when using notification center 🤔

Dawqss: I will check those scenarios & check iOS guideline and I will be back (it can take some time)

PR #4743, opened 2025-10-24
feat: add E2E tests

KrzysztofMoch:
## Summary
Adds ([React Native Harness](https://www.react-native-harness.dev/)) E2E Tests for react-native-video


.github/workflows/tests.yml
```diff
@@ -0,0 +1,222 @@
...
+
+on:
+  push:
+    branches: [v7]
+    paths:
+      - ".github/workflows/tests.yml"
+      - "example/android/**"
+      - "**/nitrogen/generated/shared/**"
+      - "**/nitrogen/generated/android/**"
+      - "**/nitrogen/generated/ios/**"
```
moskalakamil, line 15: Probably we don't need this here, we have all the folders listed which triggers changes to the nitrogen folder

moskalakamil (approved):

moskalakamil:
Also, I’m not sure if we should call them e2e tests - these are just harness tests

PR #4747, opened 2025-10-30
feat(ios): add PublicAudioSessionManager for audio session management…

pioner92:
Added a public PublicAudioSessionManager class on iOS to allow enabling or disabling the audio session manager directly from the native side.
This makes it possible for other native modules or app code to manage the audio session state without relying solely on JS integration.
**Useful for call apps**

### DIFF
--- a/ios/Video/AudioSessionManager.swift
+++ b/ios/Video/AudioSessionManager.swift
@@ -7,6 +7,7 @@ class AudioSessionManager {
     private var videoViews = NSHashTable<RCTVideo>.weakObjects()
     private var isAudioSessionActive = false
     private var remoteControlEventsActive = false
+    private var isAudioSessionManagementForcedDisabled = false
 
     private var isAudioSessionManagementDisabled: Bool {
         // If no views are registered, disable audio session management
@@ -42,6 +43,10 @@ class AudioSessionManager {
     }
 
     // MARK: - Public API
+
+    func setIsAudioSessionManagementForcedDisabled(disabled:Bool) {
+        isAudioSessionManagementForcedDisabled = disabled
+    }
 
     func registerView(view: RCTVideo) {
         if videoViews.contains(view) {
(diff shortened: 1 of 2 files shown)


KrzysztofMoch:
You can use [`disableAudioSessionManagement`](https://docs.thewidlarzgroup.com/react-native-video/docs/v6/component/props#disableaudiosessionmanagement) to do it

pioner92:
@KrzysztofMoch 
I needed to disable the audio manager before calling `reportIncomingCall` to be able to configure the audio session for the call immediately
When using this prop, I would need to send an event to the JS side to change this flag, which would be asynchronous and not optimal
In all other cases, this prop should remain true. In my case, I only disable it during the call and then re-enable it afterward to remove the restriction

KrzysztofMoch:
Okay it's reasonable - but in that case I think we need new function to disable audio session or restart it? Let's say we have active audio session and RNV it's controlling it. Then we disable audio session (by setting variable to `true`) but session is still active as we don't disable it. WDYT?

Btw I have fixed CI builds, you can rebase if it isn't problem

pioner92:
In my case, when I disable the RNV audioManager, I initialize the PlayAndRecord category and perform a full configuration for the call.
After the call ends, I reset the session to category: playback and mode: default, change the flag to false, and re-enable the RNV audioManager — and as a result, it works well

pioner92:
I just needed to temporarily disable RNV’s constant attempts to modify the audio session during the call ( synchronously )

pioner92:
@KrzysztofMoch  Do you suggest something like this?
```swift
   func setIsAudioSessionManagementForcedDisabled(disabled: Bool) {
        isAudioSessionManagementForcedDisabled = disabled
        
        if !disabled {
            updateAudioSessionConfiguration()
        }
    }
```

pioner92:
> but session is still active as we don't disable it. WDYT?

We are still using the audioSession inside the app, so we don't need to deactivate it

KrzysztofMoch:
Yeah, I think we are good to merge this as it is - I will let CI finish and then merge it.
Thanks!

PR #4753, opened 2025-10-31
Fix: use top-most presented view controller for fullscreen presentation on iOS

Jianlong-Nie:
This PR fixes a layout issue on iOS when using .fullScreen presentation style.

Previously, RCTVideo would use firstAvailableUIViewController() or the root controller’s last child to present the fullscreen player.
However, this could result in incorrect layout updates when exiting fullscreen — especially in apps using react-navigation, tab scenes.

By switching to RCTPresentedViewController() and walking up the presentedViewController chain, we ensure that the fullscreen player is always presented from the top-most visible view controller.
This resolves issues where exiting fullscreen would cause the underlying content to shift upward.

before:

https://github.com/user-attachments/assets/61e176f5-ea21-465d-9584-49920fbac1c0

after:

https://github.com/user-attachments/assets/6a918fc1-6fe1-401b-a635-656167253e0d

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -1090,15 +1090,9 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
             _playerViewController?.modalPresentationStyle = .fullScreen
 
             // Find the nearest view controller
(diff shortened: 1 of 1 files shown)


KrzysztofMoch (approved):
LGTM, Thanks! 
CI is not passing as macos version change and `iPhone 14` is not available there - I will fix this

PR #4754, opened 2025-11-02
feat(android): Add expo config for notification controls

Arthi-chaud:
## Summary

Adding Expo configuration for notification controls on Android

Sources:
- v6 doc: [Link](https://docs.thewidlarzgroup.com/react-native-video/docs/v6/component/props#shownotificationcontrols)

### Notes

- First contribution, lmk if the contributed code doesn't comply to norm/coding style
- The plugin is invoked unconditionally because there is not toggle setting for notifications controls, unlike in v6. It might be relevant to add a boolean in the settings to enable/disable the plugin. I can add do that in this PR if need be.

### Motivation

Including this will avoid the user to write the configuration themselves.
(It can be a pain for unexperiences users)

### Changes

- Add Expo config for notification controls 
- Call config function in main plugin

## Test plan

- [ ] Test using `example` app
   - Can't manage to build/run it locally, but I added that same config in my project ([link](https://github.com/Arthi-chaud/Meelo/blob/main/front/apps/mobile/plugins/withMediaServicePlugin.ts)) and it works.

### DIFF
--- a/packages/react-native-video/src/expo-plugins/withAndroidNotificationControls.ts
+++ b/packages/react-native-video/src/expo-plugins/withAndroidNotificationControls.ts
@@ -1,0 +1,43 @@
+import {
+  AndroidConfig,
+  type ConfigPlugin,
+  withAndroidManifest,
+} from '@expo/config-plugins';
+
+export const withAndroidNotificationControls: ConfigPlugin = (oldConfig) => {
+  return withAndroidManifest(oldConfig, (config) => {
+    const mainApplication = AndroidConfig.Manifest.getMainApplication(
+      config.modResults
+    );
+    if (!mainApplication) {
+      console.warn(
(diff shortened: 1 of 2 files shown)


KrzysztofMoch (approved):
Thank you! I did miss that one when implementing notificationControls

jcolicchio:
I noticed that this plugin has no effect if `mainApplication.service` is undefined / null; the reference code cited in this PR has an additional line `mainApplication.service ??= []` missing from this implementation.

However, this plugin likely should check that the service hasn't already been appended, otherwise the plugin will not be idempotent and will add a new service to the existing manifest every time it's invoked.

Also, the lines adding permissions appear to do nothing in Expo; I believe these permissions are meant to be added to `app.json` under `"android"` instead?

@KrzysztofMoch should I open an issue for this apparent bug?

PR #4763, opened 2025-11-05
fix: prevent audiovisualBackgroundPlaybackPolicy crash

langemike:
This prevents an error when building your app for tvOS. I assume this also happens when building for visionOS.

It fixes the following error message: ` audiovisualBackgroundPlaybackPolicy is only available in tvOS 15.0 or newer.`

Screenshot 2025-11-05 at 12 36 17


KrzysztofMoch (approved):
Thank you!

Issue #4783, opened 2025-11-24
[BUG]: React Native Video v7 on Android does not resolve production raw resources


jiroscripts:
### What happened?

**[Android] Loading video from `require()` fails in production due to raw resource lookup**

When trying to load a video using `require()` on Android in production (`__DEV__ === false`), `react-native-video` does not automatically resolve the resource from `/res/raw`. In development mode, it works via the Metro bundler, but in a production build the URI is not correctly resolved.

**Expected behavior:**
The video should be loaded correctly from the `res/raw` folder without manually constructing the URI.

**Current behavior:**
The player fails to load the video because the native code does not look in `raw`.

#### Workaround

```javascript
import { Platform, Image } from 'react-native';
import DeviceInfo from 'react-native-device-info';

const fixedVideoSource =
    Platform.OS === 'android' && !__DEV__
        ? {
              uri: `android.resource://${DeviceInfo.getBundleId()}/raw/${Image.resolveAssetSource(
                  videoSource
              ).uri}`,
          }
        : videoSource;
```

### Steps to reproduce

1. Load a video on Android using `require()` (a local asset).
2. Run the app in development mode (`__DEV__ === true`) and confirm the video plays via the Metro bundler.
3. Build and run a production/release build (`__DEV__ === false`).
4. Observe that the player fails to load the video because the native code does not resolve the resource from `/res/raw`.

### Reproduction repository

https://github.com/jiroscripts/react-native-video/tree/test/local-asset-for-android-release

### react-native-video version

7.0.0-alpha.10

### react-native version

0.81.4

### Platforms

Android

### OS version

Android 16

### Device

Real device

### Architecture

New Architecture

### Media / source type

Local file

TWG-AI:
Thanks for the report and for including a workaround!

For anyone hitting this in the meantime, the workaround from the issue is to construct the `android.resource://` URI manually for production Android builds:

```js
const fixedSource =
  Platform.OS === 'android' && !__DEV__
    ? { uri: `android.resource://${DeviceInfo.getBundleId()}/raw/${Image.resolveAssetSource(videoSource).uri}` }
    : videoSource;
```

This has been logged for the team to add proper production raw resource resolution to v7's Android source handling.

> I'm an automated assistant — double-check the workaround against your exact setup before relying on it.

PR #4784, opened 2025-11-25
Fix/local asset for android release

jiroscripts:
## 🚀 Summary

This PR fixes an issue (#4783) where videos fail to load on **Android in production mode** when using `require()`.

In **development mode (`__DEV__ === true`)**, Metro correctly resolves the asset dynamically. However, in **release builds**, the URI returned by `resolveAssetSource()` is not correctly interpreted by the native player, which causes playback failures.

To address this, the URI is now explicitly normalized to correctly reference raw resources located in `res/raw`, ensuring functional playback across both debug and release builds.

## 🔍 Motivation

- On Android, assets bundled at build time are placed inside `res/raw`.
- Unlike development mode, the release build does **not automatically map asset identifiers to the correct native URI**.
- As a result, passing a resource ID or asset reference (e.g. via `require()`) fails when using native video playback (such as ExoPlayer).
- By normalizing the URI, we guarantee the correct resolution of the video asset regardless of the build mode.

## 🛠️ Technical details

- React Native uses `resolveAssetSource()` to retrieve asset metadata.
- In debug mode, Metro resolves the URI dynamically.
- In release mode, a fallback resolution returns an Android resource identifier (e.g. `__libs_assets_src_lib_images_splash`) which cannot be used directly by the native video player.
- We replicate React Native’s own logic for selecting the Android asset folder (`raw`) based on:
  https://github.com/facebook/react-native/blob/be947078425c972a8a0934535d3819b60e488a33/packages/assets/path-support.js#L58-L60
- A normalized URI is generated via `RawResourceDataSource.buildRawResourceUri(...)`.

## ✨ Changes

- Added a new method `normalizeUri(input: String)` to determine whether the input is:
  - A valid URI (e.g. `http`, `file`, `content`, `data`) → returned unchanged.
  - A local bundled asset name → resolved into a valid raw resource URI.

```kotlin
private fun normalizeUri(input: String): String {
  val parsedUri = Uri.parse(input)

  if (parsedUri.scheme != null) {
    return parsedUri.toString()
  }

  val resId = appContext.resources
    .getIdentifier(input, "raw", appContext.packageName)

  if (resId == 0) {
    throw IllegalArgumentException("The video resource '$input' could not be found in res/raw")
  }

  val mediaUri = RawResourceDataSource.buildRawResourceUri(resId)
  return mediaUri.toString()
}
```

- Updated `fromUri()` to use this new resolver:

```kotlin
override fun fromUri(uri: String): HybridVideoPlayerSourceSpec {
  val config = NativeVideoConfig(
    uri = normalizeUri(uri),
    externalSubtitles = null,
    drm = null,
    headers = null,
    bufferConfig = null,
    metadata = null,
    initializeOnCreation = true
  )

  return HybridVideoPlayerSource(config)
}
```

This ensures that all input video paths are correctly resolved before being passed to the native player, improving reliability in both development and production environments.

## 📋 Test plan

To be completed:

- [ ] Add a local video asset to the example app.
- [ ] Load it using `require('./video.mp4')`.
- [ ] Verify playback works in:
  - Debug mode (`npm run android`)
  - Release mode (`./gradlew assembleRelease`)
- [ ] Log the normalized URI in native code to confirm proper resolution.
- [ ] (Optional) Add instrumentation tests to validate fallback behavior when the resource is missing.

🙏 Let me know if you'd like to add:
- Edge case handling (extensions, asset folder fallback)
- Breaking change notice
- A migration note for current users


packages/react-native-video/android/src/main/java/com/twg/video/hybrids/videoplayersource/HybridVideoPlayerSourceFactory.kt
```diff
@@ -1,12 +1,43 @@
...
 import com.facebook.proguard.annotations.DoNotStrip
+import com.twg.video.react.AppContextHolder
 
 @DoNotStrip
 class HybridVideoPlayerSourceFactory: HybridVideoPlayerSourceFactorySpec() {
+  private val appContext: Context
+    get() = AppContextHolder.appContext
+      ?: throw IllegalStateException(
+        "AppContextHolder.appContext has not been initialized."
+      )
```
KrzysztofMoch: No need for that additional app context logic - we can access app context via `NitroModules.applicationContext`. It can be null so we still need to check it

KrzysztofMoch (changes requested):

packages/react-native-video/android/src/main/java/com/twg/video/hybrids/videoplayersource/HybridVideoPlayerSourceFactory.kt
jiroscripts: Thanks, I wasn't aware of that. I updated the logic accordingly — let me know if there's anything else I should adjust.

jiroscripts: Hey 🙂 Is there anything I can do to finish this one ?

KrzysztofMoch (approved):
Thank You!

PR #4811, opened 2025-12-31
fix(ios): fixed audio focus change on iOS

robertert:
The previous implementation of audio mixing mode selection had a bug where switching between different modes (e.g., from .mixWithOthers to .doNotMix) did not consistently remove the previously set options. This was because the switch statement only handled removals in specific cases, potentially leaving stale flags in audioSessionCategoryOptions.

### DIFF
--- a/packages/react-native-video/ios/core/VideoManager.swift
+++ b/packages/react-native-video/ios/core/VideoManager.swift
@@ -155,7 +155,7 @@ class VideoManager {
     
     configureAudioSession()
   }
-  
+
   private func configureAudioSession() {
     let audioSession = AVAudioSession.sharedInstance()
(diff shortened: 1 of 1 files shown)


Okelm:
is this connected to some reported issue somewhere?

robertert:
no i was just playing with the example app and found this issue

Okelm:
by any chance it might fix this? https://github.com/TheWidlarzGroup/react-native-video/issues/4810

robertert:
This issue doesn't occur here.
But I made this change in v7 version and the issue was on 6.11 so I suppose it could be fixed earlier.

moskalakamil (approved):

PR #4765, opened 2025-11-07
feat: add DAI support

maciekBudzinski:
## Summary

This PR adds Dynamic Ad Insertion (DAI) support for both Android and iOS.

| iOS Demo | Android Demo |
|----------|--------------|
| iOS Demo | Android Demo |

### Motivation

Right now, the react-native-video only supports client-side ad insertion (CSAI) where the player requests ads separately and plays them during breaks. Adding DAI support would enable server-side ad insertion (DAI), expanding the types of ads that can be used with react-native-video.

It's also feature requested by the community: https://github.com/TheWidlarzGroup/react-native-video/issues/4399

#### JS

- Added TypeScript type definitions for DAI configuration, supporting both VOD (with `contentSourceId` and `videoId`) and Live (with `assetKey`) modes using discriminated unions for type safety
- Updated the main `Video` component to accept and handle the `dai` prop in source configuration
- Modified source parsing logic to allow empty URI when DAI is configured (since IMA SDK handles stream requests directly)
- Updated native component specs to include DAI configuration in the bridge interface

#### iOS

- Created data structures to parse and store DAI configuration from React Native bridge
- Integrated DAI support into the video source pipeline and IMA ads manager
- Implemented DAI stream handling to request and play streams via IMA SDK
- Added support for both VOD and Live DAI streams with backup stream fallback
- Minor code cleanup (removed unnecessary access modifiers, formatting improvements)

#### Android

- Created Kotlin classes to parse DAI configuration from React Native bridge with proper null safety
- Updated source parsing to accept DAI configuration and allow empty URIs when DAI is used
- Implemented DAI stream request logic using ExoPlayer's IMA extension for both VOD and Live streams
- Added support for custom ad tag parameters and automatic fallback to backup streams
- Created stub classes for IMA DAI extension integration
- Updated Media3 version from `1.4.1` to `1.8.0` to support DAI features

#### Testing App

- Created a complete Expo example app (`examples/expo-dai/`) demonstrating DAI functionality
- Built UI components for configuring and testing both VOD and Live DAI streams
- Added real-time event logging to monitor video and ad events
- Added comprehensive README with installation steps, usage guide, and known issues

#### Documentation

- Added comprehensive `dai` prop documentation to the props reference page with property descriptions, examples for both VOD and Live modes, and usage notes
- Updated the ads documentation page with a dedicated DAI section explaining server-side ad insertion, configuration options, event handling, backup stream functionality, and a comparison table between DAI and CSAI

## Test plan

[...]

PR #4816, opened 2026-01-09
feat!: add DAI support

fnwk:
Note: This PR is based on #4765 (feat: add DAI support) by @maciekBudzinski 

## Summary

This PR adds Dynamic Ad Insertion (DAI) support for both Android and iOS, while also unifying the ad configuration API.

| iOS Demo | Android Demo |
|----------|--------------|
| iOS Demo | Android Demo |

### Motivation

Right now, react-native-video only supports client-side ad insertion (CSAI) where the player requests ads separately and plays them during breaks. Adding DAI support enables server-side ad insertion, expanding the types of ads that can be used with react-native-video.

This is also a feature requested by the community: https://github.com/TheWidlarzGroup/react-native-video/issues/4399

#### Breaking Change: Unified Ad Configuration API

The ad configuration has been unified into a single `source.ad` property with a `type` discriminator:

**Before (CSAI only):**
```jsx
<Video
  adTagUrl="https://example.com/vast"
  adLanguage="en"
/>
```

**After (CSAI):**
```jsx
<Video
  source={{
    uri: 'https://example.com/video.mp4',
    ad: {
      type: 'csai',
      adTagUrl: 'https://example.com/vast',
      adLanguage: 'en'
    }
  }}
/>
```

**After (DAI VOD):**
```jsx
<Video
  source={{
    ad: {
      type: 'dai',
      contentSourceId: '2548831',
      videoId: 'tears-of-steel',
      format: 'hls', // or 'dash' (Android only)
      fallbackUri: 'https://example.com/backup.m3u8'
    }
  }}
/>
```

**After (DAI Live):**
```jsx
<Video
  source={{
    ad: {
      type: 'dai',
      assetKey: 'c-rArva4ShKVIAkNfy6HUQ',
      fallbackUri: 'https://example.com/backup.m3u8'
    }
  }}
/>
```

> Note: The old `adTagUrl` and `adLanguage` props are still supported but deprecated.

#### JS

- Added TypeScript type definitions for unified ad configuration using discriminated unions (`AdConfigCSAI`, `AdConfigDAI`, `AdConfigDAIVod`, `AdConfigDAILive`)
- Updated the main `Video` component to normalize ad configuration and handle both CSAI and DAI through the unified `source.ad` prop
- Modified source parsing logic to allow empty URI when DAI is configured (since IMA SDK handles stream requests directly)
- Added `format` property to DAI configuration for HLS/DASH stream format selection (Android only)

#### iOS

- Created `AdParams` data structure to parse and store ad configuration from React Native bridge
- Integrated DAI support into the video source pipeline and IMA ads manager
- Implemented DAI stream handling to request and play streams via IMA SDK
- Added support for both VOD and Live DAI streams with backup stream fallback
- Added support for custom `adTagParameters`

#### Android

- Extended `AdsProps` Kotlin class to parse unified ad configuration with DAI properties
- Updated source parsing to accept DAI configuration and allow empty URIs when DAI is used
- Implemented DAI stream request logic using ExoPlayer's IMA extension (`ImaServerSideAdInsertionMediaSource`) for both VOD and Live streams
- Added support for custom ad tag parameters and automatic fallback to backup streams
- Added `format` property support for HLS/DASH stream format selection via `ImaServerSideAdInsertionUriBuilder`
- Fixed broken `wait()` call in DAI initialization that could cause issues
- Created stub classes for IMA DAI extension integration
- Updated Media3 version from `1.4.1` to `1.8.0` to support DAI features

#### Documentation

- Updated `source.ad` prop documentation with both CSAI and DAI configuration options
- Added `format` property documentation for DAI streams
- Updated the ads documentation page with a dedicated DAI section explaining server-side ad insertion, configuration options, event handling, backup stream functionality, and a comparison table between DAI and CSAI
- Marked old `adTagUrl` and `adLanguage` props as deprecated

## Test Plan

A complete test app is available at `examples/expo-dai/` that demonstrates DAI functionality with both VOD and Live streams. The app includes a UI for configuring DAI parameters (including format selection), real-time event logging, and controls for testing various playback scenarios. The README in `examples/expo-dai/` is worth reading before testing as it contains important setup instructions, usage guidelines, and known issues.

Key Testing Scenarios:

- **VOD DAI streams** - Verify that VOD streams with `contentSourceId` and `videoId` play correctly on both iOS and Android
- **Live DAI streams** - Verify that Live streams with `assetKey` play correctly on both iOS and Android
- **Stream format** - Test both HLS and DASH formats on Android (iOS auto-detects)
- **Ad events** - Confirm that ad events are properly received via `onReceiveAdEvent` callback
- **Backup stream fallback** - Test that the player automatically falls back to `fallbackUri` when DAI stream fails to load
- **Custom ad tag parameters** - Verify that `adTagParameters` are correctly passed through to the IMA SDK
- **CSAI backward compatibility** - Verify that existing CSAI implementations still work with the new unified API
- **Deprecated props** - Verify that old `adTagUrl` prop still works (with deprecation warning in docs)
- **Error handling** - Test behavior with missing required parameters

Note: Picture-in-Picture testing requires a real iOS device as simulators don't reliably support PiP. Google's official sample streams may have known issues with HLS VOD streams stopping around 1:10 (documented in the example app README).


android/src/main/java/com/brentvatne/common/api/Source.kt
```diff
@@ -214,25 +211,19 @@ class Source {
...
+                        uri = getUriFromAssetId(context, uriString)
+                        if (uri == null) {
+                            return source
+                        }
                     }
+                    source.uriString = uriString
+                    source.uri = uri
                 }
-                source.uriString = uriString
-                source.uri = uri
```
moskalakamil, line 235: too much nested ifs, can we change to something like this, since we already edited this?
```kt
    if (src == null) return source

    safeGetString(src, PROP_SRC_URI, null)
        ?.takeIf { it.isNotBlank() }
        ?.let { uriString ->
            var uri = Uri.parse(uriString)

            if (!isValidScheme(uri.scheme)) {
                uri = getUriFromAssetId(context, uriString) ?: return source
            }

            source.uriString = uriString
            source.uri = uri
        }
```

android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
```diff
@@ -2732,10 +2755,188 @@ public void onAdError(AdErrorEvent adErrorEvent) {
...
+     * and switches to the backup stream.
+     *
+     * @return true if backup stream was successfully used, false otherwise
+     */
+    private boolean handleDaiBackupStream() {
+        if (source == null || source.getAdsProps() == null) {
+            return false;
+        }
+        
+        String backupStreamUri = source.getAdsProps().getFallbackUri();
```
moskalakamil: let's keep the same name- let's not mix up the names (fallback|backup)

examples/expo-dai/App.tsx
```diff
@@ -0,0 +1,483 @@
+import React, {useState, useCallback, useRef} from 'react';
```
moskalakamil: we should be able to add a new video in examples/common/constans/general.ts and have only one example

src/types/video.ts
```diff
@@ -77,11 +77,44 @@ export enum DRMType {
...
+  }>;
+
+export type AdConfigDAILive = AdConfigDAIBase &
+  Readonly<{
+    assetKey: string;
+    contentSourceId?: never;
+    videoId?: never;
+  }>;
+
+export type AdConfigDAI = AdConfigDAIVod | AdConfigDAILive;
```
moskalakamil, line 115: We should add a new field, such as `streamType`, and make the arguments mandatory when they are actually required

src/types/video.ts
```diff
@@ -77,11 +77,44 @@ export enum DRMType {
...
-  adTagUrl?: string;
+export type DaiFormat = 'hls' | 'dash';
+
+type AdConfigBase = Readonly<{
   adLanguage?: ISO639_1;
 }>;
 
+export type AdConfigCSAI = AdConfigBase &
+  Readonly<{
+    type: 'csai';
```
moskalakamil, line 89: We definitely cannot use type: "csai" | "dai", but there is a bigger issue here. In this case, we should use either "ima" and "dai" or "csai" and "ssai". The first option is better and easier to maintain in the future, assuming that new ad delivery providers may be added. I also believe we can find a more suitable property name than type

moskalakamil: We should discuss this internally to agree on the most appropriate naming. If anyone has suggestions, feel free to share them

src/types/video.ts
moskalakamil: then we also won't need values 'isDaiVOD' and 'isDaiLive' in native code

moskalakamil (changes requested):

examples/expo-dai/App.tsx
moskalakamil: then we can remove additional example

src/Video.tsx
```diff
@@ -53,7 +53,54 @@ import type {
...
+        type: 'dai',
+        adLanguage: ad.adLanguage,
+        contentSourceId:
+          'contentSourceId' in ad ? ad.contentSourceId : undefined,
+        videoId: 'videoId' in ad ? ad.videoId : undefined,
+        assetKey: 'assetKey' in ad ? ad.assetKey : undefined,
+        format: 'format' in ad ? ad.format : undefined,
+        adTagParameters:
+          'adTagParameters' in ad ? ad.adTagParameters : undefined,
+        fallbackUri: 'fallbackUri' in ad ? ad.fallbackUri : undefined,
```
moskalakamil: we don't need this check here- if we don't have this param, it will be undefined

docs/pages/component/ads.md
```diff
@@ -23,21 +45,156 @@ To receive events from the IMA SDK, pass the `onReceiveAdEvent` prop to the `Vid
...
+/>
+```
+
+---
+
+## Server-Side Ad Insertion (SSAI)
+
+SSAI (Server-Side Ad Insertion) is a server-side ad insertion solution where ads are stitched into the video stream before it reaches the player. This provides a seamless viewing experience with no playback interruptions, as the stream appears as a single continuous video.
+
+SSAI is ideal for:
```
moskalakamil, line 83:
```suggestion
SSAI (Server-Side Ad Insertion) is a server-side ad insertion solution where ads are stitched into the video stream before it reaches the player. This provides a seamless viewing experience with no playback interruptions, as the stream appears as a single continuous video.

Currently, we support **Google IMA DAI**

SSAI is ideal for:
```

ios/Video/Features/RCTIMAAdsManager.swift
```diff
@@ -28,6 +34,14 @@
             adsLoader.delegate = self
         }
 
+        func setupDaiLoader() {
+            let settings = IMASettings()
+            // Enable background playback for Picture-in-Picture support
+            settings.enableBackgroundPlayback = true
```
moskalakamil: shouldn’t we set this only when playInBackground is true, not every time?

moskalakamil (approved):
left 3 comments, overall lgtm 🚀 thanks again guys!

docs/pages/component/ads.md
moskalakamil, line 1: The v6 documentation is on the master (v7) branch. We’ll need to update it there later as well

PR #4828, opened 2026-01-26
fix(ios): prevent KVO crash in HybridVideoPlayer.release()

erank-pixel:
## Summary

This PR fixes a crash that occurs when rapidly creating/destroying video players (e.g., fast swiping through a FlatList of videos).

**Crash message:**
```
Cannot remove an observer for the key path 'currentItem.status'
```

## Root Cause

When `release()` was called, it would:
1. Set `playerItem = nil`
2. Set `playerObserver = nil` (triggering the observer's deinit)

However, the KVO observers were still registered on the `playerItem`, causing a crash when the system tried to remove them during deallocation.

## Solution

1. **Explicitly invalidate all KVO observers BEFORE changing any state** - Call `invalidatePlayerItemObservers()` and `invalidatePlayerObservers()` at the start of `release()`
2. **Change `invalidatePlayerItemObservers()` from private to internal** - So it can be called from `HybridVideoPlayer.release()`
3. **Skip `self.player.replaceCurrentItem(with: nil)`** - This was also triggering the crash by changing `currentItem` while observers were still active

This ensures clean KVO removal before any state changes that could interfere with the observer lifecycle.

## Test Plan

- [x] Tested with rapid swiping through a FlatList containing multiple video players
- [x] Verified no crashes when quickly navigating between screens with videos
- [x] Confirmed video playback still works correctly after the changes

## Related Issues

This fixes crashes reported when using video players in scrollable lists with rapid scrolling.

### DIFF
--- a/packages/react-native-video/ios/core/VideoPlayerObserver.swift
+++ b/packages/react-native-video/ios/core/VideoPlayerObserver.swift
@@ -162,7 +162,9 @@ class VideoPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate, AVP
     metadataOutput.setDelegate(self, queue: .global(qos: .userInteractive))
   }
   
-  private func invalidatePlayerItemObservers() {
+  // KVO CRASH FIX: Changed from private to internal so HybridVideoPlayer.release()
+  // can explicitly invalidate these observers BEFORE changing playerItem.
+  func invalidatePlayerItemObservers() {
     // Remove NotificationCenter observers
     if let playbackEndedObserver = playbackEndedObserver {
       NotificationCenter.default.removeObserver(playbackEndedObserver)
--- a/packages/react-native-video/ios/core/VideoPlayerObserver.swift
+++ b/packages/react-native-video/ios/core/VideoPlayerObserver.swift
@@ -162,8 +162,6 @@ class VideoPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate, AVP
     metadataOutput.setDelegate(self, queue: .global(qos: .userInteractive))
   }
   
-  // KVO CRASH FIX: Changed from private to internal so HybridVideoPlayer.release()
-  // can explicitly invalidate these observers BEFORE changing playerItem.
   func invalidatePlayerItemObservers() {
     // Remove NotificationCenter observers
     if let playbackEndedObserver = playbackEndedObserver {
--- a/packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
+++ b/packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
@@ -218,8 +218,9 @@ class HybridVideoPlayer: HybridVideoPlayerSpec, NativeVideoPlayerSpec {
     // Clear player observer
     playerObserver?.invalidatePlayerItemObservers()
     playerObserver?.invalidatePlayerObservers()
-    self.player.replaceCurrentItem(with: nil)
     self.playerObserver = nil
+
+    self.player.replaceCurrentItem(with: nil)
     status = .idle
 
     VideoManager.shared.unregister(player: self)
--- a/packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
+++ b/packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
@@ -206,17 +206,24 @@ class HybridVideoPlayer: HybridVideoPlayerSpec, NativeVideoPlayerSpec {
   func release() {
     sourceLoader.cancelSync()
     NowPlayingInfoCenterManager.shared.removePlayer(player: player)
-    
+
     try? _eventEmitter?.clearAllListeners()
-    
-    self.player.replaceCurrentItem(with: nil)
+
+    // KVO CRASH FIX: Explicitly invalidate all observers BEFORE changing any state.
+    // The crash "Cannot remove an observer for the key path 'currentItem.status'"
+    // happens when playerItem/playerObserver changes while KVO observers are still active.
+    // By invalidating observers first, we ensure clean KVO removal before any state changes.
+    playerObserver?.invalidatePlayerItemObservers()
+    playerObserver?.invalidatePlayerObservers()
+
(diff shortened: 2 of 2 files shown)


packages/react-native-video/ios/core/VideoPlayerObserver.swift
```diff
@@ -162,7 +162,9 @@ class VideoPlayerObserver: NSObject, AVPlayerItemMetadataOutputPushDelegate, AVP
     metadataOutput.setDelegate(self, queue: .global(qos: .userInteractive))
   }
   
-  private func invalidatePlayerItemObservers() {
+  // KVO CRASH FIX: Changed from private to internal so HybridVideoPlayer.release()
+  // can explicitly invalidate these observers BEFORE changing playerItem.
```
moskalakamil:
```suggestion
```

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
```diff
@@ -206,17 +206,24 @@ class HybridVideoPlayer: HybridVideoPlayerSpec, NativeVideoPlayerSpec {
...
-    
+
     try? _eventEmitter?.clearAllListeners()
-    
-    self.player.replaceCurrentItem(with: nil)
+
+    // KVO CRASH FIX: Explicitly invalidate all observers BEFORE changing any state.
+    // The crash "Cannot remove an observer for the key path 'currentItem.status'"
+    // happens when playerItem/playerObserver changes while KVO observers are still active.
+    // By invalidating observers first, we ensure clean KVO removal before any state changes.
```
moskalakamil:
```suggestion
```

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
```diff
@@ -206,17 +206,24 @@ class HybridVideoPlayer: HybridVideoPlayerSpec, NativeVideoPlayerSpec {
   func release() {
     sourceLoader.cancelSync()
     NowPlayingInfoCenterManager.shared.removePlayer(player: player)
-    
+
     try? _eventEmitter?.clearAllListeners()
-    
-    self.player.replaceCurrentItem(with: nil)
```
moskalakamil, line 212: why you removed this? Is it connected with issue?

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
```diff
@@ -206,17 +206,24 @@ class HybridVideoPlayer: HybridVideoPlayerSpec, NativeVideoPlayerSpec {
...
+
+    // KVO CRASH FIX: Explicitly invalidate all observers BEFORE changing any state.
+    // The crash "Cannot remove an observer for the key path 'currentItem.status'"
+    // happens when playerItem/playerObserver changes while KVO observers are still active.
+    // By invalidating observers first, we ensure clean KVO removal before any state changes.
+    playerObserver?.invalidatePlayerItemObservers()
+    playerObserver?.invalidatePlayerObservers()
```
moskalakamil: we should move this under `Clear player observer` section

moskalakamil:
hey @erank-pixel, thanks for the PR! left some comments

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
moskalakamil: we can keep it as it was, but call this after cleaning up the observers

erank-pixel:
Thanks for the review @moskalakamil! I'll address all your feedback:

1. **Remove verbose comments** - Agreed, I'll clean up the lengthy 'KVO CRASH FIX' comments in both files.

2. **Restore `replaceCurrentItem(with: nil)`** - Good point! I initially removed it because it was also triggering the KVO crash when changing `currentItem` while observers were still registered. However, you're right - if we clean up the observers first, we can safely call it afterward. I'll restore it and move it after the observer cleanup.

3. **Move invalidation calls** - I'll move the invalidation calls under the 'Clear player observer' section for better code organization.

I'll push an updated commit with these changes.

packages/react-native-video/ios/core/VideoPlayerObserver.swift
erank-pixel: Thanks! I'll remove these verbose comments.

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
erank-pixel: Will do!

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
erank-pixel: Makes sense, I'll move these under the 'Clear player observer' section.

packages/react-native-video/ios/hybrids/VideoPlayer/HybridVideoPlayer.swift
erank-pixel: Yes, it was connected to the issue. Calling `replaceCurrentItem(with: nil)` was triggering the same KVO crash because it changes `currentItem` while the KVO observers were still active, causing the 'Cannot remove an observer for the key path currentItem.status' error.

But as you suggested, I can restore it and call it AFTER cleaning up the observers. Will update the PR!

erank-pixel:
I've pushed a new commit addressing all the feedback:

- Removed the verbose 'KVO CRASH FIX' comments from both files
- Moved the observer invalidation calls under the 'Clear player observer' section
- Restored `self.player.replaceCurrentItem(with: nil)` - it's now called after the observers are cleaned up, which prevents the KVO crash

The `release()` function now has a cleaner structure:
1. Cancel source loader & remove from NowPlayingInfoCenter
2. Clear event listeners
3. Clear playerItem reference
4. Release source asset
5. **Clear player observer section**: invalidate observers → replaceCurrentItem → nil out observer
6. Reset status and unregister

Let me know if you'd like any other changes!

PR #4724, opened 2025-10-07
feat: web support

zoriya:
## Summary

Initial implementation for the web. I decided to use video.js to support more advanced features (like hls/dash/ads/drm...)

Missing features i don't plan on implementing:
- buffer config
- drm
- ads

### Motivation

Fixes #4605 

Note that i found some missing APIs while implementing this (crossed items are those I added an api & implementation but only on the web):
- ~missing `mimeType` in source~ 
- missing `startTime`
- ~missing audio/video selections~
- missing metadata in text tracks (default, forced, cc, codecs)
- multi-source setup (give two uris + mimetype for the same video & let the player determine which one it can play)
- ~multi audios~
- ~multi variants (hls/dash only)~
- next/prev custom actions for the media notification

## Test plan

I've been using this PR for a few months on https://github.com/zoriya/kyoo and it works perfectly. This should be ready to merge.

PR #4849, opened 2026-03-04
chore: Upgrade to Nitro 0.35.0

mrousavy:
Upgrades to [Nitro 0.35.0](https://github.com/mrousavy/nitro/releases/tag/v0.35.0), which contains a **BREAKING CHANGE** related to Kotlin HybridObjects (it fixes a critical memory leak, so it is advised everyone upgrades)

Steps I did (you can follow):
1. Upgrade react-native-nitro-modules and nitrogen to 0.35.0
2. Run bun install & pod install
3. Replace all usages of `bigint` with `Int64`/`UInt64` (in this repo; `duration` and `fileSize` *1)
4. Re-generate specs using `bun nitrogen`
5. (Optionally) replaced the `cpp-adapter.cpp` call from `initialize(vm)` with `registerAllNatives()` - this just allows you to register custom JNI natives in your libs, if needed.

Also, I removed the nitrogen and react-native-nitro-modules patches since I don't think they are necessary anymore. The boolean/number thing in Swift has been fixed upstream, and the exception handling is fixed in react-native core :)

- *1: Shouldn't `duration` be a `number` (double)? Right now, it can't have commas. It's documented as "The duration of the video in seconds.", and seconds definitely has commas.


KrzysztofMoch (approved):
>Also, I removed the nitrogen and react-native-nitro-modules patches since I don't think they are necessary anymore. The boolean/number thing in Swift has been fixed upstream, and the exception handling is fixed in react-native core :)

I agree. latest React Native is already 0.84, so it shoud be safe to remove that.

>*1: Shouldn't duration be a number (double)? Right now, it can't have commas. It's documented as "The duration of the video in seconds.", and seconds definitely has commas.

Yes, nice find — we need to change it to ms or leave as it is and change type to a number. 

cc @moskalakamil

mrousavy:
I think seconds is good, and double. I do the same in VisionCamera

moskalakamil (approved):
LGTM! thanks a lot @mrousavy @KrzysztofMoch and @fnwk. last thing - docs, could you @fnwk please update requirements?

SGamrekelashvili:
Could you release package with this commit included? 
Thanks.

zoriya:
Building with this PR causes a build error on android:

`e: file:///home/expo/workingdir/build/front/node_modules/react-native-video/android/src/main/java/com/twg/video/core/utils/VideoInformationUtils.kt:58:18 Argument type mismatch: actual type is 'Long', but 'Double' was expected.`

This can be fixed by da91f4fba6afd9bc8331da146e3e6bd861831214 (i pushed it in my branch #4724)

PR #4871, opened 2026-04-01
fix(android/v6): implement native viewType switching in ExoPlayerView

saseungmin:
## Summary

This fixes Android `viewType` handling in `ExoPlayerView` for the v6 branch.

`react-native-video` already exposes `viewType` / `useTextureView` on the JS side, and the Android manager forwards that prop to `ExoPlayerView`. 
However, `ExoPlayerView.updateSurfaceView()` is currently a no-op, so requesting `ViewType.TEXTURE` does not actually switch the underlying native view type.

This becomes visible in layouts with multiple adjacent videos using `resizeMode="cover"` on Android, where the video can render outside of its clipped bounds because the default `SurfaceView` path is still used.

## What changed

- implement `updateSurfaceView(viewType)` in `ExoPlayerView`
- create a `PlayerView` backed by `texture_view` when `ViewType.TEXTURE` is requested
- recreate the internal `PlayerView` when the native view type changes
- restore the relevant `PlayerView` state after recreation
  - player
  - resizeMode
  - shutter color
  - controller configuration
  - focusability
  - controller/fullscreen listeners
  - layout change listeners
  - subtitle style
- preserve `SURFACE_SECURE` behavior when the secure surface type is requested

## Why this is needed

The JS API already supports selecting the Android view type, but the native implementation in `ExoPlayerView` does not currently apply that choice.

With this change, `viewType={ViewType.TEXTURE}` and the deprecated `useTextureView` path can correctly use a `TextureView`-backed `PlayerView`, which makes clipping behavior work as expected in Android UI layouts that rely on view bounds.

## Scope

This PR intentionally keeps the change limited to the Android `viewType` implementation in `ExoPlayerView`.

It does not include app-specific behavior changes or unrelated refactors.

## Related issues

- https://github.com/TheWidlarzGroup/react-native-video/issues/4635
- https://github.com/TheWidlarzGroup/react-native-video/issues/4202
- https://github.com/TheWidlarzGroup/react-native-video/issues/4809
- https://github.com/TheWidlarzGroup/react-native-video/pull/4031

## Notes

`PlayerView` does not expose a runtime setter for `surface_type`, so switching between `surface_view` and `texture_view` requires recreating the internal `PlayerView` instance and restoring its state.

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt
@@ -4,6 +4,8 @@ import android.content.Context
 import android.graphics.Color
 import android.graphics.drawable.GradientDrawable
 import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.view.SurfaceView
 import android.view.View
 import android.view.View.MeasureSpec
 import android.widget.FrameLayout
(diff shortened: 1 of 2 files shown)


Osilos:
Thanks for this, we need it ! 👍

PR #4903, opened 2026-05-20
fix(android): add null check for player in videoLoaded() to prevent NPE

### LINKED ISSUE #4902: [BUG][Android]: NullPointerException in ReactExoplayerView.videoLoaded() — player.isPlayingAd() called without null-check
### What happened?

Fatal `NullPointerException` thrown from the main looper:

```
NullPointerException: Attempt to invoke interface method
'boolean androidx.media3.exoplayer.ExoPlayer.isPlayingAd()' on a null object reference
  at com.brentvatne.exoplayer.ReactExoplayerView.videoLoaded(ReactExoplayerView.java:1486)
  at com.brentvatne.exoplayer.ReactExoplayerView.onEvents(ReactExoplayerView.java:1443)
  at androidx.media3.exoplayer.ExoPlayerImpl.lambda$new$0(ExoPlayerImpl.java:301)
  at androidx.media3.common.util.ListenerSet$ListenerHolder.iterationFinished(ListenerSet.java:353)
  at androidx.media3.common.util.ListenerSet.handleMessage(ListenerSet.java:297)
  at android.os.Handler.dispatchMessage(Handler.java:106)
  at android.os.Looper.loopOnce(...)
```

**Root cause**

`videoLoaded()` (line 1486 on `support/6.x.x`) calls `player.isPlayingAd()` **directly** on the field, with no null check:

```java
// android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
private void videoLoaded() {
    if (!player.isPlayingAd() && loadVideoStarted) {   // ← NPE here
        ...
    }
}
```

This is a race condition between `ListenerSet` dispatch and the player being released:

1. ExoPlayer's `ListenerSet` posts a `Player.Listener.onEvents` callback on the main looper.
2. Before the message is dispatched, the React view is unmounted → `cleanUpResources()` / `releasePlayer()` runs and nulls `player`.
3. The pending message fires → `onPlaybackStateChanged(STATE_READY)` → `videoLoaded()` → `player.isPlayingAd()` on a null reference.

This happens in feed-style UIs (vertical/horizontal lists of videos remounted on scroll, or apps that key-remount `<Video>` on source change) where the unmount window overlaps with `STATE_READY` events.

**Already-existing helper that should be used**

A private null-safe wrapper already exists in the same file (`support/6.x.x` line 345):

```java
private boolean isPlayingAd() {
    return player != null && player.isPlayingAd();
}
```

…and is used elsewhere in the file (e.g. line 285). The call site at line 1486 was simply missed.

**Suggested fix**

Replace the raw field access at line 1486 with the existing helper:

```diff
 private void videoLoaded() {
-    if (!player.isPlayingAd() && loadVideoStarted) {
+    if (player == null) return;
+    if (!isPlayingAd() && loadVideoStarted) {
```

(The early `player == null` guard also protects the `player.getVideoFormat()` call a few lines below, which would NPE under the same race.)

**Impact**

In our production app: **392 crashes / 262 unique users** over 6 months on a feed of looping background videos that key-remount on source change. Crash rate scales with how aggressively the app destroys/creates Video views.

**Reproducer**

Hard to nail down a deterministic repro because of the race, but it is easy to surface in a FlatList of `<Video>` items where the `key` prop changes on scroll (forces native release+create). The Sentry replays we have show the crash consistently happening right at the moment a video tile is swapped out.

**Expected Behavior**

`videoLoaded()` should be a no-op (or use the existing null-safe `isPlayingAd()` helper) when the player has already been released.

### Steps to reproduce

1. Render a feed-style list of `<Video>` components that mount/unmount on scroll (or key-remount on source change).
2. Scroll so a video is unmounted right as it reaches `STATE_READY`.
3. Observe the `NullPointerException` from `videoLoaded()` -> `player.isPlayingAd()` on the main looper.

### Reproduction repository

_No response_

6.19.2 (also affects 6.19.1; same code on `support/6.x.x`)

### react-native version

_No response_

_No response_

### Platforms

Android

### OS version

Reproduced on Android 13 → Android 16. Most reports are Samsung devices but the bug is platform-wide; observed across 260+ unique users in production.

### Device

Real device (Samsung SM-S936B / Galaxy S24 among many others)

### Architecture

Old Architecture

### Expo

_No response_

### Last working version

_No response_

### Media / source type

_No response_

vanhci:
## Fix Summary
Add null check for  in  method to prevent NullPointerException.

**Bug**: 

**Root cause**: Race condition where ExoPlayer listener posts callback on main looper, before it's dispatched the player is released (nulled), then the callback fires and causes NPE when calling .

**Fix**: Changed:

To:

Fixes #4902

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
+++ b/android/src/main/java/com/brentvatne/exoplayer/ReactExoplayerView.java
@@ -1483,7 +1483,7 @@ public class ReactExoplayerView extends FrameLayout implements
     }
 
     private void videoLoaded() {
-        if (!player.isPlayingAd() && loadVideoStarted) {
+        if (player != null && !player.isPlayingAd() && loadVideoStarted) {
             loadVideoStarted = false;
             if (audioTrackType != null) {
                 setSelectedAudioTrack(audioTrackType, audioTrackValue);


Osilos:
@vanhci you push a PR very quickly to fix this issue a just open Thanks ! 
How did you do to do it so fast ?

Issue #4614, opened 2025-07-17
[BUG]: Upgrading from 6.14.1 to 6.16.1 suddenly stops textTrack from loading on Android only


jung35:
### What happened?

After upgrading react-native-video from `6.14.1` to `6.16.1`, text tracks (subtitles) stop loading on Android only. iOS (simulator and physical) works correctly, as does the `6.14.1` version on all platforms.

Text tracks are passed via the `source` prop. On Android (both physical device and emulator):

- `onTextTracks={({ textTracks }) => {}}` returns an empty array `[]`.
- `onTextTrackDataChanged={({ subtitleTracks }) => {}}` is no longer called when expected (it should be called every time a text track emits a new set of text based on the time).

Downgrading back to `6.14.1` makes the subtitles display correctly again.

**Source / selected track config used:**

```jsx
source={{
  uri: "...",
  textTracks: [{
    title: "English",
    language: "en",
    type: TextTrackType.VTT,
    uri: "..."
  }]
}}
selectedTextTrack={{ type: SelectedTrackType.LANGUAGE, value: "en" }}
```

**Workaround:** Stay on `6.14.1`, where subtitles display fine.

### Steps to reproduce

1. Set up a `Video` component passing `textTracks` in the `source` prop and a `selectedTextTrack` (see config above).
2. Add `onTextTracks` and `onTextTrackDataChanged` handlers.
3. Upgrade react-native-video to `6.16.1` and run on an Android device or emulator.
4. Observe that `onTextTracks` returns `[]` and `onTextTrackDataChanged` is never called, so subtitles do not display.

### react-native-video version

6.16.1

### Platforms

Android

### OS version

Android 15

### Device

Real device

### Architecture

New Architecture

### Last working version

6.14.1

### Media / source type

Other

fireinnti:
I second this issue, issue was resolved moving back to 6.14.1

alexduraibi:
when source have `textTracks`, `OnLoadStart` event fire multiple times.

**Issue only on Android**. downgraded to `6.14.1` still same

```
source={{
  uri: "...",
  textTracks: [{
    title: "English",
    language: "en",
    type: TextTrackType.VTT
    uri: "..."
  }]
}}
selectedTextTrack={{ type: SelectedTrackType.LANGUAGE, value: "en" }}

```


logs
```
VPLAYER-handlePlay
 (NOBRIDGE) LOG  VPLAYER-handleProgress 4.061
[...]

PR #4906, opened 2026-05-20
fix(android): merge sideloaded subtitle tracks via MergingMediaSource; guard PiP activity lookup

### LINKED ISSUE #4792: [BUG]: External Subtitles Not Working in Latest Versions (`onTextTracks` returns null)
### What happened?

A breaking regression affecting all versions **after 6.15.0**: external subtitles (VTT/SRT) do not load.

**Working version:** `react-native-video@6.15.0`

**Broken versions (tested):** `6.16.0`, `6.17.0`, `6.18.0` (latest)

#### Issue Description

External subtitles (VTT/SRT) **do not load** in the latest versions.

* The video plays normally, but subtitles never display.
* `onTextTracks()` returns `null` or an empty array.
* No subtitle item appears for selection.
* Selecting subtitle manually also does nothing.

The **same code works perfectly** on version **6.15.0**, so this is a regression.

#### Code Example (Working only on 6.15.0)

```js
<Video
  useExoPlayer={true}
  source={{ uri: videoUrl }}
  textTracks={[
    {
      title: "English",
      language: "en",
      type: TextTrackType.VTT,
      uri: subtitleUrl,
    },
  ]}
  selectedTextTrack={{
    type: "index",
    value: 0,
  }}
  onTextTracks={(tracks) => {
    console.log("Subtitle tracks:", tracks);
  }}
/>
```

#### Current Behavior (Bug)

* `onTextTracks` → returns: `null` or `[]`
* Subtitles never appear
* Cannot select subs or display them at runtime

#### Expected Behavior

* `textTracks` should load the external subtitle file
* `onTextTracks` should return available subtitle tracks
* Selected subtitle should be displayed during playback (like in 6.15.0)

#### Notes

This affects:

* Android (**ExoPlayer path**)
* External subtitles (not embedded)
* `.vtt`, `.srt`, `.vtt` remote URLs

#### Impact

This breaks accessibility and multilingual playback support. Many streaming apps rely on subtitles — so this is a major regression.

#### Request

Please fix this subtitle regression in the latest version or provide guidance / patch. Available to test builds and provide logs if needed.

### Steps to reproduce

1. Use `react-native-video` 6.16.0 or later on Android (ExoPlayer path).
2. Render a `<Video>` component with an external subtitle file passed via the `textTracks` prop (e.g. a remote `.vtt`/`.srt` URI) and `selectedTextTrack` set to index 0.
3. Play the video and observe `onTextTracks` returns `null` or `[]`; no subtitles display and none can be selected.
4. Downgrade to 6.15.0 with the same code — subtitles load and display correctly.

### Reproduction repository

_No response_

6.18.0

### react-native version

_No response_

_No response_

### Platforms

Android

### OS version

Android 12

### Device

Real device

### Architecture

Old Architecture

### Expo

_No response_

### Last working version

6.15.0

### Media / source type

Other

WithakErik:
## Problem

External subtitle/caption files passed via `textTracks` never appeared on Android. Two related bugs:

**Bug 1 — Subtitles silently ignored**
`ProgressiveMediaSource.Factory` (and the HLS/DASH variants) ignore `MediaItem.SubtitleConfiguration` entries entirely. Subtitles set via `setSubtitleConfigurations` are attached to the `MediaItem` but never loaded into `MappedTrackInfo`, so `setSelectedTextTrack` has nothing to select.

Reported in #4792 and #4614.

**Bug 2 — PiP crash after subtitle change**
`findActivity()` threw `IllegalStateException` when the `ComponentActivity` wasn't reachable through the `ContextWrapper` chain. This surface became more likely after the subtitle fix changes ExoPlayer's internal session/lifecycle state via `MergingMediaSource`.

## Fix

**Subtitle fix (`ReactExoplayerView.java`)**

After the primary `MediaSource` is built, check for `SubtitleConfiguration` entries on the `MediaItem`. For each one, create a dedicated `ProgressiveMediaSource` backed by a `SubtitleExtractor`, then merge all sources with `MergingMediaSource`. This makes sideloaded tracks appear in `MappedTrackInfo` and respond to `setSelectedTextTrack` exactly like embedded tracks.

```java
if (mediaItem.localConfiguration != null && !mediaItem.localConfiguration.subtitleConfigurations.isEmpty()) {
    // build one ProgressiveMediaSource per subtitle, merge with video source
    mediaSource = new MergingMediaSource(mergedSources);
}
```

**PiP fix (`PictureInPictureUtil.kt`)**

- `findActivity()` now returns `ComponentActivity?` instead of throwing, with a `ThemedReactContext.getCurrentActivity()` fallback
- All call sites updated to handle null with early returns
- Replaces deprecated `context.currentActivity` field with `context.getCurrentActivity()`

## Testing

Verified on a physical Android device (arm64-v8a) with external `.vtt` subtitle files. Subtitles now load and are selectable. PiP entry/exit no longer crashes after subtitles are active.

Closes #4792
Related: #4614

### DIFF
--- a/android/src/main/java/com/brentvatne/exoplayer/PictureInPictureUtil.kt
+++ b/android/src/main/java/com/brentvatne/exoplayer/PictureInPictureUtil.kt
@@ -24,13 +24,18 @@ import com.brentvatne.common.toolbox.DebugLog
 import com.brentvatne.receiver.PictureInPictureReceiver
 import com.facebook.react.uimanager.ThemedReactContext
 
-internal fun Context.findActivity(): ComponentActivity {
+internal fun Context.findActivity(): ComponentActivity? {
     var context = this
     while (context is ContextWrapper) {
         if (context is ComponentActivity) return context
         context = context.baseContext
     }
-    throw IllegalStateException("Picture in picture should be called in the context of an Activity")
+    // Fallback: ThemedReactContext stores the activity separately from the ContextWrapper chain
+    if (this is ThemedReactContext) {
+        val activity = this.getCurrentActivity()
+        if (activity is ComponentActivity) return activity
+    }
+    return null
 }
 
 object PictureInPictureUtil {
@@ -39,7 +44,7 @@ object PictureInPictureUtil {
 
     @JvmStatic
     fun addLifecycleEventListener(context: ThemedReactContext, view: ReactExoplayerView): Runnable {
-        val activity = context.findActivity()
+        val activity = context.findActivity() ?: return Runnable {}
 
         val onPictureInPictureModeChanged = Consumer<PictureInPictureModeChangedInfo> { info ->
             view.setIsInPictureInPicture(info.isInPictureInPictureMode)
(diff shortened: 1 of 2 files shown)


WithakErik:
Updated with three hardening improvements (based on additional code review):

- **Strip subtitle configs from base MediaItem** before passing to `mediaSourceFactory` — prevents any factory variant from also creating legacy subtitle sources alongside our `MergingMediaSource`, which would trigger the Media3 1.8 *Legacy decoding is disabled* crash
- **Guard unsupported formats** via `subtitleParserFactory.supportsFormat()` before calling `create()` — unknown MIME types previously threw `IllegalArgumentException` at source creation time, killing playback; now we log a warning and skip the track, matching `DefaultMediaSourceFactory`'s behavior
- **Preserve `selectionFlags` and `roleFlags`** on the rebuilt `Format` so default/forced-caption auto-selection continues to work correctly

fryingpanuu:
I am begging you guys please merge this PR if it fixes the issues. It's been a year since this issue is open. There are actual human beings who use Android, not the whole world runs on ios

PR #4908, opened 2026-05-31
Fix iOS player view controller containment

### LINKED ISSUE #3138: Show video controls by default when paused
# Feature Request

I have a video that I want to start paused. This works, however, the user must tap on the paused thumbnail in order to see the controls. The controls should show by default, not require the user to tap on the still image.

sorinc03:
## Summary
- Call `didMove(toParent:)` after embedding the iOS `AVPlayerViewController` for native controls.
- Pair controller removals with `willMove(toParent: nil)` before detaching/removing the controller.
- Backports the proper containment lifecycle pattern already used on the v7/master implementation to the 6.x support branch.

Fixes #3138

## Test Plan
- `git diff --check`
- `xcodebuild -list -project ios/RCTVideo.xcodeproj`
- `xcodebuild -project ios/RCTVideo.xcodeproj -scheme RCTVideo -configuration Debug -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO build` *(fails in this standalone clone because React and generated bridge headers are unavailable: `unable to resolve module dependency: React` and `RCTVideoSwiftLog.h file not found`)*

## Notes
This keeps the change scoped to UIKit containment lifecycle. `AVPlayerViewController` relies on being correctly added as a child view controller, and the 6.x branch was missing the `didMove(toParent:)` call after `addChild`.

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -1109,7 +1109,8 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
                 if let playerViewController = _playerViewController {
                     if _controls {
                         // prevents crash https://github.com/TheWidlarzGroup/react-native-video/issues/3040
-                        self._playerViewController?.removeFromParent()
+                        playerViewController.willMove(toParent: nil)
+                        playerViewController.removeFromParent()
                     }
 
                     viewController.present(playerViewController, animated: true, completion: { [weak self] in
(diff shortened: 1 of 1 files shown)


Duya3fithou:
Thanks for your contribution. Is it possible to support Video v7?

sorinc03:
This PR is targeted at `support/6.x.x`, so I kept the change scoped to v6. If v7 has the same containment issue, I can prepare a separate v7-targeted PR/cherry-pick against the appropriate branch rather than mixing both release lines here.

PR #5035, opened 2026-08-07
fix(ios): report observed bitrate for HLS

### LINKED ISSUE #4153: [BUG]: onBandwidthUpdate does not work on HLS iOS.
### What happened?

`onBandwidthUpdate` does not work on HLS (`.m3u8`) sources on iOS. It does work as expected on MP4 files.

### Steps to reproduce

1. Set up a `<Video>` component on iOS with an `onBandwidthUpdate` callback handler.
2. Use an HLS `.m3u8` source, e.g. `https://rbmn-live.akamaized.net/hls/live/590964/BoRB-AT/master_928.m3u8`.
3. Play the stream and observe that `onBandwidthUpdate` is never called.
4. Switch the source to an MP4 file and observe that `onBandwidthUpdate` fires correctly.

### Reproduction repository

_No response_

6.5.0

### react-native version

_No response_

_No response_

### Platforms

iOS

### OS version

17.4

### Device

Simulator / Emulator

### Architecture

Old Architecture

### Expo

_No response_

### Last working version

_No response_

### Media / source type

HLS (.m3u8)

LizunovSergey:
## Summary

Closes #4153.

### Motivation

`AVPlayerItemAccessLogEvent.indicatedBitrate` can be negative for HLS streams even when `observedBitrate` contains a valid measured throughput. The existing implementation emits the negative indicated value once and then suppresses later updates while that value remains unchanged.

### Changes

- preserve `indicatedBitrate` when AVPlayer provides a valid value
- fall back to `observedBitrate` when the indicated value is unknown
- skip the callback when both values are unknown
- initialize the cached bitrate with a sentinel below the valid range

## Test plan

- `swiftformat --lint ios/Video/RCTVideo.swift`
- `swiftlint lint --no-cache ios/Video/RCTVideo.swift`
- direct AVPlayer access-log probe on an iPhone 16 Pro simulator (iOS 18.6)
  - issue stream: `indicated=-1`, `observed=16351824`, selected value `16351824`
  - Apple HLS control stream: three access-log events received and valid indicated values preserved

### DIFF
--- a/ios/Video/RCTVideo.swift
+++ b/ios/Video/RCTVideo.swift
@@ -61,8 +61,8 @@ class RCTVideo: UIView, RCTVideoPlayerViewControllerDelegate, RCTPlayerObserverH
     private var _startPosition: Float64 = -1
     var _disableAudioSessionManagement: Bool = false
     var _showNotificationControls = false
-    // Buffer last bitrate value received. Initialized to -2 to ensure -1 (sometimes reported by AVPlayer) is not missed
-    private var _lastBitrate = -2.0
+    // Buffer the last valid bitrate value received
+    private var _lastBitrate = -1.0
     private var _enterPictureInPictureOnLeave = false {
         didSet {
             if isPictureInPictureActive() { return }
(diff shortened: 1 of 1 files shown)


LizunovSergey:
CI note: all three `Build iOS` jobs fail before compilation because the workflow requests `platform=iOS Simulator,name=iPhone 16`, while the current runner image has no matching iPhone 16 simulator. Each job exits with code 70 during destination resolution.

The patch-specific checks pass (`Swift-Format` and `Swift-Lint`). I also verified the behavior directly on an iPhone simulator using the HLS URL from #4153: AVPlayer reported `indicated=-1` and `observed=16351824`, and the new fallback selected `16351824`.

The three build failures therefore appear unrelated to this one-file change.

LizunovSergey:
The three red iOS checks aren't from this PR. `Build iOS` pins `-destination 'platform=iOS Simulator,name=iPhone 16'`, and the runner image no longer provides that device, so xcodebuild stops with `Unable to find a device matching the provided destination specifier` before compiling anything.

The same job is red on every run in the repo — the last twelve, including your own `codex/*` branches — so there's nothing in this changeset to fix.

I'm happy to send a separate PR pinning the destination to a device the image actually has, if that would help.

PR #5040, opened 2026-08-14
fix(ios): support PhotoKit ph:// video sources

### LINKED ISSUE #5029: [BUG]: v7 iOS cannot play `ph://` PhotoKit assets (regression from v6)
### Version

`7.0.0-beta.10`

iOS (real device)

### Architecture

New Architecture / Nitro player API

### What happened?

Videos returned by the iOS photo library use `ph://...` PhotoKit identifiers. Passing one of those identifiers to `useVideoPlayer()` does not load or play the video. The native video view remains black and displays the crossed-out/unavailable play glyph.

This is a regression from v6, which explicitly detected `ph://` sources and resolved them through PhotoKit before creating the player item.

The current v7 implementation appears to treat every string URI as a directly playable URL:

- [`HybridVideoPlayerSource.init`](https://github.com/TheWidlarzGroup/react-native-video/blob/v7.0.0-beta.10/packages/react-native-video/ios/hybrids/VideoPlayerSource/HybridVideoPlayerSource.swift) creates `URL(string: uri)`.
- `initializeAsset()` then creates `AVURLAsset(url: url)`.
- There is no `PHAsset`/`PHImageManager` handling in the v7 iOS implementation.

A `ph://` value is a PhotoKit local identifier, not a file or network URL that `AVURLAsset` can open directly.

v6 had the required special case in `RCTVideo.swift`:

```swift
if let uri = source.uri, uri.starts(with: "ph://") {
  return Promise {
    RCTVideoUtils.preparePHAsset(uri: uri).then { asset in
      return self.playerItemPrepareText(
        asset: asset,
        assetOptions: nil,
        uri: source.uri ?? ""
      )
    }
  }
}
```

Related history:

- #1572 originally tracked adding support for the Camera Roll `ph://` format.
- #3611 tracked a later v6 regression and was fixed in `6.0.0-beta.7`.

### Reproduction

Obtain a video from `@react-native-camera-roll/camera-roll` and pass its URI to the v7 player:

```tsx
import { CameraRoll } from '@react-native-camera-roll/camera-roll'
import { useVideoPlayer, VideoView } from 'react-native-video'

// CameraRoll.saveAsset(...) and CameraRoll.getPhotos(...) both return
// an iOS video asset whose node.image.uri is "ph://...".

function PhotoLibraryVideo({ uri }: { uri: string }) {
  const player = useVideoPlayer({ uri }, (instance) => {
    instance.play()
  })

  return <VideoView player={player} style={{ width: 300, height: 500 }} />
}
```

Steps:

1. Record or select a video on iOS.
2. Save/query it with Camera Roll.
3. Confirm that `asset.node.image.uri` starts with `ph://`.
4. Pass that URI to `useVideoPlayer()` and render `VideoView`.
5. Playback fails and the unavailable-video glyph is shown.

A bundled `file://` video passed to the same player works.

### Expected behavior

v7 should retain v6's iOS PhotoKit support and resolve `ph://` identifiers using `PHAsset.fetchAssets(withLocalIdentifiers:)` plus `PHImageManager.requestAVAssetForVideo(...)` (with network access enabled for iCloud assets) before creating the player item.

Because PhotoKit may return an `AVAsset` that is not an `AVURLAsset` for edited/composed videos, the v7 source abstraction may need to hold `AVAsset` rather than only `AVURLAsset` for this source type.

fnwk:
## Summary

Restores support for playing iOS PhotoKit `ph://` video sources with the v7 `useVideoPlayer()` API.

## Motivation

Fixes #5029.

v7 created an `AVURLAsset` directly from every source URI. A PhotoKit local identifier is not a directly playable URL, so `ph://` sources displayed a black video view. v6 resolved these identifiers through PhotoKit before creating the player item.

## Changes

- Resolve `ph://` local identifiers through `PHAsset` and `PHImageManager.requestAVAsset`.
- Allow network access for videos stored in iCloud.
- Propagate task cancellation to the PhotoKit request.
- Support generic `AVAsset` results returned for edited or composed PhotoKit videos.
- Keep URL-only behavior explicitly guarded for DRM, HLS manifest injection, and file-size lookup.
- Report PhotoKit sources as local.
- Add the typed `source/photo-library-asset-not-found` error.

## Platforms affected

- [ ] Android
- [x] iOS
- [ ] visionOS
- [ ] tvOS / Android TV
- [ ] Windows
- [ ] Web

## Type of change

- [ ] Documentation only

## Test plan

- Built and installed the example app as a signed Release build on an iPhone 17 Pro running iOS 26.6.
- Selected a real video from the Photos library and passed its exact `ph://<localIdentifier>` value to `useVideoPlayer()`.
- Verified that the player reached `readyToPlay`, rendered a video frame, and advanced playback time without Metro.
- Reproduced the regression before the fix: a direct `AVURLAsset(ph://...)` was not playable.
- Ran a local PhotoKit integration harness with 16 passing tests, covering generic `AVAsset` handling, cancellation races, DRM guards, subtitles, missing assets, and case-insensitive schemes.
- Ran Swift parsing, TypeScript type checking, ESLint, and `git diff --check`.

## Checklist

- [x] I read the [contributing guidelines](https://github.com/TheWidlarzGroup/react-native-video/blob/master/CONTRIBUTING.md).
- [x] I updated the documentation / README where relevant. No update is required because this restores an existing source type and does not change usage.
- [x] If this changes how the library is used (API, props, events, behavior), I updated the AI agent skill (`skills/react-native-video/`) to match. No update is required because the public usage remains unchanged.
- [x] This PR is focused on a single concern.
- [x] I tested my changes on at least one platform.

### DIFF
--- a/packages/react-native-video/ios/core/Extensions/AVAsset+getAssetInformation.swift
+++ b/packages/react-native-video/ios/core/Extensions/AVAsset+getAssetInformation.swift
@@ -1,6 +1,6 @@
 import AVFoundation
 
-extension AVURLAsset {
+extension AVAsset {
   func getAssetInformation() async throws -> VideoInformation {
     // Default values
     var bitrate = Double.nan
@@ -11,7 +11,13 @@ extension AVURLAsset {
     var isLive = false
     var orientation: VideoOrientation = .unknown
 
-    let fileSize = try await VideoFileHelper.getFileSize(for: url)
+    let url = (self as? AVURLAsset)?.url
+    let fileSize: Int64
+    if let url {
+      fileSize = try await VideoFileHelper.getFileSize(for: url)
+    } else {
+      fileSize = -1
+    }
 
     // Check if asset is live stream
     if duration.flags.contains(.indefinite) {
@@ -34,7 +40,7 @@ extension AVURLAsset {
       if #available(iOS 14.0, tvOS 14.0, visionOS 1.0, *) {
         isHDR = videoTrack.hasMediaCharacteristic(.containsHDRVideo)
       }
-    } else if url.pathExtension == "m3u8" {
+    } else if let url, url.pathExtension.lowercased() == "m3u8" {
       // For HLS streams, we cannot get video track information directly
       // So we download manifest and try to extract video information from it
 
--- a/packages/react-native-video/ios/core/Extensions/AVPlayerItem+externalSubtitles.swift
+++ b/packages/react-native-video/ios/core/Extensions/AVPlayerItem+externalSubtitles.swift
@@ -9,13 +9,13 @@ import AVFoundation
 import Foundation
 
 extension AVPlayerItem {
-  static func withExternalSubtitles(for asset: AVURLAsset, config: NativeVideoConfig) async throws
+  static func withExternalSubtitles(for asset: AVAsset, config: NativeVideoConfig) async throws
     -> AVPlayerItem
   {
     if config.externalSubtitles?.isEmpty != false {
       return AVPlayerItem(asset: asset)
     }
-      
+
     let supportedExternalSubtitles = config.externalSubtitles?.filter { subtitle in
       ExternalSubtitlesUtils.isSubtitleTypeSupported(subtitle: subtitle)
     }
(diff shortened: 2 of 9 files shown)


moskalakamil (approved):

mrousavy:
fire

PR #5041, opened 2026-08-18
fix(types): accept bundled assets in Video source

### LINKED ISSUE #4901: [BUG] `source` TS type rejects React Native's `ImageRequireSource` (number) for bundled assets
### What happened?

Affects the TypeScript type-checker only — all runtime targets work correctly — when using manually declared typing for `*.mp4` files.

`react-native-video` accepts numeric Metro asset ids at **two** positions at runtime ([`Video.tsx:205-212`](https://github.com/TheWidlarzGroup/react-native-video/blob/v6.19.2/src/Video.tsx#L205-L212)):

```ts
const isLocalAssetFile =
  typeof _source === 'number' ||                            // legacy `source={require('./X.mp4')}` form
  ('uri' in _source && typeof _source.uri === 'number') ||  // wrapped `source={{ uri: require('./X.mp4') }}` form
  ('uri' in _source &&
    typeof _source.uri === 'string' &&
    (_source.uri.startsWith('file://') ||
     _source.uri.startsWith('content://') ||
     _source.uri.startsWith('.')));
```

Neither numeric shape is allowed by the TS type. `ReactVideoSource` is currently object-only with `string | NodeRequire` for `uri` ([video.ts](https://github.com/TheWidlarzGroup/react-native-video/blob/v6.19.2/src/types/video.ts)):

```ts
export type ReactVideoSource = Readonly<Omit<ReactVideoSourceProperties, 'uri'> & {
  uri?: string | NodeRequire;
}>;
```

A TS project that types *.mp4 requires using `ImageRequireSource` (since that is the runtime type returned by Metro), breaks usage of `react-native-video`. Even the [README's own example](https://github.com/TheWidlarzGroup/react-native-video/blob/v6.19.2/docs/pages/component/props.mdx#asset-loaded-via-require) fails to compile:

```ts
// Project-level asset declaration (RN-conventional):
declare module '*.mp4' {
  const url: ImageRequireSource; // resolves to `number`
  export = url;
}

const sintel = require('./sintel.mp4'); // ImageRequireSource (= number)
<Video source={sintel} />                // TS2322: number not assignable to ReactVideoSource (object type)
<Video source={{ uri: sintel }} />       // TS2322: number not assignable to string | NodeRequire
```

**Suggested fix**

Widen `ReactVideoSource` at both positions to match the runtime:

```ts
export type ReactVideoSource =
  | ImageRequireSource
  | Readonly<Omit<ReactVideoSourceProperties, 'uri'> & {
      uri?: string | ImageRequireSource;
    }>;
```

`ImageRequireSource` from `react-native` is the documented type produced by `require('./X.png')` / `require('./X.mp4')` etc. and is also what the runtime path expects. `NodeRequire` can stay in the `uri` union for backward compatibility, but as noted it's `@deprecated` in newer `@types/node`.

**Why this matters**

The README explicitly recommends `source={{ uri: require('./video.mp4') }}` as the v6.0.0-beta.6+ canonical form. In codebases following React Native's standard asset-typing convention, this canonical example doesn't compile against the v6 types. Consumers are instead forced to cast at the call site, widen workspace asset declarations to `NodeRequire` (which carries its own deprecation), or skip type-checking with `// @ts-expect-error`.

The runtime cleanly supports numeric ids at both source positions. Widening the type closes the documented "happy path" gap without changing any runtime behaviour.

### Steps to reproduce

1. Declare `*.mp4` modules as `ImageRequireSource` (RN-conventional asset typing).
2. Use `<Video source={require('./x.mp4')} />` or `<Video source={{ uri: require('./x.mp4') }} />`.
3. Run `tsc` - both fail with `TS2322` (number not assignable to `ReactVideoSource`).

### Reproduction repository

_No response_

`react-native-video`: `^6.19.2` (applies to the v6 stable line — the type was tightened around v6.0.0-beta.6 per the README's deprecation note about the legacy bare-require source form)

### react-native version

_No response_

_No response_

### Platforms

_No response_

### OS version

_No response_

### Device

_No response_

### Architecture

_No response_

### Expo

_No response_

### Last working version

_No response_

### Media / source type

_No response_

giaBaoJS:
## Summary

Fixes #4901.

### Motivation

The v6 runtime already accepts Metro asset IDs in both supported forms:

```tsx
<Video source={require('./video.mp4')} />
<Video source={{uri: require('./video.mp4')}} />
```

However, `ReactVideoSource` only accepted an object whose `uri` was a string or `NodeRequire`. Projects that declare media modules with React Native's `ImageRequireSource` therefore received `TS2322` for both documented runtime forms.

The earlier type-only attempt in #4904 was closed without merging. Widening the public union also requires narrowing the existing object-property reads in `Video.tsx` and `Video.web.tsx`; otherwise the v6 package no longer type-checks. This PR includes those internal accommodations and a regression fixture.

### Changes

- allow `ImageRequireSource` at the root of `ReactVideoSource` and in its `uri` field while retaining `NodeRequire` compatibility
- narrow or resolve numeric sources before reading object-only properties in the native and web components
- add a committed public-type fixture and run it from the JavaScript CI workflow
- unblock the branch CI so those checks can actually run (see below)

The existing documentation already shows the wrapped `require()` form, so no documentation text changes are needed.

### Why a CI commit is included

Both `Check JS` jobs on this branch were failing in the `Install node_modules` step, before `tsc` or `eslint` ever ran, so this PR could not be verified by CI. That failure is not caused by this change. Reproduced on a clean, unmodified `support/6.x.x` checkout with Node 20.20.2, the Node version the shared `setup-node` action pins:

```
$ yarn install --immutable --ignore-scripts
info No lockfile found.
error react-native@0.87.1: The engine "node" is incompatible with this module. Expected version "^22.13.0 || ^24.3.0 || >= 26.0.0". Got "20.20.2"
error Found incompatible module.
```

`yarn.lock` is git-ignored on this branch, so every run re-resolves the whole tree from the registry. `react-native@0.87.1` is not the top-level dependency; `react-native: ^0.78.0` still resolves to `0.78.3`. It arrives through `@types/react-native-web`, which declares `"react-native": "*"`. Pinning that single package only moves the failure to the next package that now requires Node 22 (`content-type@3.1.0`), so the fix is the Node version rather than a per-package pin.

Every `Check JS` run against `support/6.x.x` has failed since 2026-05-20, across three unrelated contributors. The last green run on the branch was 2026-04-28.

The included commit raises `.github/actions/setup-node/action.yml` from `node-version: 20.x` to `22.x`. On the unmodified base branch under Node 22 the install succeeds and both `yarn tsc` and `yarn lint` are already clean, so the bump changes nothing else. Happy to split it into its own PR if you would rather land it separately.

## Test plan

On the unmodified base branch, Node 22: `yarn tsc` clean, `yarn lint` clean over 39 files, 0 errors, 0 warnings.

On this branch, Node 22:

- `yarn tsc`: 0 errors
- `yarn test:types`: 0 errors
- `yarn lint -f @jamesacarr/github-actions`: 39 files, 0 errors, 0 warnings
- `yarn lint --fix` followed by `git diff --exit-code HEAD`: no changes

Counterfactuals, with the type fixture kept in place:

- reverting all three source files reproduces exactly the two reported errors, and nothing else:
  `.type-tests/video-source.tsx(8,14): error TS2559: Type 'number' has no properties in common with type 'Readonly<Omit<ReactVideoSourceProperties, "uri"> & { uri?: string | NodeRequire | undefined; }>'`
  `.type-tests/video-source.tsx(9,23): error TS2322: Type 'number' is not assignable to type 'string | NodeRequire | undefined'`
- keeping the widened public type but reverting the `Video.tsx` and `Video.web.tsx` narrowing fails `yarn tsc` with 16 `TS2339` errors, which is why those accommodations are part of the change
- the two `@ts-expect-error` controls are live: making either source valid raises `TS2578: Unused '@ts-expect-error' directive`, so invalid boolean and array sources are still rejected

Packed-consumer check: with v6.19.2 types both numeric forms fail with `TS2559` and `TS2322`; with this branch both compile.


giaBaoJS:
Heads up that the two red checks here aren't coming from this PR.

Both jobs die inside the shared `setup-node` action at install time, before eslint or tsc ever run:

```
error react-native@0.87.0: The engine "node" is incompatible with this module.
Expected version "^22.13.0 || ^24.3.0 || >= 26.0.0". Got "20.20.2"
```

`.github/actions/setup-node/action.yml` pins `node-version: 20.x`, and there's no lockfile committed on `support/6.x.x`, so the tree gets re-resolved from scratch on every run. Once a published react-native started requiring Node 22, the install broke on its own. The last green `check-js` run on this branch was the v6.19.2 release back in April, so I don't think anything opened against 6.x can go green right now.

Nothing here touches dependencies. The only `package.json` change is two script entries (`test` and `test:types`).

I'm happy to bump the action to Node 22 in this PR if that's useful, or leave it for a separate infra change if you'd rather keep this one focused on the types.

PR #5046, opened 2026-08-22
fix(web): respect resizeMode prop

### LINKED ISSUE #4297: [BUG]: `resizeMode` prop is not implemented on web
### What happened?

The `object-fit` style property of the video element in [Video.web.tsx](https://github.com/TheWidlarzGroup/react-native-video/blob/d934f214f5ac309490f0cf3fd80927c8657e2af3/src/Video.web.tsx#L371) should respect the `resizeMode` prop.

Currently the `resizeMode` prop is not implemented on web, so setting it (e.g. to `"cover"`) has no effect on how the video is displayed.

### Steps to reproduce

1. Set `resizeMode` to `"cover"` on a `Video` component.
2. Run the app on web and observe the video in a browser.
3. Notice that the video does not respect the `resizeMode` value.

### Reproduction repository

_No response_

6.8.0

### react-native version

_No response_

_No response_

### Platforms

Web

### OS version

Chrome

### Device

Real device

### Architecture

Old Architecture

### Expo

_No response_

### Last working version

_No response_

### Media / source type

_No response_

huytdps13400:
## Summary

Fixes #4297.

### Motivation

The v6 web renderer always applied `object-fit: contain` to the underlying `<video>`, so the public `resizeMode` prop had no effect in browsers. The v7 renderer already maps these values to CSS; this ports that behavior to the supported v6 branch.

### Changes

- map `contain` to `object-fit: contain`
- map `cover` to `object-fit: cover`
- map `stretch` to `object-fit: fill`
- preserve the existing `none` fallback as `object-fit: contain`
- add a built-output regression test for all four resize modes

The mapping uses static style objects and keeps the existing user `style` last in the style array, preserving v6 web override behavior.

## Test plan

- Red: `yarn test:web` failed because `resizeMode="cover"` still rendered `object-fit:contain`.
- Green: `yarn test:web` passes (TypeScript build + 4/4 Node DOM-render tests).
- `yarn lint` passes.
- `git diff --check` passes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
         "docs:llms-full": "node scripts/generate-llms-full.js",
         "release": "release-it",
         "test": "echo no test available",
+        "test:web": "yarn build && node --test tests/Video.web.test.mjs",
         "check-ios": "scripts/swift-format.sh && scripts/swift-lint.sh && scripts/clang-format.sh",
         "check-android": "scripts/kotlin-lint.sh",
         "check-all": "yarn check-android; yarn check-ios; yarn lint",
--- a/src/Video.web.tsx
+++ b/src/Video.web.tsx
@@ -54,6 +54,7 @@ const Video = forwardRef<VideoRef, ReactVideoProps>(
       rate,
       repeat,
       controls,
+      resizeMode = 'none',
       showNotificationControls = false,
       poster,
       fullscreen,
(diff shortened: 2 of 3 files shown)


huytdps13400:
CI note: both checks currently fail before running this PR’s code, in the shared **Install node_modules** step. The v6 workflow installs Node 20.20.2, while the lockfile-less `support/6.x.x` dependency range `react-native: ^0.78.0` now resolves to React Native 0.87.0, which requires Node `^22.13.0 || ^24.3.0 || >=26`. The same two install failures are visible on #5041 against this base branch.

Local verification on Node 24.16.0:
- `yarn test:web`: build passes; 4/4 resize-mode DOM-render tests pass
- `yarn lint`: passes
- `git diff --check`: passes

I kept the base-branch Node/dependency repair out of this focused behavior fix.

PR #5047, opened 2026-08-22
fix(android): annotate manifest track lookup worker thread

huytdps13400:
## Summary

Fixes #4885.

### Motivation

The retrying `getVideoTrackInfoFromManifest(int)` overload is correctly annotated with `@WorkerThread`, but its zero-argument wrapper was not. Android lint therefore inferred the wrapper as a UI-thread method and rejected its call into the worker-only overload, even though the production call site runs inside `ExecutorService.execute`.

### Changes

- annotate the zero-argument wrapper with the same `@WorkerThread` contract
- keep runtime behavior unchanged; this documents the existing background-thread requirement instead of suppressing `WrongThread`

## Test plan

Created a clean React Native 0.85.0 project matching the issue and installed this branch as the local `react-native-video` package.

- Red, before the annotation:
  - `./gradlew :react-native-video:lintDebug --no-daemon`
  - failed with exactly one lint error at `ReactExoplayerView.java:1611`: `getVideoTrackInfoFromManifest must be called from the worker thread`
- Green, after the annotation:
  - the same Gradle 9.3.1 lint command completed with `BUILD SUCCESSFUL`
- `yarn build`
- `yarn lint`
- `git diff --check`

PR #5048, opened 2026-08-23
fix: forward pointerEvents to video wrapper

### LINKED ISSUE #3315: [Bug]: pointerEvents is not respected
### What happened?

The `pointerEvents` prop is not respected by the `Video` component. When `pointerEvents="none"` is set, it is not passed through to the rendered `View`, so pointer/touch events do not pass through to content behind the video.

**Expected behaviour**

1. `pointerEvents` is passed through to the rendered `View`.
2. Pointer events go through to content behind the video.

**Reproducible sample code**

```jsx
<Video pointerEvents="none" />
```

### Steps to reproduce

1. Render a `Video` component with `pointerEvents="none"`.
2. Place interactive content behind the video.
3. Observe that the `pointerEvents` value is not passed through to the underlying `View`.
4. Attempt to interact with the content behind the video — interaction is blocked.

### Reproduction repository

_No response_

6.0.0

### react-native version

_No response_

_No response_

### Platforms

iOS, Android, Windows

### OS version

_No response_

### Device

iPhone 14

### Architecture

_No response_

### Expo

_No response_

### Last working version

_No response_

### Media / source type

_No response_

huytdps13400:
Fixes #3315

## Summary

Forward the `pointerEvents` prop to the outer `View` rendered by `Video`.

### Motivation

`ReactVideoProps` inherits `ViewProps`, but `pointerEvents` was left in `...rest` and forwarded to the native video child. The outer wrapper remained interactive, so `pointerEvents="none"` still blocked touches from reaching content behind the video.

Applying the prop at the wrapper boundary also preserves React Native's semantics for `box-none` and `box-only`; the native child is no longer given a duplicate value that could override those parent/child rules.

### Changes

- consume `pointerEvents` in the common `Video` component and pass it to the wrapper `View`
- add a bare-example React renderer regression test that verifies the wrapper receives the value and the native child does not
- enable the bare example's existing Jest/react-test-renderer toolchain with a React Native preset and local React/RN resolution

## Test plan

TDD evidence: before the source change, the renderer test failed with wrapper `pointerEvents: undefined`; after the change it passes.

- `yarn build` — passed
- `yarn lint` — passed
- `cd examples/bare && yarn test --runInBand` — 1/1 passed
- `cd examples/bare && yarn tsc --noEmit` — passed
- `cd examples/bare && yarn eslint __tests__/Video.test.tsx` — passed
- targeted Prettier check for all changed files — passed
- `git diff --check` — passed

The bare example's full `yarn lint` still reports pre-existing formatting/require-style errors in `metro.config.js` and `react-native.config.js`; the new test itself is lint-clean.


huytdps13400:
CI status: both checks stop before running this PR's code, in the shared root `Install node_modules` step. The support/6.x.x workflow uses Node 20.20.2 and has no lockfile, so `react-native: ^0.78.0` now resolves to `react-native@0.87.0`, whose engine requires Node `^22.13.0 || ^24.3.0 || >=26.0.0`.

This is the same base-branch infrastructure failure currently visible on #5046 and #5047. On Node 24.16.0, this head passes root build/lint, the bare regression test, bare typecheck, targeted test lint, and Prettier. No feature-branch change can make the current CI jobs reach those checks without also changing the shared workflow/dependency pin, which should be handled separately.

PR #5049, opened 2026-08-23
fix(types): accept bundled video asset ids

### LINKED ISSUE #4901: [BUG] `source` TS type rejects React Native's `ImageRequireSource` (number) for bundled assets

huytdps13400:
## Summary

- accept React Native `ImageRequireSource` values as direct video sources
- accept Metro asset IDs in the documented `source.uri` form while retaining `NodeRequire` compatibility
- narrow numeric sources before reading object-only options in native and web components
- add compile-only public API fixtures for both source shapes and wire them into `yarn tsc`

The runtime already resolves numeric Metro assets; this aligns the V6 TypeScript contract with those existing paths.

## Verification

- RED: the new fixture produced four TS assignment errors before the type change
- `yarn build` (root TypeScript + type fixtures)
- `yarn lint`
- Prettier check on all changed TypeScript/JSON files
- `git diff --check`

Fixes #4901


huytdps13400:
CI note: both jobs stop before running this PR in the shared `Install node_modules` step. The V6 workflow uses Node 20.20.2 and has no lockfile, so `react-native: ^0.78.0` resolves to `react-native@0.87.0`, which requires Node `^22.13.0 || ^24.3.0 || >=26.0.0`.

This is the same base-branch infrastructure failure currently affecting #5046–#5048. On Node 26.5.0, this head passes `yarn build` (including the new public type fixtures), `yarn lint`, Prettier checks, and `git diff --check`. I kept the shared workflow/dependency repair out of this focused V6 type fix.

PR #5085, opened 2026-09-06
fix(ios): do not publish the Swift bridging header as a public header

### LINKED ISSUE #5084: iOS build fails on React Native 0.87 (prebuilt core): "declaration of 'facebook' must be imported from module 'React'" — Video-Bridging-Header.h published as a public header
### What happened?

Building an iOS Debug simulator target fails while compiling the `ReactNativeVideo` Objective-C module, with:

```
Pods/Headers/Private/NitroModules/JSIConverter.hpp:19:17: error: declaration of 'facebook' must be imported from module 'React' before it is required
Pods/Headers/Private/NitroModules/JSIConverter.hpp:36:32: error: missing '#include "jsi/jsi.h"'; 'Runtime' must be declared before it is used
<unknown>:0: error: could not build Objective-C module 'ReactNativeVideo'
```

raised by `swift-frontend -emit-module` for the `ReactNativeVideo` target.

**Mechanism:** `ReactNativeVideo.podspec` sets `s.public_header_files = ["ios/Video-Bridging-Header.h"]`. `add_nitrogen_files()` (in `nitrogen/generated/ios/ReactNativeVideo+autolinking.rb`) then appends the Nitrogen-generated public headers to whatever `public_header_files` already contains:

```ruby
current_public_header_files = Array(spec.attributes_hash['public_header_files'])
spec.public_header_files = current_public_header_files + [
  "nitrogen/generated/shared/**/*.{h,hpp}",
  "nitrogen/generated/ios/ReactNativeVideo-Swift-Cxx-Bridge.hpp"
]
```

so `Video-Bridging-Header.h` always lands **first** in the generated `ReactNativeVideo-umbrella.h`, ahead of the Nitrogen `Hybrid*Spec.hpp` headers. `Video-Bridging-Header.h` is one line: `#import <React/RCTViewManager.h>`.

Under React Native 0.87's default prebuilt core (`RCT_USE_PREBUILT_RNCORE=1`), module `React`'s `module.modulemap` (`umbrella header "React-umbrella.h"`) owns `jsi/jsi.h` - the prebuilt `Headers/` tree that ships with the core contains it. Because the umbrella imports `React/RCTViewManager.h` first, module `React` is already fully loaded by the time NitroModules' `JSIConverter.hpp` does its own textual `#include "jsi/jsi.h"`. Clang then treats `jsi.h` as already owned by module `React` and refuses the textual include from inside a different module's build, producing the error above.

**Control:** two other Nitro-based libraries in the same app - `react-native-iap` 16.5.1 (`NitroIap`) and `react-native-mmkv` 4.3.2 (`NitroMmkv`) - build cleanly against the same `NitroModules` headers in the same build. Neither publishes a bridging header in its own umbrella (only their Nitrogen-generated headers are public), which is what isolates the trigger to this one podspec line.

Also checked and confirmed on this repo: no Swift file in `packages/react-native-video/ios` references any symbol from `RCTViewManager.h`, and both ObjC view manager files that need it (`ios/view/paper/RCTVideoViewViewManager.m`, `ios/view/fabric/RCTVideoViewViewManager.mm`) already `#import <React/RCTViewManager.h>` directly themselves. There is no `SWIFT_OBJC_BRIDGING_HEADER` wiring anywhere in the package that references `Video-Bridging-Header.h` by name - the only place it is referenced today is that one `public_header_files` line.

This reproduces the same way on `v7.0.0-beta.9` and current `master` (`v7.0.0-beta.11` plus later commits) - `git diff` between those two tags shows `ReactNativeVideo.podspec`'s `public_header_files` line and `ios/Video-Bridging-Header.h` are byte-identical between them, and unchanged on `master` since `v7.0.0-beta.11`.

The same app built cleanly against `react-native-video` on React Native 0.86.2 with the same Nitro versions.

I'm opening this alongside a one-line fix PR: removing `Video-Bridging-Header.h` from `public_header_files` (it stays compiled into the pod via the existing `ios/*.{h,m,mm,swift}` source_files glob; it just no longer sits in the public umbrella ahead of the Nitrogen headers).

### Steps to reproduce

There's no separate reproduction repository; any React Native 0.87.1 app with `react-native-video` 7.0.0-beta.9 (or beta.11), CocoaPods static linkage (no `use_frameworks!`), the default prebuilt core, and New Architecture on reproduces this:

1. Add `react-native-video` (7.0.0-beta.9 or 7.0.0-beta.11) to a React Native 0.87.1 app that also uses at least one other Nitro module (e.g. `react-native-nitro-modules` 0.36.5, `react-native-mmkv`, or `react-native-iap`).
2. `pod install` with static libraries (no `use_frameworks!`), New Architecture on, default `RCT_USE_PREBUILT_RNCORE=1` / `RCT_USE_RN_DEP=1`.
3. `xcodebuild ... -sdk iphonesimulator` (Debug configuration) for the app.
4. The build fails compiling the `ReactNativeVideo` Objective-C module with the error above; `react-native-mmkv` / `react-native-iap` in the same build succeed.

### Reproduction repository

_No response_ - see "Steps to reproduce"; the bug reproduces in any RN 0.87.1 app matching the environment above, not from anything specific to one app's code.

7.0.0-beta.9 (also confirmed reproducing on 7.0.0-beta.11 - see mechanism above)

### react-native version

0.87.1

0.36.5

### Platforms

iOS

### OS version

Xcode 26.6 (Build 17F113), iOS Simulator

### Device

Simulator / Emulator

### Architecture

New Architecture

### Expo

No (bare / React Native CLI)

### Last working version

Same react-native-video and react-native-nitro-modules versions built cleanly against react-native 0.86.2.

### Media / source type

_No response_ - this is a build-time module-compilation failure, unrelated to media/source type.

GratwickEnt:
## Summary

Fixes #5084. Removes `ios/Video-Bridging-Header.h` from `ReactNativeVideo.podspec`'s `public_header_files` so it no longer lands in the generated umbrella header ahead of the Nitrogen-generated headers.

## Motivation

Fixes #5084 - on React Native 0.87 with the default prebuilt core (`RCT_USE_PREBUILT_RNCORE=1`), building the `ReactNativeVideo` iOS module fails with:

```
Pods/Headers/Private/NitroModules/JSIConverter.hpp:19:17: error: declaration of 'facebook' must be imported from module 'React' before it is required
Pods/Headers/Private/NitroModules/JSIConverter.hpp:36:32: error: missing '#include "jsi/jsi.h"'; 'Runtime' must be declared before it is used
<unknown>:0: error: could not build Objective-C module 'ReactNativeVideo'
```

`add_nitrogen_files()` appends its generated public headers to whatever `public_header_files` already holds:

```ruby
current_public_header_files = Array(spec.attributes_hash['public_header_files'])
spec.public_header_files = current_public_header_files + [
  "nitrogen/generated/shared/**/*.{h,hpp}",
  "nitrogen/generated/ios/ReactNativeVideo-Swift-Cxx-Bridge.hpp"
]
```

so the bridging header (a single `#import <React/RCTViewManager.h>`) always precedes the Nitrogen `Hybrid*Spec.hpp` headers in `ReactNativeVideo-umbrella.h`. Under RN 0.87's prebuilt core, module `React` owns `jsi/jsi.h`; importing `React/RCTViewManager.h` first fully loads module `React`, and NitroModules' `JSIConverter.hpp`'s subsequent textual `#include "jsi/jsi.h"` then collides with a header Clang considers already owned by module `React`.

Full write-up, including the control test (`react-native-iap`/`react-native-mmkv` build fine against the same NitroModules headers because neither publishes a bridging header in its own umbrella) and confirmation this reproduces unchanged on `master`/`v7.0.0-beta.11`, is in #5084.

I checked whether this header needs to stay public for Swift access to `RCTViewManager`:
- No Swift file in `packages/react-native-video/ios` references any symbol from `RCTViewManager.h`.
- Both ObjC view manager files that use it (`ios/view/paper/RCTVideoViewViewManager.m`, `ios/view/fabric/RCTVideoViewViewManager.mm`) already `#import <React/RCTViewManager.h>` directly - they don't rely on this bridging header.
- There is no `SWIFT_OBJC_BRIDGING_HEADER` xcconfig entry, or any other reference to `Video-Bridging-Header.h` by name, anywhere in the package.

So its only present effect is occupying the first slot in the public header list. The header stays compiled into the pod (it's still matched by the existing `ios/*.{h,m,mm,swift}` `source_files` glob) - it's just no longer exposed as a public/umbrella header ahead of the Nitrogen headers. This does not change the public/private status of anything else: the podspec already fully overrides the "all headers public by default" behavior the moment it first sets `public_header_files`, so every other `.h` in the package was already private before this change; only this one file's status moves from public to private.

## Changes

- `packages/react-native-video/ReactNativeVideo.podspec`: remove the `s.public_header_files = ["ios/Video-Bridging-Header.h"]` line (and its now-stale `# Cxx to Swift bridging helpers` comment).

## Platforms affected

- [ ] Android
- [x] iOS
- [ ] visionOS
- [ ] tvOS / Android TV
- [ ] Windows
- [ ] Web

## Type of change

- [ ] Documentation only

## Test plan

- `ruby -c ReactNativeVideo.podspec` - syntax OK.
- Confirmed by reading `nitrogen/generated/ios/ReactNativeVideo+autolinking.rb` and grepping all Swift/ObjC sources in `packages/react-native-video/ios` that no code depends on this header being public (see "Motivation" above).
- Have not been able to run a full `pod install` + `xcodebuild` against this exact branch in this environment (no local app wired to a path dependency here); the fix is a pure `public_header_files` list edit with no effect on `source_files`, so nothing about what gets compiled changes, only what's exposed in the umbrella. Verified the mechanism instead by tracing `add_nitrogen_files()`'s header-ordering logic and confirming the described error against the actual failing app (details in #5084).

## Checklist

- [x] I read the contributing guidelines.
- [ ] I updated the documentation / README where relevant. (n/a - no public API/props/events changed, this is an internal build-config-only fix)
- [ ] If this changes how the library is used (API, props, events, behavior), I updated the AI agent skill (`skills/react-native-video/`) to match. (n/a - no such change)
- [x] This PR is focused on a single concern.
- [ ] I tested my changes on at least one platform. (see Test plan - traced and reasoned through rather than built end-to-end in this environment; happy to have a maintainer or CI confirm on the real app)


mziolkowskii (changes requested):
Thanks for the detailed write-up. I reproduced #5084 locally and verified this branch against it: our RNTA test app bumped to RN 0.87.1 / React 19.2.3, CocoaPods static libs, New Arch, default prebuilt core (`pod install` logs `[ReactNativeCore] Configuring xcconfig for prebuilt React Native Core`), Xcode 26, iOS 26 simulator.

### Confirmed

Your diagnosis holds, and the compiler is explicit about the ownership it rests on:

```
Pods/Headers/Private/NitroModules/JSIConverter.hpp:19:17: error: declaration of 'facebook'
    must be imported from module 'React' before it is required
Pods/Headers/Public/React-Core-prebuilt/jsi/jsi.h:705:18: note: declaration here is not visible
```

The rest of the write-up checks out too:

- **Header ordering.** Before the patch the generated `ReactNativeVideo-umbrella.h` starts with `#import "Video-Bridging-Header.h"` ahead of every Nitrogen `*.hpp`; after the patch it starts directly at the Nitrogen headers.
- **The header is unreferenced.** No Swift file in the package touches `RCTViewManager` or any `RCT*` symbol, both view managers import `<React/RCTViewManager.h>` directly, and no `SWIFT_OBJC_BRIDGING_HEADER` points at it.
- **Header status is as you describe.** After the patch the file is still installed, as a private header (`Pods/Headers/Private/ReactNativeVideo/`), and `Headers/Public/ReactNativeVideo/` holds only the Nitrogen headers both before and after.
- **No regression on the RN version we pin.** On RN 0.77.3: `pod install` + `xcodebuild -scheme ReactNativeVideo -sdk iphonesimulator -configuration Debug` → `** BUILD SUCCEEDED **`.

### The fix works, but it isn't sufficient on its own

With the podspec line removed, the `JSIConverter.hpp` error is gone. The build then fails one layer down:

```
ios/view/fabric/RCTVideoViewViewManager.mm:3:9: fatal error: 'RCTBridge.h' file not found
```

Three quoted imports of React headers in the package don't resolve under the prebuilt core:

- `ios/view/fabric/RCTVideoViewViewManager.mm:3` — `#import "RCTBridge.h"`
- `ios/view/paper/RCTVideoViewViewManager.m:2` — `#import "RCTEventDispatcher.h"`
- `ios/view/fabric/RCTVideoViewComponentView.mm:8` — `#import "RCTFabricComponentsPlugins.h"`

Under RN 0.77 built from source, CocoaPods flattens the React-Core headers and the quoted form resolves. Under 0.87's prebuilt core it doesn't — `RCTBridge.h` isn't anywhere in the Pods header tree at all, it lives inside the vendored `React.framework`. Switching all three to `<React/...>` fixes it.

One more blocker beyond this PR's scope, for the record: with `react-native-nitro-modules@0.35.0` the app builds and launches but immediately red-screens with `Failed to install Nitro: installJSIBindingsWithRuntime: was not called`. Bumping to `0.37.1` resolves it, and the library's committed nitrogen-0.35 output compiles and runs fine against that newer runtime.

With all three changes in place, our full Maestro E2E suite passes 10/10 on RN 0.87.1.

### What I'd like changed

**1. Delete `ios/Video-Bridging-Header.h` outright.** Leaving an unreferenced private header around invites someone to re-add it to `public_header_files` later and reintroduce exactly this bug. I tested that variant: with both the file and the podspec line removed, `pod install` is clean and the `ReactNativeVideo` target builds. The then-empty `ios/*.{h,m,mm,swift}` glob is fine — CocoaPods only requires `source_files` as a whole to match something, and we don't run `pod lib lint` in CI. Better to do this during the 7.0.0 beta than after a stable release.

**2. Add a short comment in the podspec** explaining that public headers must not import React, so the next person adding one knows why.

**3. Your call on the three import fixes:** fold them into this PR, or keep it single-concern and leave them to a follow-up? Either works for me. If you'd rather keep this PR minimal, let's drop "Fixes #5084" from the description so the issue doesn't auto-close while the build still fails on `RCTBridge.h`.

mziolkowskii:
Thanks @GratwickEnt for tracking this down in #5084. I've carried this change forward in #5090, together with the review feedback above (deleting `ios/Video-Bridging-Header.h` and a podspec comment on why public headers must not import React) and the three framework-style imports, so the iOS build works on React Native 0.87's prebuilt core. You're credited as co-author on the commit. Closing this in favour of #5090.

PR #4907, opened 2026-05-27
fix(android): controls always show in fullscreen

janpe:
Fixes #4734
Closes #4751

## Summary

This is a rebase of the stale [PR #4813](https://github.com/TheWidlarzGroup/react-native-video/pull/4813) by @robertert (preserved as the first commit, original authorship intact), with the second commit addressing the review feedback from @moskalakamil:

- Drop the two inline explanatory comments
- Inline the `shouldShowControls` boolean in `updateControllerVisibility`
- Replace the explicit `updateControllerConfig()` + `exoPlayerView.showController()` on fullscreen entry with `refreshControlsStyles()`, which (after the first commit removes the `!controls` guard) refreshes visibility unconditionally

Issue #4623, opened 2025-07-24
[Feature]: Live badge cannot be hidden on android exo player


apolyanov:
### Description

Currently, when using the ExoPlayer backend in react-native-video (on Android), any HLS or DASH stream flagged as live will automatically render a “LIVE” badge in the bottom‑right corner of the video surface. There is no exposed prop or configuration option to disable or customize this badge.

I’m proposing to add a new boolean prop (e.g. disableLiveBadge) to the <Video> component which, when set to true, prevents ExoPlayer from drawing its default live indicator overlay. Under the hood, this would simply call ExoPlayer’s UI controller method to hide the live badge (for example, by toggling playerView.setShowBuffering(false) or suppressing the DefaultTimeBar live indicator) without affecting any other live‑streaming functionality.

### Why it is needed ?

Design Consistency: Many apps have their own live‑stream branding, logos, or custom UI elements. Being forced to display ExoPlayer’s default “LIVE” badge can clash with established styles or on‑screen layouts.

Customization & Localization: Some products need a differently styled badge (different text, color, or language). Rather than stripping out ExoPlayer’s entire UI, it’s better to simply allow turning off the built‑in indicator and providing a hook for custom rendering.

Cleaner UX for Embedded Players: In use cases where the video is embedded inside a larger UI (e.g. within a card or alongside other controls), the badge can overlap important controls or content. Turning it off helps avoid accidental occlusion.

Parity with iOS/AVPlayer: The iOS AVPlayer integration in react-native-video does not show a hardcoded live badge—developers can implement their own if desired. Offering the same flexibility on Android ensures feature parity and predictable behavior across platforms.

### Possible implementation

Technical POV how to do it

### Code sample

<Video
  source={{ uri: liveStreamUri }}
+ disableLiveBadge={true}
  /* ...other props... */
/>

mlhtnc:
For ones who wait this feature, I have a workaround for android, but it's not recommended for all. Because you need to change native code in node_modules.

Open "your_project_path/node_modules/react-native-video/android/src/main/java/com/brentvatne/exoplayer/ExoPlayerView.kt". Then find updateLiveUi function and change **liveBadge.visibility to View.GONE**.

```kotlin
private fun updateLiveUi() {
    val player = playerView.player ?: return
    val isLive = player.isCurrentMediaItemLive
    val seekable = player.isCurrentMediaItemSeekable
    
    // Show/hide badge
    liveBadge.visibility = View.GONE
[...]

PR #5102, opened 2026-09-15
feat(android): add controlsStyles.hideLiveBadge to hide the LIVE badge

jmonsellier:
## Summary

Adds `hideLiveBadge` to `controlsStyles` on Android, so an app can hide the "LIVE" badge that `ExoPlayerView` draws over live streams. Default `false`: nothing changes for existing users.

### Motivation

Since the media3 `PlayerView` rewrite (#4581, shipped in 6.16.0), `ExoPlayerView` pins a red "LIVE" `TextView` to the top-left corner of the video surface whenever `Player.isCurrentMediaItemLive` is true. It is drawn even with `controls={false}`, and there is no prop to hide or replace it, while iOS draws no badge at all. Apps that render their own live indicator, or embed the player in a card, end up with a second badge they cannot remove.

Requested in #5004 (originally #4623), where exposing a prop was described as the targeted, contained change.

### Changes

- `ControlsConfig.kt`: new `hideLiveBadge` boolean, parsed from `controlsStyles`, default `false`.
- `ExoPlayerView.kt`: `setHideLiveBadge()` stores the flag and re-runs `updateLiveUi()`, which now keeps the badge `GONE` when the flag is set. The seek bar handling in `updateLiveUi()` is untouched.
- `ReactExoplayerView.java`: `setControlsStyles()` forwards the flag to the view, so toggling the prop at runtime takes effect immediately.
- `src/types/video.ts` and `src/specs/VideoNativeComponent.ts`: `hideLiveBadge?: boolean` on `ControlsStyles`.
- `docs/pages/component/props.mdx`: new row and example entry. The row notes that, unlike the other `controlsStyles` properties, this one also applies when `controls={false}`, since the badge itself is not tied to the controller.

`liveLabel` is parsed by `ControlsConfig` but has not been applied to this badge since the rewrite; that is left out of scope here.

## Test plan

1. Play a live HLS stream, for example `https://cph-p2p-msl.akamaized.net/hls/live/2000341/test/master.m3u8`, with `controls={false}`: the red "LIVE" badge shows in the top-left corner, as before.
2. Add `controlsStyles={{ hideLiveBadge: true }}`: the badge is gone. Toggle it back to `false` while playing: the badge returns.
3. Same with `controls={true}`: only the badge changes, the controller and its seek bar behave as before.
4. Play a VOD source with `hideLiveBadge: true`: no badge, nothing else changes.
5. iOS: no change, the prop is Android only like the rest of `controlsStyles`.

Checks run locally: `tsc`, `eslint` on the touched TypeScript files, `ktlint 1.0.1` with the repository `.editorconfig` on the touched Kotlin files.

~~CI note: on `support/6.x.x` the JS and Android example jobs fail before reaching the code, at `yarn install`, because react-native 0.87 now requires Node >= 22.13 while the workflows run Node 20. The same jobs failed on #4907; #5101 proposes the fix.~~ (already fixed ~@mziolkowskii). Kotlin lint passes.


mziolkowskii (approved):
LGTM, I dropped the "CI note" from the description, it no longer applies

PR #5113, opened 2026-09-17
fix(e2e): make nightly results trustworthy and unblock iOS on RN 0.82

### LINKED ISSUE #5104: [e2e nightly] e2e-android-rn0.77-api34-exclude-tags-flaky
Nightly E2E is failing for **e2e-android-rn0.77-api34-exclude-tags-flaky**.

Run: https://github.com/TheWidlarzGroup/react-native-video/actions/runs/35173748419

This issue is opened once per matrix row and closes automatically when the row goes green.

mziolkowskii:
## Summary

The first `e2e-nightly` runs on `master` were red for reasons that have nothing to do with the library: 9 issues (#5104–#5112) from one night, three of them false. This PR fixes the harness and CI so that a red nightly row means something. No library code is touched.

| # | Problem | Cause | Fix |
|---|---|---|---|
| 1 | All iOS rows `missing-report`, even green legs (#5107, #5111, #5112) | An upload path outside the workspace (`~/.maestro/tests/**`) moved the artifact root, so the report job did not find `e2e-report.xml` | Copy those logs into the workspace first (`_e2e-ios.yml`) |
| 2 | Every API 34 leg red, random flow on a white screen (#5104–#5106) | Android 14 kills the package's process ~1 s after `pm clear` ("remove task"); `launchApp: {clearState: true}` starts the app inside that window | `clearState`, 2.5 s wait (Android only), `launchApp` |
| 3 | Job summary never showed why a flow failed | Maestro 2.10 writes the reason as element text, the script read a `message` attribute | `junit-summary.mjs` reads both |
| 4 | One iOS leg lost 10/10 flows to a leftover "Open in RNVideoE2E?" prompt | The prompt appeared after the warm-up stopped waiting; it is SpringBoard's, survives relaunches and hides the app | `launch-app.yaml` answers a prompt that is already up, then relaunches |
| 5 | iOS never built on RN 0.82 (#5109, #5110) | `fmt` consteval errors when NitroModules' Swift compile builds the `fmt` module as C++20, and behind that a `glog` module error (include inside a namespace) | `Podfile` `post_install` patches `fmt/base.h` and `glog.modulemap` |
| 6 | A `simctl openurl` timeout failed a flow; the `retry` written for it never ran | Maestro's `retry`/`optional` only catch `MaestroException`; this is an `IllegalStateException` | iOS opens deep links through the fixture server, which retries simctl |

Details, logs and measurements are in the individual commit messages; `e2e/CONTEXT.md` gets a gotcha for 2, 4 and 6.

## Where to look

Two changes deserve a real review, the rest is mechanical:

- **`e2e/fixtures/open-link.mjs` + the `/__open-link` route in `serve.mjs`.** The fixture server now runs a process on request. It answers loopback only, accepts nothing but `rnvtest://scenario/<name>` and a simulator UDID (both regex-validated), and calls `xcrun` through `execFile`, never a shell. It retries the transport only: what the app shows is still asserted once (D11 holds). A link delivered twice is harmless because the test app keys its screen on the scenario; that was tested, see below. Android keeps Maestro's `openLink`, and the `retry` around it is removed rather than kept as decoration.
- **`test-app/ios/Podfile`.** `post_install` edits two files inside `Pods/`. Both edits are idempotent and skipped when the file is absent. The module map's mtime is pinned, because CocoaPods regenerates it on every `pod install` and a changed mtime would defeat the DerivedData cache. To be removed once RN ships an fmt that builds on Xcode 26 and Nitro stops pulling glog into its umbrella.

Cost: about 4 s per Android flow and about 7 s per iOS flow in extra waits. "Run flows" durations did not grow against runner variance (314–535 s after, 359–643 s before, same legs).

## Verification

Nightlies dispatched from this branch (they write to `e2e-results` and manage the nightly issues themselves, which is why this PR has no "Fixes #"):

| Run | Contains | Result |
|---|---|---|
| [35198102655](https://github.com/TheWidlarzGroup/react-native-video/actions/runs/35198102655) | fixes 1–3 | Android 9/9 (API 34: 3/3, was 0/3). iOS rows carry real reports. The report job closed #5104–#5108 and #5112. |
| [35204573155](https://github.com/TheWidlarzGroup/react-native-video/actions/runs/35204573155) | + fix 4 | Android 9/9, the leg from problem 4 green. Left: RN 0.82 build, one simctl timeout. |
| [35212563832](https://github.com/TheWidlarzGroup/react-native-video/actions/runs/35212563832) | + fixes 5, 6, review fixes | **15/15 rows green, 150/150 flows.** Both RN 0.82 iOS legs built and passed for the first time; RN 0.77 iOS builds with the new Podfile. The report job closed the last three issues (#5109, #5110 and the reopened RN 0.77 / iOS 26 row); no `e2e-nightly` issue is open. |

Reproduced locally before fixing (Xcode 26.6, iOS 26.5 simulator, API 34 emulator):

- **2:** `pm clear; sleep D; am start`, 15 launches each: 2 killed at D=0.9 s, 0 at D=2.5 s.
- **4:** an unanswered prompt makes the old `launch-app.yaml` fail exactly as in CI; the new one passes and later links open without a prompt.
- **5:** the NitroModules target fails with the CI errors before the change; after it the whole workspace builds and the app runs.
- **6:** warm-up and three flows pass through the new transport. With an `xcrun` wrapper that delivers the link and then reports a timeout, every open takes two attempts and the flows still pass. With the server down, a flow fails with a clear `open-link failed: … localhost:8090`.

Unit tests: 188 pass, 22 new (link/UDID validation, retry, the route against a fake `xcrun`, `junit-summary` edge cases). `bun lint` and `bun typecheck` pass.

## Risk and rollback

Nothing here ships to users. Each fix is its own commit and reverts cleanly on its own. The least proven one is 2: it rests on one logcat analysis plus the adb reproduction, so a few more nights will tell. RN 0.77 on iOS with the new Podfile was not built locally; run 35212563832 built and passed it on both runners.

Not addressed: two lost taps on RN 0.77 / iOS 26 from the first night (#5108, closed itself on the next green run), and the prompt check in 4 runs once per launch, so a prompt that appears later still costs that one flow.

## Platforms affected

- [x] Android
- [x] iOS
- [ ] visionOS
- [ ] tvOS / Android TV
- [ ] Windows
- [ ] Web

## Type of change

- [ ] Documentation only

## Checklist

- [x] I read the [contributing guidelines](https://github.com/TheWidlarzGroup/react-native-video/blob/master/CONTRIBUTING.md).
- [x] I added or updated tests that cover this change, or explained under "Test plan" why none apply.
- [x] For a bug fix the E2E suite can exercise, this PR includes a Maestro flow that fails without the fix, or "Test plan" explains why none does. _(None: harness, CI and build defects, not library behaviour.)_
- [x] `bun run test`, `bun lint` and `bun typecheck` pass locally.
- [x] I updated the documentation / README where relevant.
- [ ] If this changes how the library is used (API, props, events, behavior), I updated the AI agent skill (`skills/react-native-video/`) to match. _(Not applicable.)_
- [ ] This PR is focused on a single concern. _(One theme, six fixes: they were found by verifying each other on the nightly, and the later ones only became visible once the earlier ones were in. Each is a separate commit.)_
- [x] I tested my changes on at least one platform, or described under "Test plan" what CI (which runs both) should show.


Okelm (approved):

e2e/shared/open-link-ios.js
```diff
@@ -0,0 +1,28 @@
+// Run by open-scenario.yaml on iOS in place of Maestro's `openLink`. Asks the fixture
+// server to open the link, which retries `simctl openurl` when it times out - something
+// no flow can do, because Maestro's `retry` and `optional` do not catch that failure. See
+// e2e/fixtures/open-link.mjs. The port is the fixture server's (8090, the same one
+// test-app/src/e2e/fixtures.ts points the players at).
+//
+// env: LINK; SIM_UDID (optional, `maestro test -e SIM_UDID=...`: which simulator, else the
+// booted one); SOFT=true to report a failure instead of throwing (the warm-up, whose
+// later steps must still run).
+var device = typeof SIM_UDID === 'undefined' ? '' : SIM_UDID;
```
moskalakamil: let's change var's to let/const

e2e/fixtures/open-link.test.mjs
moskalakamil, line 1: I'm not sure if we need tests for scripts

e2e/fixtures/open-link.mjs
moskalakamil, line 1: I'm generally more in favor of using the .js extension, since marking something as a module is or should usually be handled in package.json. ES modules are the standard now anyway, so there’s no real need to additionally signal that in the file extension itself; but it's up to you.

moskalakamil (changes requested):
Let's also clean up the comments- we have too many of them imo

e2e/shared/open-link-ios.js
mziolkowskii: good catch! fixed

e2e/fixtures/open-link.test.mjs
mziolkowskii: This follows the convention [#5096](https://github.com/TheWidlarzGroup/react-native-video/pull/5096) already set: every script under `scripts/e2e/` and `e2e/fixtures/` has a `*.test.mjs` next to it (7 files on master), and they run in the pre-commit hook and CI. `open-link.mjs` is also the one place where request input becomes a process argument (xcrun simctl openurl <device> <link>), so the url/device validation and the retry loop are exactly the parts I want pinned down - the nightly on a macOS runner is the only other place this code runs, once a day. Keeping the tests.

e2e/fixtures/open-link.mjs
mziolkowskii: Agreed in principle, but the root package.json has no `"type": "module"`, and flipping it affects the whole repo (jest config, tooling scripts), which is out of scope for a nightly-fix PR. All 9 existing scripts under `scripts/e2e/` and `e2e/fixtures/` are .mjs, so a lone .js here would be the inconsistent one. Happy to do the .js + `"type": "module"` switch across the board in a separate PR if we want it.

moskalakamil (approved):

